Issue #28, opened 2017-05-11
Windows installation


bquistorff:
I know that Windows isn't officially supported, but it was fairly easy to get working so I thought I'd post how. The problem is that there are symlinks in `r-package/gradient.forest` to files elsewhere and while the Windows filesystem can make symlinks, it is now well supported in tools like unzip and some git implementations. The `ranger` package had a similar [issue](https://github.com/imbs-hl/ranger/issues/145). They fixed it by reversing the symlinks (having the real files in the package area and then symlinks in the other folders pointing there). 

To work around this issue on Windows, just:
1. Replace the symlinks in `r-package/gradient.forest` with the real files and directories (replace the `src/src` and `src/third_party` directories, and copy over the files from `bindings` to `src`). 
2. Run `r-package/build_package.R`
3. Then in R, `install.packages("path/to/gradient.forest_X.X.X.X.tar.gz", repos = NULL, type="source")`

susanathey:
Thanks SO much for the suggestion--this sounds like an easy fix and we'll try to implement it asap.  Let us know if you have other comments or suggestions--we're actively working on the package.

jtibshirani:
Thanks so much @bquistorff for the ticket! It's good to know it's very minor additional work to support Windows. Would it be possible for you to list your version of R, RTools, and the C++ compiler, so we can make sure to test Windows installation on our end?

I'm hesitant to simply reverse the symlinks, since we'd eventually like to add other bindings (python, etc.) and it'd be best if the `core` directory was the C++ source of truth. For now I've uploaded a version of the package to `releases/gradient-forest-alpha.tar.gz` without symlinks. This non-symlink version is also available as a branch, `test/windows-support`.

bquistorff:
I was using Microsoft R Open 3.3.2 and RTools 3.4 (which has the C++ compiler I believe). I was also able to install from the tar.gz you posted. Great! On Windows, installation can then be done in one line with:

```
install.packages("https://raw.github.com/swager/gradient-forest/master/releases/gradient-forest-alpha.tar.gz", repos = NULL, type="source")
```

jtibshirani:
I ended up removing the symlinks, and having `build_package.R` copy in the relevant source files before building the R package. Running `devtools::install_github`will no longer work, but I've updated the README to suggest using the command above (thanks!)

PR #45, opened 2017-05-18
Copy source files into the R package instead of using symlinks.

jtibshirani:
This should allow the 'build package' script to work on Windows.


jtibshirani:
Addresses #28.

Issue #146, opened 2017-11-28
Remove embedded copy of Eigen3 and use 'LinkingTo: RcppEigen'


eddelbuettel:
The underlying CRAN package could be made lighterweight by re-using the Eigen headers as shipped by RcppEigen.  Many other packages do (currently 127 reverse depends).  Code duplication is generally seen as something to avoid.  Obviously, as for other rules there we may be good exceptions and I may be overlooking something here.

But I tested this real quick by starting from yesterday's CRAN upload; just making in 'LinkingTo: Rcpp, RcppEigen' was all it took; I could then remove the `src/third_party` subtree.   Happy to send a PR if you're interested.

eddelbuettel:
FWIW:  

```diff
/tmp/grf$ diff -ru grf.orig/ grf
diff -ru grf.orig/DESCRIPTION grf/DESCRIPTION
--- grf.orig/DESCRIPTION        2017-11-27 02:49:29.000000000 -0600
+++ grf/DESCRIPTION     2017-11-27 16:03:53.387221249 -0600
@@ -1,6 +1,6 @@
 Package: grf
 Title: Generalized Random Forests (Beta)
-Version: 0.9.4
+Version: 0.9.4.1
 Author: Julie Tibshirani [aut, cre],
     Susan Athey [aut],
     Stefan Wager [aut],
@@ -15,7 +15,7 @@
     improvements to its performance and usability.
 Depends: R (>= 3.3.1)
 License: GPL-3
-LinkingTo: Rcpp
+LinkingTo: Rcpp, RcppEigen
 Imports: Rcpp (>= 0.12.10)
 RoxygenNote: 6.0.1
 Suggests: sandwich, testthat
Only in grf: grf_0.9.4.1.tar.gz
diff -ru grf.orig/src/Makevars grf/src/Makevars
--- grf.orig/src/Makevars       2017-11-26 17:30:45.000000000 -0600
+++ grf/src/Makevars    2017-11-28 09:31:09.200124034 -0600
@@ -4,7 +4,7 @@
 CC=/usr/local/bin/gcc-4.9
 CXX=/usr/local/bin/g++-4.9

-PKG_CXXFLAGS = -I. -Isrc -Ithird_party -DR_BUILD
+PKG_CXXFLAGS = -I. -Isrc -DR_BUILD

 SOURCES = $(wildcard *.cpp */*.cpp */*/*.cpp */*/*/*.cpp)
 OBJECTS = $(SOURCES:.cpp=.o)
[...]

PR #148, opened 2017-12-28
In Makevars, remove unused settings for CC and CCX.

jtibshirani:
These only make sense in a local context, and are not used when building the R package.


jtibshirani:
Addresses comment in #146.

PR #149, opened 2017-12-29
Use RcppEigen for the R package, as opposed to the eigen source.

jtibshirani:
This cuts down on the number of source files we need to distribute as part of
the R package, and will hopefully let us add support for sparse matrices.

Note that this introduces some fragility in the build, as the version of eigen we
include as source must match that in RcppEigen.


jtibshirani:
Addresses #146.

PR #160, opened 2018-01-07
Ensure we always default mtry to sqrt(p) + 20.

jtibshirani:
The old default of 2/3 * p was still being used in places, which was causing poor performance with even moderately large p.


jtibshirani:
Thank you @tibshirani for the bug report!

PR #184, opened 2018-03-27
Along with OOB predictions, return an unbiased MSE estimate.

jtibshirani:
This is a step towards being able to cross-validate parameters using a small number of trees.

### DIFF
--- a/core/src/prediction/RegressionPredictionStrategy.cpp
+++ b/core/src/prediction/RegressionPredictionStrategy.cpp
@@ -110,3 +110,28 @@ PredictionValues RegressionPredictionStrategy::precompute_prediction_values(
   return PredictionValues(values, num_leaves, 1);
 }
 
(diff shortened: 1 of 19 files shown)


core/src/prediction/RegressionPredictionStrategy.cpp
```diff
@@ -110,3 +110,32 @@ PredictionValues RegressionPredictionStrategy::precompute_prediction_values(
...
+    }
+
+    double tree_error = leaf_values.get(n, OUTCOME) - outcome;
+    bias += tree_error * tree_error;
+    num_trees++;
+  }
+
+  bias /= num_trees * (num_trees - 1);
+
+  if (bias < mse / 2) {
```
jtibshirani: @swager can you take a look here before we merge? I've seen cases where without this check, `mse - bias` can be negative.

PR #214, opened 2018-04-06
Remove the ability to sample with replacement.

jtibshirani:
Many critical parts of our approach, including honesty and subsampling for confidence
intervals, do not make conceptual sense unless sampling is done without replacement.


jtibshirani:
CC @swager 🌲🌲

PR #222, opened 2018-04-13
Add the ability to make local linear predictions.

jtibshirani:


core/test/forest/ForestCharacterizationTest.cpp
```diff
@@ -61,7 +61,7 @@ void update_predictions_file(const std::string file_name, std::vector<Prediction
   FileTestUtilities::write_csv_file(file_name, values);
 }
 
-TEST_CASE("quantile forest predictions have not changed", "[quantile], [characterization]") {
+TEST_CASE("quantile forest predictions have not changed", "[quantile, characterization]") {
```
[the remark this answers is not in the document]
jtibshirani: Actually... are you sure this is right? Now our automated travis tests are failing, because the characterization tests aren't being properly excluded when running this line: `./build/grf exclude:[characterization]`

core/test/forest/LocalLinearForestTest.cpp
```diff
@@ -0,0 +1,66 @@
...
+#include "forest/ForestTrainers.h"
+#include "utilities/ForestTestUtilities.h"
+
+#include "catch.hpp"
+
+
+TEST_CASE("LLF predictions vary linearly with Y", "[local_linear, forest]") {
+    // Run the original forest.
+
+    Data* data = load_data("test/forest/resources/gaussian_data.csv");
```
jtibshirani: Would it be possible to use a smaller dataset here? The tests takes quite a while to run. You could probably just copy the current dataset and cut it down to 50 samples.

r-package/grf/tests/testthat/test_regression_forest.R
```diff
@@ -96,3 +96,34 @@ test_that("regression forest tuning decreases prediction error", {
...
+
+    preds.grf = predict(forest, X.test)
+    preds.ll = predict(forest, X.test, local.linear=T)
+
+    mse.grf = mean( (preds.grf$predictions - Y.test)^2 )
+    mse.ll = mean( (preds.ll$predictions - Y.test)^2 )
+
+    expect_true( mse.ll < 1.5 )
+    expect_true( mse.ll < mse.grf )
+})
```
jtibshirani: newline

core/src/forest/ForestPredictors.cpp
```diff
@@ -48,3 +49,16 @@ ForestPredictor ForestPredictors::regression_predictor(uint num_threads,
...
+                                                         const Data *test_data,
+                                                         double lambda,
+                                                         bool use_unweighted_penalty) {
+  num_threads = ForestOptions::validate_num_threads(num_threads);
+  std::shared_ptr<DefaultPredictionStrategy> prediction_strategy(new LocalLinearPredictionStrategy(original_data,
+                                                                                                   test_data,
+                                                                                                   lambda,
+                                                                                                   use_unweighted_penalty));
+  return ForestPredictor(num_threads, prediction_strategy);
+}
```
jtibshirani: newline

r-package/grf/R/regression_forest.R
```diff
@@ -217,14 +217,18 @@ get_params <- function(X, draw) {
...
-#' 
+#'
 #' Gets estimates of E[Y|X=x] using a trained regression forest.
 #'
 #' @param object The trained forest.
 #' @param newdata Points at which predictions should be made. If NULL,
 #'                makes out-of-bag predictions on the training set instead
 #'                (i.e., provides predictions at Xi using only trees that did
 #'                not use the i-th training example).
+#' @param local.linear Optional local linear prediction correction
```
jtibshirani: Might be good to elaborate a tiny bit more on the algorithm? You can also mention that setting this to true will cause quite a slow down in prediction, and that this is a beta feature that we expect to make improvements to. Note these comments are also typically full sentences.

PR #240, opened 2018-04-23
Incorporate local linear correction for selected variables

rinafriedberg:
Added linear correction variables feature with options for manual and automatic variable selection, and basic tests of functionality.

### DIFF
--- a/r-package/grf/tests/testthat/test_regression_forest.R
+++ b/r-package/grf/tests/testthat/test_regression_forest.R
@@ -157,10 +157,4 @@ test_that("linear correction variables function as expected", {
     mse.selected = mean((preds.selected$predictions - MU)^2)
 
     expect_true(mse.selected < mse / 1.5)
-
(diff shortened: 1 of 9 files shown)


r-package/grf/tests/testthat/test_regression_forest.R
```diff
@@ -113,27 +113,54 @@ test_that("regression forest tuning only cross-validates null parameters", {
...
+    forest = regression_forest(X, Y)
+    preds = predict(forest, linear.correction.variables = 1:20)
+    mse = mean((preds$predictions - MU)^2)
+
+    preds.selected = predict(forest, linear.correction.variables = 1:3)
+    mse.selected = mean((preds.selected$predictions - MU)^2)
+
+    expect_true(mse.selected < mse / 1.5)
+
+    preds.automatic = predict(forest, select.correction.variables = TRUE)
```
jtibshirani: I think you need to remove these lines, as `select.correction.variables` is no longer an option.

Issue #238, opened 2018-04-22
Using causal_forest to estimate average treatment effects over subgroups


markhwhiteii:
After reading a few of the papers on causal forests, it seems like the idea is the trees will decide to split more often on variables that cause the treatment effect to vary (that is, moderators of the experimental effect). However, the object returned by `causal_forest` seems to work like any other machine learning algorithm focused on prediction—not causal statistical inference—in the way the `predict` function associated with it works.

How can I use the `causal_forest` function to find where the treatment varies?

For example, I simulate a randomized experiment with a binary outcome where there is a positive effect for women and a negative effect for black men. There are also nuisance variables in here. How would I use `causal_forest` to tell me that, "The trees are tending to split most on gender, and then when it leads to male, it the trees also are more likely to split on race." This would help show me where the treatment effects are occurring.

Here are the data:

```
set.seed(1839)
n <- 5000
X <- data.frame(
  condition = factor(sample(c("control", "treatment"), n, TRUE)),
  gender = factor(sample(c("male", "female"), n, TRUE)),
  race = factor(sample(c("black", "white", "hispanic", "asian"), n, TRUE)),
  generation = factor(sample(c("millennial", "x", "babyboomer"), n, TRUE)),
  has_kids = factor(sample(c("no", "yes"), n, TRUE))
)
y <- factor(ifelse(
  X$gender == "female" & X$condition == "treatment",
  sample(c("positive", "negative"), n, TRUE, c(.50, .50)),
  ifelse(X$race == "black" & X$condition == "treatment",
    sample(c("positive", "negative"), n, TRUE, c(.12, .88)),
    sample(c("positive", "negative"), n, TRUE, c(.40, .60))
  )
))
```

And then I fit a model using `grf::causal_forest`:

```
library(grf)
mod <- causal_forest(
  model.matrix(~ -1 + ., X[, -1]), # convert factors to dummies
  (as.numeric(y) - 1), # convert outcome to 0 or 1
  (as.numeric(X[, 1]) - 1), # convert treatment to 0 or 1
  num.trees = 4000
)
```

Looking at `mod`, I see some variable importance issues that hints at where the splits are occurring most:

```
Variable importance: 
    1     2     3     4     5     6     7     8 
0.233 0.203 0.267 0.061 0.055 0.046 0.081 0.054 
```

However, this doesn't capture how multiple `X` variables may depend on one another in their interaction with `W` on `Y` (what may be called a three-way interaction in the general linear model world).

[...]

PR #249, opened 2018-05-29
Added subsetting to average_treatment_effect

kingsharaman:


kingsharaman:
Currently working on the suggestions for precomputed predictions from @swager

jtibshirani:
Thank you for the PR @kingsharaman! As I mentioned in #238, it would be great to keep this PR scoped to adding support for subsetting. We can follow-up with a PR that allows OOB predictions to only be computed once (after we've agreed on a good approach on the issue).

r-package/grf/R/average_treatment_effect.R
```diff
@@ -47,12 +48,27 @@
...
+    Y.hat <- forest$Y.hat[subset]
+    W.orig <- forest$W.orig[subset]
+    W.hat <- forest$W.hat[subset]
+    clusters <- forest$clusters[subset]
+  } else {
+    Y.orig <- forest$Y.orig
+    Y.hat <- forest$Y.hat
+    W.orig <- forest$W.orig
+    W.hat <- forest$W.hat
+    clusters <- forest$clusters[subset]
```
swager: This line doesn't seem right.

r-package/grf/R/average_treatment_effect.R
```diff
@@ -18,6 +18,7 @@
 #'
 #' @param forest The trained forest.
 #' @param target.sample Which sample to aggregate treatment effects over.
+#' @param subset The subset to calculate CATE for.
```
swager: It's important to emphasize that subsetting should only be done based on subgroups defined by the X-variable (as opposed to the W- or Y-variables).

Also, please add an example using subsetting (e.g., at the end of the examples section, add one that subsets on X[,1] < 0).

swager:
Hi @kingsharaman, I just wanted to check in on this. Are you still interested in getting this PR merged?

kingsharaman:
Hi, definitely but I am afraid I will have no time in the next 2 weeks...maybe it should be reopened when I finished implementing your suggestions

swager:
Ok thanks for the update! No problem keeping this open for the next couple weeks.

markhwhiteii:
Any update on this?

r-package/grf/R/average_treatment_effect.R
```diff
@@ -125,16 +125,13 @@ average_treatment_effect = function(forest,
   
   # Retreive pointwise treatment effect predictions from forest, and
   # compute naive average effect estimates (notice that this uses OOB)
-  forest.with.preds <- forest
-  
-  if (is.null(forest.with.preds$predictions)) {
-    forest.with.preds$predictions <- predict(forest.with.preds)$predictions
-    eval.parent(substitute(forest <- forest.with.preds))
-  }  
+  predictions <- predict(forest)$predictions
```
markhwhiteii: Is it possible to allow the user to do this outside of the function and give the function these predictions? I'm thinking about a use case where one is trying to look at CATEs across multiple subgroups of multiple variables. I have a wrapper function that will take a forest and some variables and then look at CATEs for every level of each of those variables. This means that one has to generate predictions—the same predictions—every time. Allowing the user to specify this would allow the wrapper function to do it just once. Perhaps it's too focused on my specific use case, and I can edit the code myself.

Also, what is the significance that one is using out-of-bag and how do we know from the `predict` function that we are using out-of-bag?

For reference, here's that wrapper function:

```
# get conditional average treatment effects
  cates <- data.frame(var = rep(NA, 100), level = NA, 
                      cate = NA, lb = NA, ub = NA)
  counter <- 1
  for (i in seq_along(vars)) {
    for (j in levels(data[[vars[i]]])) {
      tmp <- cate(
        forest, 
        subset = data[, vars[i]] == j
      )
      cates$var[[counter]] <- vars[i]
      cates$level[[counter]] <- j
      cates$cate[[counter]] <- tmp[[1]]
      cates$lb[[counter]] <- tmp[[1]] - 1.96 * tmp[[2]]
      cates$ub[[counter]] <- tmp[[1]] + 1.96 * tmp[[2]]
      counter <- counter + 1
    }
  }
```

swager:
This feature is rather high-priority. Would it be possible to have this merged within a week?

Note that #261 saves OOB predictions, which will make subgroup ATE estimation faster.

kingsharaman:
I'm afraid I can't do that in the near future. Maybe someone else could finish this.

markhwhiteii:
What else needs to be done to have it merged?

swager:
The main TODO is around thorough testing; my comments from May 30 also had some other smaller points.

PR #254, opened 2018-06-22
sparse diagonal weights optimization

edgan8:
Two local optimizations to improve local linear regression

1) Treat the weights as a diagonal vector rather than a matrix. This avoids the cost of a large matrix multiply.
2) Only fill in entries with nonzero weights. This reduces the size of the matrices substantially for small leaf sizes.

In the new LLFBench test, the runtime on a 2000x10 dataset goes from 70s originally -> 0.62s with optimization 1 -> 0.10s with both optimizations. The speedup is even more dramatic on larger datasets. Switching from QR decomposition on the covariance matrix to Cholesky also improves runtime slightly (~15%).

friedman.csv is a small dataset generated from:

```
n = 2000
x = np.random.uniform(size=(10, n))
eps = np.random.normal(size=n)
y = (
    10*np.sin(math.pi*x[0]*x[1])
    + 20*(x[2] - 0.5)**2
    + 10*x[3]
    + 5*x[4]
    + eps
)
```


core/src/prediction/LocalLinearPredictionStrategy.cpp
```diff
@@ -46,35 +46,35 @@ LocalLinearPredictionStrategy::LocalLinearPredictionStrategy(const Data* origina
...
-
-  for (auto it = weights_by_sampleID.begin(); it != weights_by_sampleID.end(); ++it){
-    size_t i = it->first;
-    double weight = it->second;
-    weights(i,i) = weight;
+  size_t num_nonzero_weights = weights_by_sampleID.size();
+
+  std::vector<size_t> indices(num_nonzero_weights);
+  Eigen::MatrixXd weights_vec = Eigen::VectorXd::Zero(num_nonzero_weights);
+  size_t weights_index = 0;
```
jtibshirani: (nitpick) it would be more consistent with the loops below if this were named `i`, and we did `size_t index = it.first;`.

core/test/forest/LocalLinearForestTest.cpp
```diff
@@ -21,9 +21,48 @@
 #include "forest/ForestTrainer.h"
 #include "forest/ForestTrainers.h"
 #include "utilities/ForestTestUtilities.h"
+#include <chrono>
+#include <prediction/LocalLinearPredictionStrategy.h>
+#include <prediction/collector/DefaultPredictionCollector.h>
 
 #include "catch.hpp"
 
+TEST_CASE("LLF_friedman_2k") {
```
jtibshirani: We try to write test case descriptions in plain English -- I think this one would be 'LLF gives reasonable predictions on data generated from the Friedman function'? Or is this a characterization test (i.e. it just exists to check that predictions have not changed too much as we make performance improvements)?

rinafriedberg:
Looks good to me as well! Go for it @edgan8

core/src/prediction/LocalLinearPredictionStrategy.cpp
```diff
@@ -43,38 +43,41 @@ LocalLinearPredictionStrategy::LocalLinearPredictionStrategy(const Data* origina
...
-
-  for (auto it = weights_by_sampleID.begin(); it != weights_by_sampleID.end(); ++it){
-    size_t i = it->first;
-    double weight = it->second;
-    weights(i,i) = weight;
+  size_t num_nonzero_weights = weights_by_sampleID.size();
+
+  std::vector<size_t> indices(num_nonzero_weights);
+  Eigen::MatrixXd weights_vec = Eigen::VectorXd::Zero(num_nonzero_weights);
+  {
```
edgan8: braces here are to keep the scope of the loop variable encapsulated

jtibshirani: Oops, didn't see that this would conflict with the below -- feel free to ignore my nitpick and keep as before :)

jtibshirani:
I just realized that @edgan8 may not have merge privileges -- I'll merge this now, thanks again for the PR.

Issue #253, opened 2018-06-22
Performance of regression_forest depends on covariate order


halflearned:
**TL;DR**
The output of regression_forest seems to depend on covariate order. In one example below, reversing the order of covariates in the matrix caused a ~40% performance loss.

**Example overview** 

This [minimal example](https://gist.github.com/halflearned/2f510e8fd0b0d6134c6995ab5ebf9f13) reproduces the bug for different initial parameters. 

We create a matrix `X` of size (n, p). Only the first regressor, `X[,1]` is relevant. In fact, it is equal to `Y`. Other columns are just noise. 

We train the `regression_forest` on `(X, Y)` twice, with the same seed. The second time around, we feed `X` with its columns in reverse order (`X[,p]`, ..., `X[,1]` with the relevant regressor coming in last). 

**Results**

The performance , as measured by in-sample MSE, can be _very_ different. The table below shows how bad the performance difference can be, for different hyper-parameters and datasets sizes. We call the usual order "A" and the reversed order "B".

Note how the dependence on `mtry` is also quite bizarre -- see the numbers in bold below. This suggests a problem in how variables are being randomly selected.

You can also modify the file above so that all other columns of `X` are not just noise, but constant, and the results remain the same.

|   n|  p| num.trees| mtry|    A|    B| ratio|
|---:|--:|---------:|----:|----:|----:|-----:|
| 100|  2|       100|    1| 0.11| 0.12|  1.11|
| 100|  2|      1000|    1| 0.11| 0.11|  1.04|
| 500|  2|       100|    1| 0.02| 0.02|  0.90|
| 500|  2|      1000|    1| 0.02| 0.02|  1.00|
| 100| 10|       100|    1| 0.56| 0.55|  0.97|
| 100| 10|      1000|    1| 0.56| 0.59|  1.06|
| 100| 10|       100|   10| 0.07| 0.08|  1.21|
| 100| 10|      1000|   10| 0.07| 0.08|  1.19|
| 500| 10|       100|    1| 0.34| 0.36|  1.07|
| 500| 10|      1000|    1| 0.35| 0.34|  0.98|
| 500| 10|       100|   10| 0.01| 0.01|  1.27|
| 500| 10|      1000|   10| 0.01| 0.01|  1.23|
| 100| 20|       100|    1| 0.78| 0.78|  **1.00**|
| 100| 20|      1000|    1| 0.76| 0.76|  1.00|
| 100| 20|       100|   10| 0.22| 0.19|  **0.88**|
[...]

PR #259, opened 2018-07-13
RandomSampler::draw now uses Fisher-Yates algorithm

halflearned:
Removed dependence on covariate order documented in #253.

Note that `draw_simple` and `draw_knuth` were removed. Both are now just `draw`.


jtibshirani:
Thanks @halflearned for the PR. I noticed the build was failing, and think there are a few tests that need updating.

In particular, we have a set of 'characterization' tests that simply check that the algorithm's predictions haven't changed. These tests are helpful in determining that a refactor isn't accidentally changing the algorithm's statistical properties. In this PR, we actually expect to the algorithm's predictions to change, so we'll need to update the test files. You can do so by uncommenting `#define UPDATE_PREDICTION_FILES` in `ForestCharacterizationTest`, then running the test once. You'll notice that several files of the form `*_predictions.csv` should have changed -- you can then comment out the line again, and check that the tests pass.

I also noticed that `LLF gives reasonable prediction on friedman data` in `LocalLinearForestTest` is failing. @edgan8 would you be able to weigh in as to whether we can loosen the failing assertion? Or would this be better cast as a 'characterization' test as well, and moved to `ForestCharacterizationTest`?

core/src/sampling/RandomSampler.cpp
```diff
@@ -139,69 +139,27 @@ void RandomSampler::draw(std::vector<size_t>& result,
                          size_t max,
                          const std::set<size_t>& skip,
                          size_t num_samples) {
-  if (num_samples < max / 2) {
```
jtibshirani: I'm not sure we should remove the `draw_simple` algorithm. I think the motivation is that for small numbers of samples, the simple strategy can be substantially faster. Would it be possible to follow the same strategy as in the ranger commit here, where we keep the two sampling strategies, and decrease the threshold to `max / 10`?

core/src/sampling/RandomSampler.cpp
```diff
@@ -139,69 +139,27 @@ void RandomSampler::draw(std::vector<size_t>& result,
...
-  }
-}
 
-void RandomSampler::draw_simple(std::vector<size_t>& result,
-                                size_t max,
-                                const std::set<size_t>& skip,
-                                size_t num_samples) {
-  result.reserve(num_samples);
+  // Create indices
+  result.resize(max);
```
jtibshirani: For the logic in this method, I also think we should also go with exactly what ranger has. The ranger maintainers are generally very good about benchmarking changes, so we'll have confidence that the approach is fast. It also makes it easier to debug our code when there is a direct comparison to ranger when possible, as we can consult their bug reports, PRs, etc.

core/src/sampling/RandomSampler.cpp
```diff
@@ -139,69 +139,27 @@ void RandomSampler::draw(std::vector<size_t>& result,
...
-      for (auto& skip_value : skip) {
-        if (draw >= skip_value) {
-          ++draw;
-        }
-      }
-    } while (temp[draw]);
-    temp[draw] = true;
-    result.push_back(draw);
+  // Remove values that are to be skipped
+  result.erase(std::remove_if(result.begin(), result.end(),
```
jtibshirani: For example, I think that compared to the ranger approach, this will incur an extra linear lookup per element (through the check involving `skip.find(i)`).

core/src/sampling/RandomSampler.cpp
halflearned: Of course. Will do.

core/src/sampling/RandomSampler.cpp
halflearned: Sure -- I was a bit thrown off by `grf`s use of a `std::set` instead of a `std::vector` for `skip`, but I've figured it out.

jtibshirani: Makes sense, I had actually forgotten this was the case (and had to change my comment below to read 'log(n)' :) ).

halflearned:
Thank you for the detailed explanation about the 'characterization tests', @jtibshirani! After updating my local files, all tests pass. But should I commit the updated `*_predictions.csv` files, so that the tests will pass?

jtibshirani:
You're right, you should commit those new predictions files.

core/src/sampling/RandomSampler.cpp
```diff
@@ -142,7 +139,7 @@ void RandomSampler::draw(std::vector<size_t>& result,
   if (num_samples < max / 2) {
```
jtibshirani: I think this should now be changed to `max / 10`, following ranger's approach.

core/src/sampling/RandomSampler.cpp
```diff
@@ -228,4 +219,4 @@ void RandomSampler::draw_weighted(std::vector<size_t>& result,
 size_t RandomSampler::sample_poisson(size_t mean) {
   std::poisson_distribution<size_t> distribution(mean);
   return distribution(random_number_generator);
-}
+}
```
jtibshirani: (nitpick) missing line ending here

core/src/sampling/RandomSampler.cpp
halflearned: Sorry for making you point this out twice.

core/src/sampling/RandomSampler.cpp
```diff
@@ -172,36 +169,29 @@ void RandomSampler::draw_simple(std::vector<size_t>& result,
...
+                                      size_t max,
+                                      const std::set<size_t>& skip,
+                                      size_t num_samples) {
+
+  // Populate result vector with 0,...,max-1
+  result.resize(max);
+  std::iota(result.begin(), result.end(), 0);
+
+  // Remove values that are to be skipped
+  std::for_each(skip.rbegin(), skip.rend(),
```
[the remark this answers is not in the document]
edgan8: Also I'm worried that the complexity of repeatedly deleting from a vector will be worse than just copying over all the non-skipped elements of the vector, do you happen to know?

edgan8:
@jtibshirani  relaxing the assertion makes sense, I'm fine with expanding it to say +/- 3.0. I'm a little bit worried about making it a characterization test because it seems like the exact value depends on which machine you run it on: it's weird how the answer is different on my mac vs the CI machines. Have you ever had this issue with the other characterization tests?

core/src/sampling/RandomSampler.cpp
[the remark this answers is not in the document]
halflearned: 1. **for_each**
Ah, I wasn't exploiting any subtlety here. Is this a style violation? (There shouldn't be a problem in terms of performance either way).

2. Good point...
...But note that if there's a real performance bottleneck here, then the fastest way to construct `result` is to never have any of the values in `skip` in it to begin with. Like [this](https://gist.github.com/halflearned/d4a89ad1b28332c72037683a82d7838e), perhaps.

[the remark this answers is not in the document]
jtibshirani: With respect to the `for_each` call, I'm not aware of performance difference. I would be fine if we changed it to a standard for loop for consistency with the rest of the code (and with the ranger approach).

As for the repeated deletion, I am not too worried about this:
- The number of values in `skip` is expected to be very small. In our set-up, it should always be equal to the number of 'observables', which will be <= 3 values.
- We're following the same approach taken in ranger, which gives a strong baseline in terms of performance.

jtibshirani:
Thanks @edgan8, that makes sense to me -- it is a test on real data (based on a simulation from the LLF paper, if I'm not mistaken).

The characterization tests indeed will give different results depending on the machine, as the source of randomness is tied to compiler/ OS. This is certainly a downside, and means that we need to exclude them from our CI build (https://github.com/swager/grf/blob/master/.travis.yml#L32). I've found them indispensable when doing large refactors though, so I'm inclined to keep them.

core/src/sampling/RandomSampler.cpp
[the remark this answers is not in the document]
halflearned: So I think there's actually a good reason to use `for_each` here. Here it is.

We have to loop over `skip` in reverse order: from its tail (containing the largest element) to its head (containing the smallest). 

Now, in `ranger`, the variable `skip` [is a std::vector](https://github.com/imbs-hl/ranger/blob/98ec17048213d824c920bdae3ac9482275f02e41/src/utility.cpp#L128). This allows the code to access elements of `skip` using subscripts. That's what the `skip[skip.size() - 1 + i]` is doing.

Here, because `skip` is a std::set, we don't have the option of using subscripts, but we can still use iterators. This leaves us with two options. `for_each` like it's written now, or something like
```
for (auto it=skip.rend(); it != skip.rbegin(); ++i) {
    result.erase(result.begin() + (*it));
}
```
Neither of these should be better than the other in terms of performance, but I still find the `for_each` more readable.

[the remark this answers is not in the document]
jtibshirani: Right, that's the iterator-based for loop we had in mind (very common in C++). Happy to go with your preference in terms of readability. There may be a small overhead to using function pointers, but as I said above, this is called so few times I suspect there is not much difference.

PR #261, opened 2018-07-24
Compute OOB predictions during training

swager:
Most grf workflows involve computing OOB predictions on the training set. The current implementation doesn't cache these predictions, resulting in needless repeat computation. To avoid this, we now simply make OOB predictions by default during training, so that they are immediately available whenever needed.

Addresses a point discussed in #238

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,16 +107,6 @@ tau.forest = causal_forest(X[, selected.vars], Y, W,
                            W.hat = W.hat, Y.hat = Y.hat,
                            tune.parameters = TRUE)
 tau.hat = predict(tau.forest)$predictions
-
-# A simple omnibus test for signal uses tau.hat as a linear predictor for CATE.
-# A coefficient of 1 in the regression below suggests that tau.hat is well calibrated.
-DF = data.frame(dY = Y - Y.hat, dY.hat = (W - W.hat) * tau.hat)
-best.linear.predictor = lm(dY ~ dY.hat + 0, data = DF)
-
-# We need to use some extra packages to get
-# heteroskedasticity-consistent inference.
-lmtest::coeftest(best.linear.predictor,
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -107,6 +107,16 @@ tau.forest = causal_forest(X[, selected.vars], Y, W,
                            W.hat = W.hat, Y.hat = Y.hat,
                            tune.parameters = TRUE)
 tau.hat = predict(tau.forest)$predictions
+
+# A simple omnibus test for signal uses tau.hat as a linear predictor for CATE.
+# A coefficient of 1 in the regression below suggests that tau.hat is well calibrated.
+DF = data.frame(dY = Y - Y.hat, dY.hat = (W - W.hat) * tau.hat)
```
jtibshirani: It'd be great to have a helper function which performs this test.

README.md
```diff
@@ -107,6 +107,16 @@ tau.forest = causal_forest(X[, selected.vars], Y, W,
                            W.hat = W.hat, Y.hat = Y.hat,
                            tune.parameters = TRUE)
 tau.hat = predict(tau.forest)$predictions
+
+# A simple omnibus test for signal uses tau.hat as a linear predictor for CATE.
```
jtibshirani: Ideally this update to the `README` would be pulled out into its own PR, since it's unrelated to the OOB prediction change.

README.md
```diff
@@ -48,7 +48,7 @@ X.test = matrix(0, 101, p)
 X.test[,1] = seq(-2, 2, length.out = 101)
 
 # Train a causal forest.
-W = rbinom(n, 1, 0.5)
+W = rbinom(n, 1, 0.25 + 0.5 * (X[,1] > 0))
```
jtibshirani: I think we should remove this update, as it's not related to the OOB prediction change.

PR #282, opened 2018-08-09
Added wrapper to let users call local_linear_forest

rinafriedberg:
This PR adds a wrapper function local_linear_forest that creates a regression forest object, and ensures local linear prediction at the prediction step. The other change here is a separate test_that file for local linear forests, so that as we add this functionality it's easier to find those tests.


r-package/grf/tests/testthat/test_local_linear_forest.R
```diff
@@ -0,0 +1,112 @@
...
+    preds = predict(forest, linear.correction.variables = 1:20, tune.lambda = TRUE)$predictions
+    mse = mean((preds - truth)^2)
+
+    preds.untuned = predict(forest, linear.correction.variables = 1:20)$predictions
+    mse.untuned = mean((preds.untuned - truth)^2)
+
+    expect_true(mse < 0.75 * mse.untuned)
+})
+
+test_that("local linear forest wrapper and local.linear = TRUE are the same", {
```
swager: This test should be more explicit about what we're testing. There are 4 possible workflows

(a) train a regression forest and
(1) predict by averaging, or
(2) predict by LLF

or

(b) train an LLF with
(1) standard splitting or
(2) LLF splitting
note that we always LLF predict

We want to check that (a2) = (b1)

r-package/grf/tests/testthat/test_local_linear_forest.R
```diff
@@ -0,0 +1,112 @@
...
+
+    X = matrix(runif(n*p,-1,1), nrow = n)
+    truth = apply(X, FUN = mu, MARGIN = 1)
+    Y = truth + sigma*rnorm(n)
+
+    forest = regression_forest(X, Y)
+    preds = predict(forest, linear.correction.variables = 1:5, lambda = 0.1)$predictions
+
+    ll.forest = local_linear_forest(X, Y)
+    ll.preds = predict(ll.forest, lambda = 0.1)$predictions
```
swager: There should be a test that, if you trained an LLF, then prediction is always LLF (even if the user just calls predict(ll.forest)). If no lambda is given, we should probably CV on a plausible default range.

r-package/grf/R/local_linear_forest.R
```diff
@@ -0,0 +1,155 @@
...
+#' # Predict using the forest.
+#' X.test = matrix(0, 101, p)
+#' X.test[,1] = seq(-2, 2, length.out = 101)
+#' predictions = predict(forest, X.test)
+#'
+#' # Predict on out-of-bag training samples.
+#' predictions.oob = predict(forest)
+#' }
+#'
+#' @export
```
swager: Does this function ever get called? One would need to make sure relevant forests are of class `local_linear_forest` first; see https://adv-r.hadley.nz/s3.html#classes for more.

swager:
Looks good! One last thing -- could you please remove the call to `regression_forest` from `local_linear_forest`, and instead just duplicate the code? Like that, the next commit where we start introducing local linear splitting logic will be cleaner.

Issue #283, opened 2018-08-10
Run R tests in addition to C++ tests during continuous integration.


yatharth:
Currently, we use Travis CI for continuous integration, but the free tier only supports one language at a time, so only the C++ tests run, not the R ones.

As discussed with @jtibshirani, we could move to CircleCI and include both C++ and R tests.

PR #284, opened 2018-08-10
Continuous Integration with CircleCI

yatharth:
From #283:

> Currently, we use Travis CI for continuous integration, but the free tier only supports one language at a time, so only the C++ tests run, not the R ones.
> 
> As discussed with @jtibshirani, we could move to CircleCI and include both C++ and R tests.

 Confirmed to work on my fork.

Installing dependencies took forever (= 20–30min), so I created a [Docker image](http://hub.docker.com/r/yatharthagarwal/grf) for GRF. Most of the work is actually in [this repo](https://github.com/yatharth/grf-dockerfile), where I build and deploy the Docker image.

@jtibshirani I didn’t delete the `.travis.yml` file yet—feel free to do so after you’ve set up the CircleCI integration!

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,41 @@
+version: 2
+
+# `templates` is not a CircleCI key. I'm just defining shared YAML keys here,
+#  which I will reuse later using YAML references.
+templates:
+  job_defaults: &job_defaults
+    docker:
+      - image: yatharthagarwal/grf:3
+    steps:
+      - checkout
+      - run:
+          name: "Build and test C++ core"
+          command: |
+            mkdir ~/project/core/build && cd ~/project/core/build
+            cmake -DCMAKE_CXX_COMPILER=$COMPILER ..
+            make
+            cd ~/project/core
+            ./build/grf exclude:[characterization]
+            valgrind --leak-check=full --error-exitcode=1 ./build/grf exclude:[characterization]
+      - run:
+          name: "Build and test R package"
+          command: |
+            cd ~/project/r-package
+            cat build_package.R | R --vanilla  # Rscript does not work.
+
+jobs:
+  clang:
+    <<: *job_defaults  # YAML reference to the shared key defined above.
+    environment:
+      COMPILER: clang++-3.7
+  gpp:
+    <<: *job_defaults
+    environment:
+      COMPILER: g++-4.9
+
+workflows:
+  version: 2
+  all:
+    jobs:
+      - clang
+      - gpp


.circleci/config.yml
```diff
@@ -0,0 +1,49 @@
...
+    environment:
+      COMPILER: g++-4.9
+
+
+workflows:
+  version: 2
+  all:
+    jobs:
+      - clang
+      - gpp
```
jtibshirani: (nitpick) missing newline

.circleci/config.yml
```diff
@@ -0,0 +1,49 @@
+version: 2
+
+# This is not a CircleCI key. I'm just defining shared YAML keys here,
```
jtibshirani: Seems like a really small comment, but it'd be good to move this comment next to `job_defaults`. I scratched my head over the yaml syntax below, and didn't realize this comment was meant to explain it.

.circleci/config.yml
```diff
@@ -0,0 +1,49 @@
...
+#  which I will reuse later using YAML references.
+templates:
+  job_defaults: &job_defaults
+
+    docker:
+      - image: yatharthagarwal/grf:3
+
+    steps:
+
```
jtibshirani: (nitpick) would be good to remove extraneous newlines throughout the file

yatharth:
> My main concern with the current approach is that we're building + maintaining a custom docker image. I wonder if we could use the images published by the R consortium, which I think are pre-installed with R, plus gcc or clang: https://hub.docker.com/u/rhub/.

Hmm… the value-add is not intuitive to me; could you explain? Even with standard images, I suspect we would have to use a custom image anyway, because the standard images probably won’t come with all the tools we need, e.g., valgrind.

jtibshirani:
The main motivation is to keep our (already complex) build as simple as possible, and avoid maintenance costs associated with bumping to a new R version, etc. This is especially important because the large majority of our contributors are not software specialists, and won't be familiar with Docker. About valgrind specifically -- from a quick glance, I think most of these images come with a pretty full suite of tools: https://github.com/r-hub/rhub-linux-builders/blob/5dd16da8a94ae5d7bdd7b49c79ddaff4d383f76f/debian-gcc/Dockerfile#L65. Maybe we could try it out, and we can loop back if it turns out we need a custom image?

yatharth:
@jtibshirani The [custom Docker file](https://github.com/yatharth/grf-dockerfile/blob/master/Dockerfile#L33) also includes commands to install GRF’s R dependencies.

```
# Install dependencies to build R.
RUN apt-get -yq --no-install-suggests --no-install-recommends install libcurl4-gnutls-dev libssl-dev  # For devtools.
RUN apt-get -yq --no-install-suggests --no-install-recommends install libxml2-dev libxslt-dev  # For roxygen.
RUN apt-get -yq --no-install-suggests --no-install-recommends install gfortran libgfortran-8-dev liblapack-dev liblapack3 libopenblas-base libopenblas-dev  # For lmtest and RcppEigen.
RUN R -e 'install.packages(c("Rcpp", "devtools", "testthat", "roxygen2", "DiceKriging", "lmtest", "sandwich", "RcppEigen"), repos="http://cran.us.r-project.org")'
```
We could either (a) put the commands in CircleCI configuration, or (b) create a custom Docker image, but have the image build on one of those more standard images you helpfully mentioned, so it’s shorter/potentially easier to maintain.

Since the commands take a significant amount of time (on the order of 10 minutes, a several hundred percent increase in running time), and they’re not the purview of the GRF project, it makes me think that practically and logically, the commands are more at home in a Docker image rather than in CircleCI configuration file. Thus, the solution would be (b).

What do you think?

jtibshirani:
@yatharth I'm really sorry we dropped this on the floor. We ended up simply modifying the travis build for now (https://github.com/grf-labs/grf/pull/429), as it kept the build quite simple and resulted in a reasonable build time.

I'm going to close this PR, but we can re-open it in the case we want to switch to CircleCI.

PR #287, opened 2018-08-15
Add basic example of local linear simulation

rinafriedberg:
This PR has a quick script for generating the bias images from a random forest and a local linear forest. Adding now so we have some quick code for an example!


experiments/local_linear_examples/bias_image.R
```diff
@@ -0,0 +1,43 @@
+set.seed(1234)
+
+rm(list = ls())
+
+setwd("~/git/grf/experiments/local_linear_examples")
```
swager: It's better not to hard code paths. In RStudio, you can do the following:
```
setwd(dirname(rstudioapi::getActiveDocumentContext()$path))
```

jtibshirani: I just saw that we hardcode the paths in some of the instrumental variables examples. This set a bad precedent, we can plan to clean those up.

PR #289, opened 2018-08-27
Local linear causal prediction

rinafriedberg:
Preliminary structure to use local linear regression for causal forests. Currently includes short unit test and tests in R; would appreciate feedback on (a) structure of the code - note that I split out class files, although currently local linear and standard causal training is the same, and (b) additional tests that would be important to add here. Also happy to wait to merge until a cross-validation scheme is implemented, if you prefer!


core/test/forest/LocalLinearForestTest.cpp
```diff
@@ -105,3 +105,29 @@ TEST_CASE("LLF predictions vary linearly with Y", "[local_linear, forest]") {
   REQUIRE(equal_doubles(delta / predictions.size(), 1, 1e-1));
   delete data;
 }
+
+TEST_CASE("LLCF predictions are reasonable", "[local_linear], [causal], [forest]") {
```
jtibshirani: Is this meant as a characterization test (i.e., you're just checking that predictions won't change as you do refactors/ make performance improvements)? If so, it would be great to write it as a C++ test with the other cases in `ForestCharacterizationTest.cpp`.

Or maybe you're trying to sanity check the output of the forest? In that case, we tend to prefer invariant checks, since from C++ it can be hard to debug what's happening statistically. The suite `RegressionForestTest` has some examples of checking invariants, like 'honest regression forests are shift invariant'.

core/src/prediction/CausalPredictionStrategy.cpp
```diff
@@ -0,0 +1,111 @@
...
+  Eigen::MatrixXd Y (num_nonzero_weights, 1);
+  for (size_t i = 0; i < num_nonzero_weights; ++i) {
+    // X columns
+    for (size_t j = 0; j < num_variables; ++j){
+      size_t current_predictor = linear_correction_variables[j];
+      X(i,j+1) = test_data->get(sampleID, current_predictor)
+                 - original_data->get(indices[i],current_predictor);
+    }
+    // treatment W column
+    X(i, num_variables+1) = original_data->get(indices[i], num_total_predictors+1);
```
jtibshirani: I think it's best to do `observations.get(Observations::TREATMENT, indices[i])`.

PR #294, opened 2018-09-05
Buyan

### LINKED ISSUE #293: Zero-indexed output in trees$drawn_samples
[Example](https://gist.github.com/halflearned/66d102d25bdaad3c3b8bd18808eeac2e)

A recent commit fixed the indexing in [tree nodes](https://github.com/swager/grf/blob/660c06ca9c2196f13c378504c09465d95e74ad54/r-package/grf/bindings/AnalysisToolsBindings.cpp#L137), but `draw_samples` remains zero-indexed. 

```
forest <- grf::regression_forest(X, Y)
tree <- get_tree(forest, 1)
estimation_and_split_sample <- tree$drawn_samples  # Needs + 1 otherwise incorrect output 
```

Solution: add 1 to each entry in during tree [deserialization](https://github.com/swager/grf/blob/660c06ca9c2196f13c378504c09465d95e74ad54/r-package/grf/bindings/AnalysisToolsBindings.cpp#L158). Note: will need to remove `const` qualification.

Buyannemekh:
Fixes #293 along with a test for `draw_samples` are indexed correctly


r-package/grf/tests/testthat/test_analysis_tools.R
```diff
@@ -38,8 +38,37 @@ test_that("leaf samples are indexed correctly", {
 
 	samples = unlist(sapply(quantile.tree$nodes, function(node) node$samples))
 	expect_equal(seq(1, n), sort(samples))
+
+})
+
+test_that("draw samples are indexed correctly",{
+  p = 40
```
jtibshirani: Instead of this check, I think we should use the logic from the above test case "leaf samples are indexed correctly". It will be more consistent, and is also more of a concise + direct test.

r-package/grf/bindings/AnalysisToolsBindings.cpp
```diff
@@ -155,7 +155,9 @@ Rcpp::List deserialize_tree(Rcpp::List forest_object,
     node_objects.push_back(node_object);
   }
 
-  const std::vector<size_t>& drawn_samples = tree->get_drawn_samples();
+  std::vector<size_t> drawn_samples(tree->get_drawn_samples());
+  for (size_t& index : drawn_samples) index += 1; //R is 1-indexed.
```
jtibshirani: Two small comments:
- For clarity, it would be nice if this logic was consistent with the for loop above for `samples`. I find the approach around `samples` a little bit more clear, since it doesn't involve mutable element references.
- The standard throughout the C++ code is to use brackets with for loops, even if it's just one line.

Issue #273, opened 2018-08-04
Address poor performance of honest forests on small datasets.


jtibshirani:
When honesty is enabled, the training subsample is further split in half before performing splitting. With small datasets, this may not leave enough information for the algorithm to determine high-quality splits.

This issue is still pending a concrete proposal on how it should be addressed.

halflearned:
Just flagging that I'm currently working with @rugilmartin to allow users to select the fraction of data that should be used for splitting.

rugilmartin:
What has been done so far to allow users to select the fraction of data that should be used for honest splitting:
* The `honesty` argument in every type of forest (causal, custom, quantile, instrumental, and regression) has been changed to `honesty.fraction`.
    + **honesty:** was a boolean argument, which defaulted to `TRUE` and determined whether or not honest splitting would be used.
    + **honesty.fraction:** is a double argument, which defaults to 0.5 and determines the fraction of the sample that is used for the training and cross-validation step in honest splitting. Passing a value less than or equal to 0 or greater than or equal to 1 turns off honest splitting and results in a forest of non-honest trees.
    + This change is made in all *_forest.R files, all *_tuning.R files, and all *ForestBindings.cpp files.
    + This change is also represented in the auto-generated RcppExports.R file.
    + This change is also represented in ForestOptions.cpp, ForestOptions.h, TreeOptions.cpp, TreeOptions.h, and TreeTrainer.cpp.
* The new design passes all 145 existing `testthat` tests. 
* Advice on additional testing?
    + We don’t have any tests that deal with values other than 0.0 or 0.5 for `honesty.fraction`.
    + Only one existing test deals with `honesty.fraction` directly. This test used to set `honesty = FALSE`. It now sets `honesty.fraction = 0.0`. All other tests used the default value for `honesty`, so they now use the default value for `honesty.fraction`.
    + I’m not sure how to test that alternative values of `honesty.fraction` are performing as expected/desired, so any advice would be much appreciated!
    + One thought I had: if a test sets a value of `honesty.fraction` that is close to 0.0 or 1.0, we would expect larger confidence intervals, since in one case there isn’t much data available for training and cross-validation and in the other there isn’t much data available for prediction. Is my understanding here correct? And if so, does this sound like a good test?
* The default value for `honesty.fraction` is the same as the previous fixed value for honest splitting. The behavior of the package is unchanged in cases where the default is used: honest splitting on, with half of the sample used for training and cross validation.
[...]

PR #297, opened 2018-09-08
Allowing users to change fraction of data used for honest splitting

rugilmartin:
This originated as a possible way to address Issue #273.

## Content of changes 
* In addition to the `honesty` argument for every type of forest (causal, custom, quantile, instrumental, and regression) an `honesty.fraction` argument has been added.
    + **honesty:** is still a boolean argument, which defaults to `TRUE` and determines whether or not honest splitting will be used.
    + **honesty.fraction:** is a double argument, which defaults to 0.5 and determines the fraction of the sample that is used for the training and cross-validation step in honest splitting, if and only if `honesty = TRUE`. 
    + This change is made in all *_forest.R files, all *_tuning.R files, and all *ForestBindings.cpp files.
    + This change is also represented in the auto-generated RcppExports.R file.
    + This change is also represented in ForestOptions.cpp, ForestOptions.h, TreeOptions.cpp, TreeOptions.h, TreeTrainer.cpp, ForestTrainer.cpp, input_utilities.R, and test_regression_forest.R.
* The default value of 0.5 for `honesty.fraction` is the same as the previous fixed value for honest splitting. The behavior of the package is unchanged in cases where the default is used: honest splitting on, with half of the sample used for training and cross validation.
* See below for summary of error handling.

## Does it work?
* The new design passes all 145 existing `testthat` tests, and 4 new tests in test_regression_forest.R.
    + Two of the new tests implement the suggestion @susanathey made in a [comment](https://github.com/swager/grf/issues/273#issuecomment-417865361) on Issue #273. They confirm that the forest object contains the expected number of observations in its training and estimation sets, based on the values of `sample.fraction` and `honesty.fraction` and the number of observations available. I used a function kindly provided by @halflearned to extracts the different samples from a tree.
    + Two of the new tests ensure that the error checking described below works properly. They test that the `train` method in ForestTrainer.cpp throws an error if `honesty.fraction` is set very close to 0 or 1 in a way that prevents any observations from being sampled for the training set or the estimation set.
* @halflearned and I have not yet tested whether changing `honesty.fraction` away from the default value of 0.5 has clear benefits for the performance of honest forests on small datasets.

## Design decisions & error handling
* The decision to keep both parameters, `honesty` and `honesty.fraction` was made for the reasons outlined in @jtibshirani ‘s [comment](https://github.com/swager/grf/issues/273#issuecomment-418136290) on Issue #273 as well as the one described briefly here:
    + R uses partial string matching for arguments, which would lead to confusion if `honesty` were removed. Users who weren’t aware of the change might still passing an argument like `honesty = TRUE`, but R would interpret this as `honesty.fraction = TRUE` which would be coerced to a double by Rcpp, yielding `honesty_fraction = 1` on the C++ side of GRF. 
    + If you are interested in the problems posed by partial string matching in the case where the `honesty` argument is deprecated, I wrote a summary [here](https://github.com/swager/grf/issues/273#issuecomment-417831072).
    + Ultimately, it was most practical to keep the `honesty` argument and use the `honesty.fraction` argument only when `honesty = TRUE`. 
* Handling of unusual/erroneous cases:
    + The test in [line #43](https://github.com/swager/grf/blob/master/core/src/forest/ForestTrainer.cpp#L43) of ForestTrainer.cpp has been expanded. It now uses `honesty`, `sample.fraction`, and `honesty.fraction` to make sure that there is at least one observation in the training sample and at least one observation in the estimation sample.
    + The test described above also handles the case where the user passes a value less than or equal to 0 or greater than or equal to 1 for `honesty.fraction`. In this case, if `honesty = FALSE`, `honesty.fraction` is ignored. If `honesty = TRUE`, an error is produced.
    + If the user passes `honesty = FALSE` but sets `honesty.fraction` to some positive value less than one, but not equal to 0.5, we ignore the value of `honesty.fraction`. This case would indicate that the user deliberately tried to override both the default value of 0.5 for `honesty.fraction` and the default value of TRUE for `honesty`, but in a way that makes the value of `honesty.fraction` meaningless. This may indicate some misunderstanding on the user’s part, but the naming and documentation around both `honesty` and `honesty.fraction` should make it obvious that the user shouldn't expect `honesty.fraction` to work in this case. This is also consistent with how grf handles other parameters that depend on a flag being set (e.g. `clusters` and `samples.per.cluster`).


r-package/grf/R/instrumental_forest.R
```diff
@@ -68,27 +71,35 @@ instrumental_forest <- function(X, Y, W, Z,
                                 seed = NULL,
                                 clusters = NULL,
                                 samples_per_cluster = NULL) {
+    if (!is.double(honesty.fraction)){
```
jtibshirani: I think this check is left over from your initial approach.

r-package/grf/R/causal_forest.R
```diff
@@ -131,13 +134,14 @@ causal_forest <- function(X, Y, W,
...
     samples_per_cluster <- validate_samples_per_cluster(samples_per_cluster, clusters)
+    honesty.fraction <- validate_honesty_fraction(honesty.fraction, honesty)
     
     reduced.form.weight <- 0
 
     if (is.null(Y.hat)) {
       forest.Y <- regression_forest(X, Y, sample.fraction = sample.fraction, mtry = mtry, tune.parameters = tune.parameters,
-                                    num.trees = min(500, num.trees), num.threads = num.threads, min.node.size = NULL, 
-                                    honesty = TRUE, seed = seed, ci.group.size = 1, alpha = alpha, imbalance.penalty = imbalance.penalty,
+                                    num.trees = min(500, num.trees), num.threads = num.threads, min.node.size = NULL, honesty = honesty,
```
jtibshirani: Here, we're changing some behavior that is unrelated to the addition of `honesty.fraction`: before the orthogonalization forests were always trained with `honesty`, and now they inherit the passed-in value of `honesty`. In general, it's best not to tack on a non-core change to a PR (especially one that affects user results), as it makes it hard to track down when/ why a change was made.

If we feel that the orthogonalization forests should inherit the `honesty` parameter, I think we should pull this out and commit it in a separate PR.

core/src/forest/ForestTrainer.cpp
```diff
@@ -39,11 +39,18 @@ const Forest ForestTrainer::train(Data* data,
   size_t num_samples = data->get_num_rows();
   uint num_trees = options.get_num_trees();
 
-  // Ensure that the sample fraction is not too small.
+  // Ensure that the sample fraction is not too small and honesty fraction is not too extreme.
+    bool honesty = TreeOptions(options.get_tree_options()).get_honesty();
+    double honestyfraction = TreeOptions(options.get_tree_options()).get_honesty_fraction();
```
jtibshirani: (small comment) `honestyfraction` -> `honesty_fraction`

jtibshirani: I don't think we need to wrap `options.get_tree_options()` in another constructor call (`TreeOptions(...)`).

r-package/grf/tests/testthat/test_regression_forest.R
```diff
@@ -2,6 +2,55 @@ library(grf)
...
+  leaf_nodes <- Filter(f = function(x) x$is_leaf, tree$nodes)
+  
+  # Leaf nodes' 'samples' are estimation samples
+  estimation_sample <- unlist(Map(f=function(x) x$samples, leaf_nodes)) - 1
+  
+  # Split = Drawn - Samples
+  split_sample <- base::setdiff(tree$drawn_samples, estimation_sample)
+  
+  # OOB samples
+  oob_sample <- base::setdiff(seq(2*length(tree$drawn_samples)-1), tree$drawn_samples)
```
jtibshirani: I think we can remove the `drawn_sample` and `oob_sample` keys, as they're not used in tests.

r-package/grf/tests/testthat/test_regression_forest.R
```diff
@@ -2,6 +2,55 @@ library(grf)
 
 set.seed(1234)
 
+rm(list=ls())
```
jtibshirani: I think this line was probably just for debugging, and we should nuke it.

core/src/forest/ForestTrainer.cpp
```diff
@@ -39,11 +39,19 @@ const Forest ForestTrainer::train(Data* data,
   size_t num_samples = data->get_num_rows();
   uint num_trees = options.get_num_trees();
 
-  // Ensure that the sample fraction is not too small.
+  // Ensure that the sample fraction is not too small and honesty fraction is not too extreme.
+  TreeOptions tree_options = options.get_tree_options();
```
jtibshirani: When we access the tree options, we should use a const reference to avoid making a copy: `const TreeOptions& tree_options = options.get_tree_options();`.

r-package/grf/R/instrumental_forest.R
```diff
@@ -73,22 +77,24 @@ instrumental_forest <- function(X, Y, W, Z,
...
     
     if (!is.numeric(reduced.form.weight) | reduced.form.weight < 0 | reduced.form.weight > 1) {
         stop("Error: Invalid value for reduced.form.weight. Please give a value in [0,1].")
     }
     
     if (is.null(Y.hat)) {
       forest.Y <- regression_forest(X, Y, sample.fraction = sample.fraction, mtry = mtry, 
-                                    num.trees = min(500, num.trees), num.threads = num.threads, min.node.size = NULL, 
-                                    honesty = TRUE, seed = seed, ci.group.size = 1, alpha = alpha, imbalance.penalty = imbalance.penalty,
+                                    num.trees = min(500, num.trees), num.threads = num.threads, min.node.size = NULL, honesty = honesty,
```
jtibshirani: We should keep `honesty = TRUE` in this function as well.

PR #324, opened 2018-11-05
Fix some issues in local linear tuning.

jtibshirani:
* Fix a bug in local linear prediction where the penalty wasn't properly calculated.
* Make sure the default ridge penalty is the same for tuning and prediction.

### DIFF
--- a/core/src/prediction/LocalLinearPredictionStrategy.cpp
+++ b/core/src/prediction/LocalLinearPredictionStrategy.cpp
@@ -34,12 +34,12 @@ size_t LocalLinearPredictionStrategy::prediction_length() {
 LocalLinearPredictionStrategy::LocalLinearPredictionStrategy(const Data* original_data,
                                                              const Data* test_data,
                                                              std::vector<double> lambdas,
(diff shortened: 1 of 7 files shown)


core/src/prediction/LocalLinearPredictionStrategy.cpp
```diff
@@ -76,24 +76,27 @@ std::vector<double> LocalLinearPredictionStrategy::predict(
   }
 
   // find ridge regression predictions
-  Eigen::MatrixXd M (num_variables+1, num_variables+1);
-  M.noalias() = X.transpose()*weights_vec.asDiagonal()*X;
+  Eigen::MatrixXd M_penalized(num_variables+1, num_variables+1);
+  M_penalized.noalias() = X.transpose()*weights_vec.asDiagonal()*X;
```
swager: Should this be M_unregularized rather than penalized?

jtibshirani: Yes, thanks! Was rushing on this a bit :)

Issue #325, opened 2018-11-07
instrumental_prediction is not being corrected for MC error


halflearned:
**Description of the bug**
The function `compute_debiased_error` actually returns raw estimates of the error, without debiasing.

The issue is in [this line](https://github.com/grf-labs/grf/blob/9c349c99392124e297a6a9af701c349d3801cdb8/core/src/prediction/InstrumentalPredictionStrategy.cpp#L246). The variable `num_trees` has type `size_t`, and at that point in the code we have

`error_bias *= (num_trees - 1)/num_trees` 

Since the right-hand side is a division of two integers and the result is smaller than one, it gets rounded down to zero. This means that the jackknife estimates of monte-carlo bias are multiplied by zero.

**Steps to reproduce**
It's a bit hard to observe this bug right now, since the monte carlo bias is computed but not returned by the function `compute_debiased_error`. I'm currently working on #288 that would expose this issue. (This is how I found the bug).

**Steps to resolve**
Simply cast to double:
`error_bias *= ((double)(num_trees - 1)) / (double)(num_trees);`

**GRF version**
'development'

**Notes**
This may affect causal forest tuning, since I think this function is called by `instrumental_predict_oob`. This does not affect predictions or tuning of regression forests.

halflearned:
I added a quick patch for this in #326. Before the end of the week I'll submit a PR for #288 that will include tests and ensure that MC error is nonzero. Not sure if you guys want to wait until then.

halflearned:
Merged in #326

Issue #288, opened 2018-08-15
Provide estimate of the Monte Carlo variance of a trained forest


swager:
Once we've trained a forest, it's important to know whether we grew enough trees, or if more trees would improve the accuracy of the estimates. This issue is about providing users with an idea of how much they should expect test set error to improve if they could grow arbitrarily many trees.

One simple way to do this is to use a jackknife on trees: We look at what the forest would predict if we remove one tree at a time from the forest, and use any instability here to estimate the variance of the forest.

Let \hat{\theta}(x) denote the prediction made by a forest at x using B trees (theta could be a conditional mean in the regression case, or a CATE in the causal forest case); and let \hat{\theta}^{(-b)}(x) be the forest estimate produced with all but the b-th tree (i.e., using the B - 1 remaining trees). The jackknife estimate of Monte Carlo variance for the forest is then:

```
\hat{V}_MC(x) = (B-1) / B \sum_{b = 1}^B (\hat{\theta}^{(-b)}(x) - \hat{\theta}(x))^2.
```

In other words, the claim is that the forest estimates \hat{\theta}(x) suffer from excess noise \hat{V}_MC(x) due to the fact that we are using a finite number of trees. Ideally, one would want to grow enough trees such that the average of \hat{V}_MC(X) would be much smaller than the sample variance of the \hat{\theta}(X) (as this guarantees that most of the variance in estimates across different points is not due to Monte Carlo effects).

As a simple motivation for this idea, consider the case of the regression forest, where the forest actually predicts using an average of trees: \hat{\mu}(x) = 1/B \sum_{b = 1}^B Tb(x), where Tb(.) is the b-th tree. In this case, we can verify that \hat{\mu}^{(-b)}(x) = 1/(B - 1) \sum_{b' != b} Tb'(x), and so the jackknife recovers the usual variance estimate:

```
\hat{V}_MC(x) = 1 / (B (B-1)) \sum_{b = 1}^B (Tb(x) - \hat{\mu}(x))^2
```

In other cases, the jackknife estimate of Monte Carlo variance won't have such a simple form anymore, but it will still be straight forward to compute; see, e.g., <a href="https://github.com/swager/grf/blob/a8b8963a4970cc9d07c1bdf79574a14943aaeaee/core/src/prediction/InstrumentalPredictionStrategy.cpp#L228">this snippet</a> for a part of the code where we compute something very similar for causal forests.

[...]

PR #326, opened 2018-11-07
Fix error prediction in instrumental_predict

halflearned:
Solves #325

### DIFF
--- a/core/src/prediction/InstrumentalPredictionStrategy.cpp
+++ b/core/src/prediction/InstrumentalPredictionStrategy.cpp
@@ -242,7 +242,7 @@ std::vector<double> InstrumentalPredictionStrategy::compute_debiased_error(
     double residual_loto = outcome - (treatment - treatment_loto) * treatment_effect_estimate_loto - outcome_loto;
     error_bias += (residual_loto - residual) * (residual_loto - residual);
   }
-  
-  error_bias *= ((double)(num_trees - 1)) / (double)(num_trees);
+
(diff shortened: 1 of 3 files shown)


core/src/prediction/InstrumentalPredictionStrategy.cpp
```diff
@@ -243,6 +243,6 @@ std::vector<double> InstrumentalPredictionStrategy::compute_debiased_error(
     error_bias += (residual_loto - residual) * (residual_loto - residual);
   }
   
-  error_bias *= ((num_trees - 1) / num_trees);
+  error_bias *= ((double)(num_trees - 1)) / (double)(num_trees);
```
jtibshirani: I would just have one cast (two is redundant).

halflearned:
Hey @jtibshirani, I added a couple of quick tests. It feels a bit hacky, because we're testing a quantity that was computed but not returned by the `compute_debiased_error` function. Do you have other ideas of what to test at the moment? After #288 we can have something more sensible.

core/src/prediction/InstrumentalPredictionStrategy.cpp
halflearned: Done.

jtibshirani:
Thanks for adding those tests. I agree they're hacky -- it might have been nice to pull out a method `InstrumentalPredictionStrategy#compute_excess_error` method now and unit test it. However this is an important fix and I'm okay merging it with the understanding that we'll improve these tests as part of the work on #288.

PR #329, opened 2018-11-13
Allow for a factor as the cluster variable

kendonB:
The documentation states: "cluster = Vector of integers or factors specifying which cluster each observation corresponds to."


r-package/grf/R/input_utilities.R
```diff
@@ -81,8 +81,8 @@ validate_clusters <- function(clusters, X) {
     clusters <- vector(mode="numeric", length=0)
   } else if (length(clusters) == 0) {
     clusters <- vector(mode="numeric", length=0)
-  } else if (!is.vector(clusters) | !all(clusters == floor(clusters))) {
-    stop("Clusters must be a vector of integers.")
+  } else if (!is.vector(as.numeric(clusters)) | !all(as.numeric(clusters) == floor(as.numeric(clusters)))) {
```
swager: This might be cleaner if we have two separate code paths for factors and non-factors... Something like if clusters is a factor, then just return the levels of the factor shifted by -1, else do the thing we did before. Currently this is rather hard to read.

jtibshirani:
@kendonB are you interested in continuing work on this PR? If you don't have time, we are happy to pick it up for you.

kendonB:
Apologies, I missed the comments in my emails. I have put in a parsimonious change where I remove the class at the beginning of the function (converting a `factor` to `integer`). The previous function then works the same as before.

r-package/grf/R/input_utilities.R
```diff
@@ -77,6 +77,9 @@ validate_seed <- function(seed) {
 }
 
 validate_clusters <- function(clusters, X) {
+  # This will convert factor to integer
+  clusters <- unclass(clusters)
```
jtibshirani: I find this use of `unclass` tricky to follow, and I haven't seen it used before in this context. Here is a simple suggestion that would work and is fairly readable:
```
...
} else if (!is.vector(clusters) || !(is.numeric(clusters) || is.factor(clusters))) {
     stop("Clusters must be a vector of integers or factors.")
} else if (is.numeric(clusters) && !all(clusters == floor(clusters))) {
    stop("Clusters cannot contain floating point values.")
}
```

kendonB: The above (and my most recent commit) does not work as `is.vector(iris$Species) == FALSE`. `is.vector` will only return `TRUE` if there're no attributes (including `class`).

Can I ask what the motivation for using `is.vector` is here? There might be a better way to exclude what you're trying to exclude. From what I can tell, `mode(clusters) == "numeric"`, should be sufficient to include only inputs that can be coerced to `integer` or `double`. This will also extend allowable inputs to dates and other reasonable inputs for this variable. `mode(clusters) == "numeric"` should exclude only `lists` (including `S3`) and more complicated classes (like `S4`).

jtibshirani: Thanks, I had overlooked this part. I agree, I think we can just remove the `is.vector` check here.

kendonB: OK - I have updated the PR with the `mode(clusters)` test

jtibshirani:
Merged, thanks for your contribution!

PR #334, opened 2018-11-19
Correctly weight clusters in ATE and APE functions.

swager:
When clusters are specified in grf, the default behavior is to give each cluster (not each sample) equal weight in training. However, until now, the ATE and APE functions did not respect this convention, and instead gave each sample equal weight. This PR fixes the issues, and brings the ATE and APE functions in line with the rest of the package.


r-package/grf/R/average_partial_effect.R
```diff
@@ -91,13 +111,13 @@ average_partial_effect = function(forest, calibrate.weights = TRUE, subset = NUL
   
   # Estimate variance using the calibration
   if (length(subset.clusters) == 0) {
-    cape.se = sqrt(mean((debiasing.weights * plugin.residual)^2) / length(subset.W.orig))
+    cape.se = sqrt(mean((debiasing.weights * plugin.residual)^2) / (length(subset.W.orig) - 1))
```
jtibshirani: I'm curious why we've added this `- 1` here?

r-package/grf/R/forest_summary.R
```diff
@@ -28,30 +28,47 @@
 #'
 #' @export
 test_calibration = function(forest) {
```
jtibshirani: Should we make a note here about clusters, as we did for the other methods?

r-package/grf/R/average_partial_effect.R
```diff
@@ -13,6 +18,7 @@
 #'               estimate the ATE. WARNING: For valid statistical performance,
 #'               the subset should be defined only using features Xi, not using
 #'               the treatment Wi or the outcome Yi.
+#' @param num.trees.for.variance Number of trees used to estimate Var[Wi | Xi = x].
```
jtibshirani: It's generally good to make unrelated changes like this in a separate PR. At the least, maybe we could add a note to the PR description noting these additional changes?

r-package/grf/R/average_partial_effect.R
swager: It's by analogy to <a href="https://en.wikipedia.org/wiki/Bessel%27s_correction">Bessel's correction</a> for variance estimation. Doesn't really make a difference in large samples, but still the right thing to do.

r-package/grf/R/forest_summary.R
swager: I think we also need a bigger PR where we explain how to use the `test_calibration` function. Maybe we can work on that discussion then? The way we handle clusters here is not consistent with the rest of the package (i.e., 1 cluster = 1 vote).

Issue #331, opened 2018-11-13
Nonzero imbalance.penalty prevents tree growth (and causes tuning to fail)


halflearned:
**Description of the bug**
There are certain scenarios in which having a nonzero `imbalance.penalty` causes most trees to split very few times. This obviously causes the performance of the predictions to deteriorate, and may also affect tuning.

**Steps to reproduce**
This [gist](https://gist.github.com/halflearned/8dea6ab05fa75cc7e01d7b9b779279a5) contains a minimal reproducible example. In it, we run `causal_forest` three times with different parameter specifications, and then compute the infeasible MSE of the CATE estimates. Their MSE is:

+ No tuning: ~0.014
+ Manually tuned `(mtry, min.node.side, alpha)`, and **`imbalance.penalty=0`**: ~0.012
+ Manually tuned `(mtry, min.node.side, alpha)`, and **`imbalance.penalty=1e-100`**: ~0.078

That is, setting `imbalance.penalty` to be a tiny number causes the performance to degrade by an unreasonable amount. 

Probing further, only about 3% of the trees had any splits at all in the third example, as opposed to 100% in the other two. So it looks like the positive `imbalance.penalty` is keeping the trees from splitting, and that's the root cause of the problem.

**Steps to fix?**
There aren't many places where `imbalance.penalty` shows up. My money is in [this](https://github.com/grf-labs/grf/blob/bc323178e53ddf453a388d0e3b12aa1a1b598328/core/src/splitting/InstrumentalSplittingRule.cpp#L205) set of conditions being somehow incorrect: for the setup in the gist, they almost always evaluate to TRUE, so that the variable `decrease` is never updated, and this causes the split not to happen.

Unfortunately, I don't totally understand what's going on in that piece of code, so I can't give a more precise answer.

**GRF version**
'development'

jtibshirani:
Thanks @halflearned for the great debugging and analysis of the issue. I will dig into `InstrumentalSplittingRule` to try to understand what's going wrong.

I don't think this should block our efforts to debug and solidify tuning -- perhaps we can remove `imbalance.penalty` it from the list of parameters considered for now.

jtibshirani:
I merged #335 to address the problem. @halflearned maybe you could take another look to confirm it fixes the issue you were seeing, and if so we can close out the issue?

halflearned:
Thanks for the PR, @jtibshirani. This does seem to fix the errors associated with `imbalance.penalty`, so I'll go ahead and clone this.

PR #335, opened 2018-11-19
Fix a bug in causal forest tuning where imbalance.penalty was being cast to a bool.

jtibshirani:
This bug was introduced in 0.9.6, after I performed a bad merge in #180.

Addresses #331.


r-package/grf/bindings/InstrumentalForestBindings.cpp
```diff
@@ -26,7 +26,7 @@ Rcpp::List instrumental_train(Rcpp::NumericMatrix input_data,
                               unsigned int ci_group_size,
                               double reduced_form_weight,
                               double alpha,
-                              bool imbalance_penalty,
+                              double imbalance_penalty,
```
swager: Good catch, thanks!

PR #338, opened 2018-11-21
corrected spelling of heteroskedasticity

ras44:


ras44:
I think that should take care of it.  Please let me know if you run into any issues.

PR #342, opened 2018-11-25
Disable roxygen comments processing for unavailable methods

jakobzeitler:
I had been, unfortunately, looking at outdated sample code for grf which included a use of "create_dot_body" which has been moved into a sub-method for plot.grf_tree since. As "create_dot_body" was listed in the 00index.html, I was under the assumption it should be available as a method even though it was not. Maybe this commit will help prevent confusion of that kind while still maintaining the commentary for development purposed. Would this comply with the GRF coding/documentation practices?

### DIFF
--- a/r-package/grf/R/plot.R
+++ b/r-package/grf/R/plot.R
@@ -1,8 +1,9 @@
-# Writes each node information 
-# If it is a leaf node: show it in different color, show number of samples, show leaf id
-# If it is a non-leaf node: show its splitting variable and splitting value
-# @param tree the tree to convert
-# @param index the index of the current node
+#' Writes each node information 
+#' If it is a leaf node: show it in different color, show number of samples, show leaf id
+#' If it is a non-leaf node: show its splitting variable and splitting value
+#' @param tree the tree to convert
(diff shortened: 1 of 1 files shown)


jtibshirani:
Thanks @jakobzeitler for catching this issue. I think that instead of disabling roxygen processing altogether, we should add the tag `@keyword internal` to prevent the function from being included in the package index, as suggested in https://cran.r-project.org/web/packages/roxygen2/vignettes/rd.html.

I was also wondering what outdated sample code you encountered (in case there's an example we should make sure to update somewhere)?

jakobzeitler:
Thanks for the response, Julie. I had been looking for the @keyword internal to solve this, thanks for pointing that out!

I was also looking for the example code I used, that included the "create_dot_body", but I have not found it, yet. (A repository wide search did not return anything, so it must have originated somewhere outside). I will let you know when it comes back to my mind.

jakobzeitler:
I have added the keyword to the plot.R. Do you have any other methods in mind that would require that keyword? I can add it there as well, if required.

PR #346, opened 2018-12-08
Acic18 script requirements

jakobzeitler:


experiments/acic18/script.R
```diff
@@ -3,6 +3,7 @@ set.seed(1)
 rm(list = ls())
 
 library(grf)
+if(packageVersion("grf") >= '0.10.2') {warning("This script requires grf 0.10.2 or higher");}
```
swager: I think this should be a '<' rather than '>='

swager: style nitpick: could you please put the warning on its own line, indented (and without the semicolon)

experiments/acic18/script.R
```diff
@@ -46,6 +47,9 @@ cf.raw = causal_forest(X, Y, W,
 varimp = variable_importance(cf.raw)
 selected.idx = which(varimp > mean(varimp))
 
+# Currently failing with error:
```
swager: if we have the package version check up top, I believe we don't need this comment

experiments/acic18/script.R
jakobzeitler: I totally agree.

jakobzeitler: Sure thing!

experiments/acic18/script.R
jakobzeitler: Yes, that was just my personal note!

jakobzeitler:
Hope these changes are ok.

Issue #272, opened 2018-08-04
Support for adding more trees to a trained forest.


jtibshirani:
A user may train a forest, then realize that more trees are needed (for better predictions, tighter confidence intervals, etc.) Instead of starting from scratch and training a completely new forest, it would be great if the user was able to grow some number of additional trees, and add these trees to the original forest.

Proposal for the API: each forest method (`regression_forest`, `causal_forest`, etc.) can take in an optional parameter `existing.forest` (or something similar). The final forest will contain trees from the original forest, as well as newly grown trees. The methods should error if the provided forest isn't of the correct type.

It would also be interesting to experiment with using this functionality during tuning: if parameters haven't changed too much, we could include the trees used for tuning when creating the final forest, and so need to grow fewer trees in the final training step. I think it would be misleading to do this by default, but may be interesting to explore as part of an 'optimized' option.

susanathey:
If we want to parallelize building a forest in a large dataset (parallelize across machines, for example), which seems like a common scenario for larger datasets and a natural one since the forest routine is naturally parallelizable, it would be important that we can union forests rather than take in an existing forest and add to it.  Can I suggest creating a union function, and then if we want you can do as you suggest above, but that would call the union function?

jtibshirani:
I was initially hesitant to go with a `union` function, because in the common case of adding more trees to an existing forest (on a single machine)
- there would be an extra round of serialization between C++ and R, which we know to be expensive
- now that we include OOB predictions by default as part of training, a `union` method would need to rerun OOB prediction, which is also wasteful

The use case around combining forests makes a lot of sense, though -- I'd be happy to support both in R (and still share quite a bit of common merging code under the hood in C++).

halflearned:
In #347 I submitted a first attempt at this. As it turns out, the basic functionality was easy to implement. A basic example looks like

```{r}
rf1 <- regression_forest(X, Y, num.trees=100, compute.oob.predictions=FALSE) # OOB preds will be ignored anyway
rf2 <- regression_forest(X, Y, num.trees=100, compute.oob.predictions=FALSE)
big_rf <- join_forests(list(rf1, rf2))  # Has num.trees = 200
```
The syntax is the same regardless of the forest type.

Some considerations:

1. As @jtibshirani commented above, forest concatenation happens on the C++ side, so to concatenate R-side forests we need an extra round of serialization.

[...]

PR #347, opened 2018-12-09
Union of forests

halflearned:
First attempt at union of forests functionality. See additional considerations  at https://github.com/grf-labs/grf/issues/272.

**C++ side**
+ [forest/Forest.h] Added static function `join_forests` to `Forest` class.

**R side**
+ [r-package/grf/bindings] Function `cpp_join_forests` connects R and C++ sides
+ [r-package/grf/R/join] Function `join_forests` takes forests of arbitrary type and concatenates into one larger forest with same attributes.


r-package/grf/R/join.R
```diff
@@ -0,0 +1,62 @@
+#' Concatenates a list of identical forests that were grown using the same data.
```
jtibshirani: A couple comments:
- I think the file should have the same name as the method.
- Don't mean to nitpick, but maybe we could say 'combined' or 'merged together'? Concatenated usually refers to chaining items together into a list.
- Until we've figured out how to merge forests trained on different partitions of the dataset (point 2 in https://github.com/grf-labs/grf/issues/272#issuecomment-445524760), maybe we can hold off on saying "Useful for training forests in a distributed manner"?

r-package/grf/bindings/ForestBindings.cpp
```diff
@@ -0,0 +1,34 @@
...
+#include "commons/DefaultData.h"
+#include "commons/SparseData.h"
+#include "forest/ForestOptions.h"
+#include "forest/Forest.h"
+#include "RcppUtilities.h"
+#include "serialization/ForestSerializer.h"
+
+
+// [[Rcpp::export]]
+Rcpp::List cpp_join_forests(const Rcpp::List forest_objects,
```
jtibshirani: Does this need the `cpp` prefix? We don't use it in other bindings files.

core/src/forest/Forest.h
```diff
@@ -40,6 +40,8 @@ class Forest {
   const std::vector<std::shared_ptr<Tree>>& get_trees() const;
   const size_t get_num_variables() const;
 
+  static Forest join(std::vector<std::shared_ptr<Forest>>& forests);
```
jtibshirani: It would be good to make this a `const` vector (I've been trying to encourage const correctness as much as possible).

r-package/grf/bindings/ForestBindings.cpp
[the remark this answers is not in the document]
halflearned: Right. I also hesitated about this.

core/src/forest/Forest.h
halflearned: Great point!

core/src/forest/Forest.cpp
```diff
@@ -51,6 +51,20 @@ Forest::Forest(const std::vector<std::shared_ptr<Tree>>& trees,
   observations(observations),
   num_variables(num_variables) {}
 
+Forest Forest::join(std::vector<std::shared_ptr<Forest>>& forests) {
```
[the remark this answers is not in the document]
halflearned: Ha -- I had [dplyr::join](https://dplyr.tidyverse.org/reference/join.html) in mind. No problem, will rename it.

r-package/grf/R/join.R
```diff
@@ -0,0 +1,62 @@
...
+  first_forest <- forest_list[[1]]
+  data <- create_data_matrices(first_forest$X.orig, first_forest$Y.orig)
+  
+  big_forest <- cpp_join_forests(forest_list, data$default, data$sparse)
+  
+  big_forest[["ci.group.size"]] <- first_forest$ci.group.size
+  big_forest[["X.orig"]] <- first_forest$X.orig
+  big_forest[["Y.orig"]] <- first_forest$Y.orig
+  big_forest[["clusters"]] <- first_forest$clusters
+  big_forest[["min.node.size"]] <- first_forest$min.node.size
```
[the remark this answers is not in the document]
jtibshirani: I think it should default to `TRUE` (just pinged @swager about this and he is on board too).

r-package/grf/bindings/ForestBindings.cpp
jtibshirani: I think this should be fine because they're namespaced differently (one is nested under `Forest::`)?

r-package/grf/R/join.R
```diff
@@ -0,0 +1,62 @@
...
+validate_forest_list <- function(forest_list) {
+  
+  if (length(forest_list) == 0) {
+    stop("Length of argument 'forest_list' must be positive.")
+  }
+  
+  first_forest <- forest_list[[1]]
+  if (!is(rf1, "grf")) {
+    stop("Argument 'forest_list' must be a list of grf objects. 
+           Be sure to use 'list(forest1, forest2), not 'c(forest1, forest2)'.")
```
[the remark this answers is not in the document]
jtibshirani: Just went to try this out and I think this should be `first_forest` instead of `rf1`.

r-package/grf/NAMESPACE
```diff
@@ -23,6 +23,7 @@ export(custom_forest)
 export(get_sample_weights)
 export(get_tree)
 export(instrumental_forest)
+export(merge_forests)
```
jtibshirani: Minor comment, but these should be alphabetical.

r-package/grf/R/merge_forests.R
```diff
@@ -0,0 +1,78 @@
...
+  big_forest[["min.node.size"]] <- first_forest$min.node.size
+  
+  class(big_forest) <- class(first_forest)
+  
+  if (compute.oob.predictions) {
+    oob.pred <- predict(big_forest)
+    big_forest[["predictions"]] <- oob.pred$predictions
+    # Must include checks here because some big_forest types may have this 
+    # method not yet implemented
+    if (!is.null(big_forest["debiased.error"])) {
```
jtibshirani: Could you explain a little more about this check?

r-package/grf/R/merge_forests.R
```diff
@@ -0,0 +1,78 @@
...
+  first_forest <- forest_list[[1]]
+  data <- create_data_matrices(first_forest$X.orig, first_forest$Y.orig)
+  
+  big_forest <- merge(forest_list, data$default, data$sparse)
+  
+  big_forest[["ci.group.size"]] <- first_forest$ci.group.size
+  big_forest[["X.orig"]] <- first_forest$X.orig
+  big_forest[["Y.orig"]] <- first_forest$Y.orig
+  big_forest[["clusters"]] <- first_forest$clusters
+  big_forest[["min.node.size"]] <- first_forest$min.node.size
```
jtibshirani: Is `min.node.size` actually returned as part of the forest object?

r-package/grf/R/merge_forests.R
```diff
@@ -0,0 +1,78 @@
...
+merge_forests <- function(forest_list, compute.oob.predictions=TRUE) {
+  
+  validate_forest_list(forest_list)
+  
+  first_forest <- forest_list[[1]]
+  data <- create_data_matrices(first_forest$X.orig, first_forest$Y.orig)
+  
+  big_forest <- merge(forest_list, data$default, data$sparse)
+  
+  big_forest[["ci.group.size"]] <- first_forest$ci.group.size
```
jtibshirani: Maybe we could validate that `ci.group.size` is the same across all forests? This could lead to subtle issues, and it is easy to protect against.

r-package/grf/tests/testthat/test_merge_forests.R
```diff
@@ -0,0 +1,40 @@
...
+  X = matrix(rnorm(n*p), n, p)
+  Y = X[,1] * rnorm(n)
+  W = X[,2] > 0
+  c.forest1 = causal_forest(X, Y, W, compute.oob.predictions = FALSE, num.trees = 10)
+  c.forest2 = causal_forest(X, Y, W, compute.oob.predictions = FALSE, num.trees = 10)
+  
+  # Join the forests together. 
+  big_rf = merge_forests(list(c.forest1, c.forest2))
+  
+  # Result is also a regression_forest of the same class
```
jtibshirani: `regression_forest` -> `causal_forest`

r-package/grf/R/merge_forests.R
halflearned: Interestingly, it's returned as part of `regression.forest`, but not `causal.forest`.

r-package/grf/R/merge_forests.R
halflearned: I had in mind forests that output only "predictions", but not the other statistics. 

However, it seems that we don't really need that, or at least that this isn't the fix we need. For example, what should we do about `quantile_forests`?

```
X <- matrix(runif(1000), 500, 2)
Y <- matrix(runif(500))
rf <- grf::quantile_forest(X, Y)
pred <- predict(rf) 
class(pred)  # "matrix"
pred$predictions # Error: $ operator is invalid for atomic vectors 
```

r-package/grf/R/merge_forests.R
halflearned: Should this be an optional feature? Something like `ensure.forest.compatibility=TRUE`?

r-package/grf/R/merge_forests.R
jtibshirani: I think we can get rid of it, it's inconsistent that we're adding this to the returned object.

r-package/grf/R/merge_forests.R
jtibshirani: This seems useful to me all the time, so I'm not sure a flag would make sense -- I can't see a case where you would want to train with a different `ci.group.size` than the one you use to predict.

r-package/grf/R/merge_forests.R
jtibshirani: That seems like a bug to me, since the prediction objects should have the same structure across the different forest types (quantile forests should still contain the `predictions` subfield, and just omit `debiased.error` etc.). I would vote to not worry about this in the current PR, and plan to fix the discrepancy in predictions formats in a follow-up PR.

adeldaoud:
This functionality of adding more trees to an existing grf is useful. Is this function join_forests going to be released in version 10.3? Happy to test if it is available in the development version.

jtibshirani:
Hi @adeldaoud, yes this feature will be made available in our next release. You're welcome to try it out earlier by installing the development version from source. Here is a brief description of how to do so: https://github.com/grf-labs/grf/blob/master/DEVELOPING.md#working-with-the-code

adeldaoud:
Hi @jtibshirani , trying to install the development version but no success. Can you please let me know what is wrong with my code below:

```
library(devtools)

# try this function
install_github("grf-labs/grf/master/r-package/build_package.R") # does not work
Error: HTTP error 404.
  Not Found

  Rate limit remaining: 50/60
  Rate limit reset at: 2019-01-18 23:28:05 UTC

# try different code
install.packages("https://raw.github.com/grf-labs/grf/master/r-package/build_package.R", repos = NULL, type = "source") # does not work either

Installing package into ‘C:/Users/add893/Documents/R/win-library/3.5’
(as ‘lib’ is unspecified)
trying URL 'https://raw.github.com/grf-labs/grf/master/r-package/build_package.R'
Content type 'text/plain; charset=utf-8' length 955 bytes
downloaded 955 bytes

Error in getOctD(x, offset, len) : invalid octal digit
In R CMD INSTALL
Warning in install.packages :
  installation of package ‘C:/Users/add893/AppData/Local/Temp/RtmpSIwaV4/downloaded_packages/build_package.R’ had non-zero exit status
```

jtibshirani:
Hi @adeldaoud, we generally try to keep each github issue or PR focused on the original topic. If you need help debugging or have a question about installation, it would be great if you could send an email or open a new github issue.

Issue #359, opened 2018-12-28
Calls to unordered_map cause DefaultPredictionStrategy to be very slow


halflearned:
[This](https://github.com/grf-labs/grf/blob/d4333174f876c5eca992d0e03a5fac450968822f/core/src/prediction/collector/SampleWeightComputer.cpp#L55) line is slow. 

Example: the c++ side of a call to local linear forests with n=10000 observations, p=10 regressors and 1000 trees takes about 4m30s. A quick profiling with xcode reveals that about *half* of the time is spent in this line.

<img width="1280" alt="before" src="https://user-images.githubusercontent.com/20155021/50499211-04b19b80-0a2f-11e9-8443-6dbe13d4edab.png">

The culprit seems to be the `std::unordered_map`. Its poor performance is somewhat [documented](https://youtu.be/fHNmRkzxHWs?t=2719) and we should avoid using it.

**Scope**
Affects forests that use `DefaultPredictionStrategy`. Right now, these are quantile and local_linear forests.

**GRF version**
0.10.2

---

@jtibshirani, @davidahirshberg, @swager and I have discussed this problem offline. I'll soon submit a PR for this.

halflearned:
Profiling with more recent code showed that the bottleneck is (surprisingly, this time!) _not_ `unordered_map`. Removing `unordered_map` and substituting it with a pair of `std::vector` yielded very little performance improvement (see below). A vector of pairs yielded the same.

```
library(grf)
n <- 2000
p <- 10
X <- matrix(rnorm(n * p), n, p)
Y <- pmax(X[, 1], 0) + X[, 2] + pmin(X[, 3], 0) + rnorm(n)
tau.forest <- ll_regression_forest(X, Y, num.threads = 1)
tau.hat.oob <- predict(tau.forest, num.threads = 1)
```

<img width="1680" alt="Screen Shot 2019-12-15 at 4 31 39 PM" src="https://user-images.githubusercontent.com/20155021/71499554-01621300-282f-11ea-8b11-e93ead150a71.png">

---

That being the case, I'll close this for now. I'm willing to work on this if needed, but it would be a sizable PR and I feel we may have lower-hanging fruit elsewhere first.

PR #360, opened 2018-12-28
Improving SampleWeightComputer performance

halflearned:
This PR contains a simple tweak to `SampleWeightComputer::compute_weights` to deal with  #359. 

**Main changes**

The construction of `weights_by_sample` now happens in two steps.
1. [Create](https://github.com/grf-labs/grf/blob/b17f421fd748da7336037a04de5486e82a99b058/core/src/prediction/collector/SampleWeightComputer.cpp#L30) and [populate](https://github.com/grf-labs/grf/blob/f8b0eaf1d8280ec4d4008393ddbaf836d1449355/core/src/prediction/collector/SampleWeightComputer.cpp#L45) a `std::vector` of  size n containing "raw" weights. These are unnormalized neighbor weights. Entries associated with samples that happen to never fall in the same leaf end up with weight zero.
2. [Populate](https://github.com/grf-labs/grf/blob/b17f421fd748da7336037a04de5486e82a99b058/core/src/prediction/collector/SampleWeightComputer.cpp#L52) `weights_by_sample` with normalized weights. 

**Results**

Because we only try to access an unordered_map key once per neighbor, performance losses due to unordered_map are mitigated. For the example in #359, the runtime drops from ~4m30s to ~2m30s.
 
screen shot 2018-12-28 at 00 08 59

For other (n, p) combinations (now with 2000 trees):

**Before (with 2000 trees)**

  |  n   | p   |   time (seconds)
|:--:|:--:|:--:|
 1000  | 10   | 4.78839
 1000  | 50  |  5.40171
 2000  | 10  | 18.8154
 2000  | 50  | 17.3813
 5000  | 10  | 103.222
 5000  | 50  | 108.191

**After (with 2000 trees)**

  n |   p |     time  (seconds)
|-----:|----:|---------:|
 1000 |  10 |  3.05668 |
1000 |  50 |  3.34233 |
 2000 |  10 | 10.535   |
 2000 |  50 | 10.2441  |
5000 |  10 | 63.2606  |
5000 |  50 | 56.613   |

----

**Possible next steps**

1. Creating an entire vector of size n may be expensive in terms of memory, so I'm taking suggestions for better solutions. That said: this vector is destroyed once `weights_by_sample` is populated and the function ends.

2. (Future PR?) Calls to unordered_map still take up about 5% of the runtime. It's possible that we could decrease that a bit more by replacing `unordered_map` by something like

```{c}
using SampleWeights = std::vector<std::pair>
```

or perhaps

```{c}
class SampleWeights:
public:
    SampleWeights(std::vector<double> raw_vector); 
    std::vector<size_t> neighbor_ids;
    std::vector<double> neightbor_weights;
```

I think this will also obviate the entire `SampleWeightsComputer` class. This seems like a more invasive change, so I may reach out to @jtibshirani et al again offline to discuss this, at a convenient time.

-----

**Known errors**

~There is a [test](https://github.com/grf-labs/grf/blob/f8b0eaf1d8280ec4d4008393ddbaf836d1449355/r-package/grf/tests/testthat/test_local_linear_forest.R#L162) that fails after the change. It's a bit strange that it's the only test that fails, but I'll investigate it over the weekend.~

Found bug and fixed it. All tests are passing now.


jtibshirani:
Thanks @halflearned, this is a great evidence-based approach to optimization! There are a couple changes we could profile before setting on a final approach:
1. We could remove the `unordered_map` completely, and just pass this vector of weights to the prediction strategies. This doesn't extend the lifetime of the vector much, as it will be freed as soon as we are finished processing a single test datapoint.
2. Using the raw vector everywhere will result in additional iterations over this collection of size n. So after the weights are collected, we could try switching to a `std::vector<std::pair>` as you suggested. I think `QuantilePredictionStrategy` will need to be reworked a bit, so it would be best to prototype this approach only local linear forests. I suspect this change won't make a big difference compared to just doing (1), but it could be worth a try.

jtibshirani:
> Creating an entire vector of size n may be expensive in terms of memory, so I'm taking suggestions for better solutions.

It seems okay to create a single vector with size equal to the number of examples, since we copy vectors of this size in other places (for example in the R forest methods, when preparing the training data matrix). But, it may be beneficial to avoid creating a new vector in each call to `compute_weights` to prevent excessive allocations. We could try profiling a modification of approach (1), where we only create a single shared vector, which gets zero-filled and reused across different calls to `compute_weights`. This is similar to the approach taken in the `*SplittingRule` classes, where various arrays are reused across calls to `find_best_split`.

halflearned:
Retracting this PR for now -- a lot has changed since this PR was submitted, I'll likely rework this from scratch at another time.

Issue #232, opened 2018-04-14
De-hackify the confidence interval implementation.


jtibshirani:
- The serialized forest should store what `ci.group.size` was used in training the forest, so that it can be reused during prediction. Currently, we force the client to pass in the same `ci.group.size` during both training and prediction, which is quite fragile.
- To indicate that no confidence intervals should be calculated, the `ci.group.size` is set to 1. This is non-intuitive -- there should be a more robust way of turning off confidence intervals.

PR #361, opened 2018-12-28
Clean up the confidence interval framework.

jtibshirani:
* Store ci_group_size as part of the forest to avoid a mismatch between training and prediction.
* Add an estimate_variance option to prediction, to replace the fragile check ci_group_size > 1.
* Move the check for consistent ci_group_size during merges into C++.
* Use size_t instead of uint for ci_group_size.

Addresses #232.

### DIFF
--- a/r-package/grf/R/regression_forest.R
+++ b/r-package/grf/R/regression_forest.R
@@ -221,12 +221,6 @@ predict.regression_forest <- function(object, newdata = NULL,
 
     num.threads = validate_num_threads(num.threads)
 
-    if (estimate.variance) {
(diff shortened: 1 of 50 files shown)


r-package/grf/R/regression_forest.R
```diff
@@ -221,12 +221,6 @@ predict.regression_forest <- function(object, newdata = NULL,
 
     num.threads = validate_num_threads(num.threads)
 
-    if (estimate.variance) {
-        ci.group.size = object$ci.group.size
```
swager: 🌳____ 🌳

Issue #128, opened 2017-10-16
Sample weights


patr1ckm:
Just curious if there was bandwidth/interest in adding sample weights for training? If not, would you be open to a PR?

Thanks for your time!

swager:
Yes there's been a fair amount of interest, but we haven't had the bandwidth for it yet. We'd certainly be open to a PR.

jtibshirani:
A quick update -- the following two PRs have merged, which add basic support for sample weighting: #376 (for regression forests), #418 (for causal and instrumental forests). This feature is currently marked 'experimental', and will be released in 0.10.3.

I'll leave this issue open for now as we test it further. A remaining open question is if we want to incorporate sample weights into tree splitting (in addition to prediction, as we already do).

swager:
Weighting now appears to be stable, so I'm closing this issue. Some remaining considerations are tracked in a new (non 1.0 release) ticket, #473.

Thanks everyone for your help with this!

PR #376, opened 2019-02-20
Adding Sample Weights for Training

davidahirshberg:
This is a partial implementation of sample weights for training GRFs. The proposed interpretation of the weights is that they assumed to come from a deterministic function w_i = w(X_i) of the conditioning variable, and we solve the local least squares problem argmin_{theta,nu} || sum_i alpha_i(x) w_i psi_{theta,nu}(O_i) ||, a sample-weighted version of  Equation 2 of the GRF paper, with the tree-based weights alpha_i unchanged.

Currently, this is implemented only for RegressionForest. In that context, it amounts to taking weighted averages within each leaf in the forest. There is a small experiment in which we use a regression forest to impute observations that are missing at random, and inverse propensity weighting when we train
improves both MSE and coverage when predicting the missing units.

Addresses #128.

### DIFF
--- a/core/src/prediction/RegressionPredictionStrategy.cpp
+++ b/core/src/prediction/RegressionPredictionStrategy.cpp
@@ -97,15 +97,20 @@ PredictionValues RegressionPredictionStrategy::precompute_prediction_values(
       continue;
     }
 
-    std::vector<double>& averages = values[i];
-    averages.resize(1);
-
     double sum = 0.0;
     double weight = 0.0;
     for (auto& sample : leaf_node) {
       sum += data->get_weight(sample) * data->get_outcome(sample);
       weight  += data->get_weight(sample);
     }
+
+    // if total weight is zero, treat the leaf as empty
+    if (weight == 0) {
+      continue;
+    }
+
+    std::vector<double> &averages = values[i];
+    averages.resize(1);
     averages[OUTCOME] = sum / weight;
   }
 
--- a/core/src/prediction/RegressionPredictionStrategy.cpp
+++ b/core/src/prediction/RegressionPredictionStrategy.cpp
@@ -104,8 +104,8 @@ PredictionValues RegressionPredictionStrategy::precompute_prediction_values(
       weight  += data->get_weight(sample);
     }
 
-    // if total weight is zero, treat the leaf as empty
-    if (weight == 0) {
+    // if total weight is very small, treat the leaf as empty
+    if (std::abs(weight) <= 1e-16) {
       continue;
     }
 
--- a/core/src/prediction/RegressionPredictionStrategy.cpp
+++ b/core/src/prediction/RegressionPredictionStrategy.cpp
@@ -100,11 +100,13 @@ PredictionValues RegressionPredictionStrategy::precompute_prediction_values(
     std::vector<double>& averages = values[i];
     averages.resize(1);
 
-    double average = 0.0;
+    double sum = 0.0;
+    double weight = 0.0;
     for (auto& sample : leaf_node) {
-      average += data->get_outcome(sample);
+      sum += data->get_weight(sample) * data->get_outcome(sample);
+      weight  += data->get_weight(sample);
     }
-    averages[OUTCOME] = average / leaf_node.size();
+    averages[OUTCOME] = sum / weight;
   }
 
   return PredictionValues(values, num_leaves, 1);
--- a/core/src/prediction/RegressionPredictionStrategy.cpp
+++ b/core/src/prediction/RegressionPredictionStrategy.cpp
@@ -70,7 +70,7 @@ std::vector<double> RegressionPredictionStrategy::compute_variance(
 
   // This is the amount by which var_between is inflated due to using small groups
   double group_noise = (var_total - var_between) / (ci_group_size - 1);
-  
+
   // A simple variance correction, would be to use:
   // var_debiased = var_between - group_noise.
   // However, this may be biased in small samples; we do an objective
@@ -117,7 +117,7 @@ PredictionValues RegressionPredictionStrategy::precompute_prediction_values(
   return PredictionValues(values, num_leaves, 1);
 }
 
-std::vector<double> RegressionPredictionStrategy::compute_debiased_error(
+std::vector<std::pair<double, double>>  RegressionPredictionStrategy::compute_error(
     size_t sample,
     const std::vector<double>& average,
     const PredictionValues& leaf_values,
@@ -140,9 +140,14 @@ std::vector<double> RegressionPredictionStrategy::compute_debiased_error(
   }
 
   if (num_trees <= 1) {
-    return { NAN };
+    return { std::make_pair<double, double>(NAN, NAN) };
   }
 
   bias /= num_trees * (num_trees - 1);
-  return { mse - bias };
+
+  double debiased_error = mse - bias;
+
+  auto output = std::pair<double, double>(debiased_error, bias);
+  return { output };
 }
+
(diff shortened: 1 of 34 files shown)


r-package/grf/bindings/RegressionForestBindings.cpp
```diff
@@ -47,6 +48,9 @@ Rcpp::List regression_train(Rcpp::NumericMatrix train_matrix,
 
   Data* data = RcppUtilities::convert_data(train_matrix, sparse_train_matrix);
   data->set_outcome_index(outcome_index - 1);
+  if(sample_weight_index - 1 != data->get_num_cols()) {
```
jtibshirani: I find this approach to indicating whether sample weights are present to be a bit tricky/ fragile. Ideally we'd use `nonstd::optional<size_t>` class, but I don't think it'll work in an Rcpp exported method. Maybe we could use a `size_t *` here, and set it to `NULL` if sample weights aren't used (this is a common way of expressing 'optional' in C).

davidahirshberg: I don't like it either, but it was the best I could come up with. The other idea I thought might be reasonable is to make the index signed and pass -1 as missing, but it seemed odd to have one signed index. The possibly-null pointer approach would be pretty awkward on the R side.

 or the C++ side but unnatural on the other. Passing an R list of parameters instead of multiple arguments is another.

core/src/prediction/RegressionPredictionStrategy.cpp
```diff
@@ -100,11 +100,13 @@ PredictionValues RegressionPredictionStrategy::precompute_prediction_values(
     std::vector<double>& averages = values[i];
     averages.resize(1);
 
-    double average = 0.0;
+    double sum = 0.0;
+    double weight = 0.0;
     for (auto& sample : leaf_node) {
-      average += data->get_outcome(sample);
+      sum += data->get_weight(sample) * data->get_outcome(sample);
+      weight  += data->get_weight(sample);
```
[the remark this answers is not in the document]
davidahirshberg: It's possible. I don't think any kind of bad numerical behavior can happen with positive weights, as the numerator will be roughly zero if the denominator is. I think the most natural solution is to add a positivity constraint to the validator in the R code.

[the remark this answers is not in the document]
jtibshirani: I thought about this more, and I think it makes the most sense to skip this leaf node if the total weight is too close to zero. This matches what we do when the leaf node is empty.

I like this behavior more than enforcing positivity, because I could see someone using an estimation procedure to determine sample weights. Some of the weights might end up being very close to zero, and it would be disruptive/ annoying if we rejected them.

r-package/grf/bindings/RegressionForestBindings.cpp
jtibshirani: Got it, in that case could we introduce a boolean `contains_sample_weights`? It is quite clunky, but I think safety + clarity is more important than elegance.

davidahirshberg: Good call.

core/src/prediction/RegressionPredictionStrategy.cpp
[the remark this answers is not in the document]
davidahirshberg: I meant a positivity check in the sense that we'd require that weights are > 0, so we wouldn't be rejecting weights for being small. If all weights are nonnegative, I don't think we'll have much trouble with floating point error with this, but I can also switch to a more stable algorithm, e.g. the one described in Section 4 [here](http://people.ds.cam.ac.uk/fanf2/hermes/doc/antiforgery/stats.pdf).

I have to push back a bit here on the idea of zeroing out leaves with small total weight. Inverse propensity weighting to minimize mean squared error when imputing missing observations, as in sample_weighting_examples/simple_regression_forest.R, involves weights that will often be near zero in regions in which we have plenty of data and can get a good regression estimate. A small weight w(x) just means that we think it's unlikely that we'll have to predict a value at x, so we don't want to go out of our way to get a good regression estimate there. But by ignoring low-weight leaves, we'd be ignoring the leaves that best estimate E[Y|X=x] at x where w(x') is small for x' near x.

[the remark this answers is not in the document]
davidahirshberg: @jtibshirani has informed me that 'very close to zero' means equal to zero in the floating point sense, so we are in agreement. New version allow zeros weights and ignore leaves with total weight zero.

r-package/grf/R/input_utilities.R
```diff
@@ -191,21 +191,33 @@ validate_newdata <- function(newdata, X) {
   validate_X(newdata)
 }
 
+validate_sample_weights = function(sample.weights, X) { 
+    if(!is.null(sample.weights)) {
+        if(length(sample.weights) != nrow(X)) { 
+            stop("sample.weights has incorrect length")
+        }
+        if(any(sample.weights < 0)) {
```
jtibshirani: 👍

core/src/prediction/RegressionPredictionStrategy.cpp
```diff
@@ -97,14 +97,21 @@ PredictionValues RegressionPredictionStrategy::precompute_prediction_values(
...
     }
-    averages[OUTCOME] = average / leaf_node.size();
+
+    // if total weight is zero, treat the leaf as empty
+    if (weight == 0) {
```
jtibshirani: I alluded to this above, but we can't actually use `==` when checking if a floating point is equal to zero. We'll need to do something like the following: `if (std::abs(weight) < 1e-10) { ... }`.

This got me thinking, should we ensure that the weights sum to 1? This will protect against overflow in calculating `sum` and `weight`. Maybe we can think about this for a follow-up.

r-package/grf/tests/testthat/test_regression_forest.R
```diff
@@ -129,3 +129,84 @@ test_that("regression forests with a very small imbalance.penalty behave similar
...
+    
+    forest = regression_forest(X[w,], Y[w])
+    forest.weighted = regression_forest(X[w,], Y[w], sample.weights)
+
+    ipw.mse.forest = sum((predict(forest,X) - Y)^2)
+    ipw.mse.forest.weighted = sum((predict(forest.weighted,X) - Y)^2)
+
+    expect_true(ipw.mse.forest.weighted < ipw.mse.forest)
+})
+
```
jtibshirani: Small comment: a bunch of extra newlines.

r-package/grf/R/input_utilities.R
```diff
@@ -191,21 +191,33 @@ validate_newdata <- function(newdata, X) {
...
+
 coerce_honesty_fraction <- function(honesty.fraction) {
   if(is.null(honesty.fraction)) {
     return(0)
   }
   honesty.fraction
 }
 
-create_data_matrices <- function(X, ...) {
+create_data_matrices <- function(X, ..., sample.weights=NULL) {
```
jtibshirani: I don't think this named argument `sample.weights` is ever used. From my testing, the `...` is just capturing `sample.weights` everywhere it's passed in. This works because `cbind` can handle a `NULL` argument.

I would be fine just removing this named argument, and relying on this behavior of `cbind`. I checked this works for both the normal and sparse versions of `cbind`.

davidahirshberg: It's used if you pass by keyword, which I apparently intended to do but did not. Preference for passing by keyword or getting rid of the keyword argument?

davidahirshberg: The new version passes by keyword, but doesn't do any such preprocessing.

jtibshirani: 👍 works for me, as long as we're actually using the keyword argument.

Issue #373, opened 2019-02-04
Make sure to validate the observation objects.


jtibshirani:
Currently, we don't validate the user-provided observation objects in R like `Y` and `W`. We should perform at least the following validation:
- Make sure that they are numeric vectors. Currently the observations are allowed to be matrices, which has lead to surprising bugs like #369.
- Ensure that there are no `NA` observations. We don't yet support `NA` values and we suspect that this could be causing the issue seen in #368, where grf can produce `NA` for all predictions.

halflearned:
Thanks, @jtibshirani! 

Related: when the number of rows is zero, the error messages are not very informative. Here's an example.

```
library(grf)

X <- matrix(rnorm(0), nrow = 0, ncol=10)  # Maybe by mistake these matrices were constructed with zero rows
Y <- matrix(rnorm(0), nrow = 0, ncol=1)
W <- matrix(rnorm(0), nrow = 0, ncol=1)

causal_forest(X, Y, W)
```

[This](https://github.com/grf-labs/grf/blob/996f3b3fd4d89761b7b9ed6b0da31dbbba6895ac/core/src/forest/ForestTrainer.cpp#L45) exception will be thrown, triggering the message

```r
  The honesty fraction is too close to 1 or 0, as no observations will be sampled.
```

Sounds like a corner case, but I've encountered this message several times.

ras44:
@jtibshirani I hope this PR is helpful in addressing this issue.  Please let me know if you have any comments!

jtibshirani:
Thanks a lot @ras44, I just had the chance to review the PR.

@halflearned that is confusing, maybe we could open a PR to add a clearer message or file a new issue?

jtibshirani:
Closed by #378.

PR #378, opened 2019-03-15
Add better validation for the input data.

ras44:


ras44:
I've provided an initial start to addressing #373.  Looking forward to your feedback 👍

r-package/grf/R/input_utilities.R
```diff
@@ -11,6 +11,36 @@ validate_X <- function(X) {
...
+  }
+}
+
+validate_W <- function(W) {
+  if(!is.vector(W) || !is.numeric(W)) {
+    stop(paste("The treatment W must be numeric vector. GRF does not", 
+               "currently support non-numeric or non-vector treatments."))
+  }
+
+  if(length(unique(W))!=2) {
```
jtibshirani: I think we should omit this check, as we actually allow for continuous treatments.

Since `validate_Y` and `validate_W` will be the same after removing the check, I think it'd be nice to have a single function `validate_observations(...)` that takes all the observation vectors and validates each one.

r-package/grf/R/input_utilities.R
```diff
@@ -11,6 +11,36 @@ validate_X <- function(X) {
   if (inherits(X, "Matrix") & !(inherits(X, "dgCMatrix"))) {
       stop("Currently only sparse data of class 'dgCMatrix' is supported.")
   }
+  
+  if (!all(!is.na(X))){
```
jtibshirani: Small comment, but I find `if (any(is.na(X)) { ... }` clearer (avoids a double negative).

r-package/grf/R/causal_forest.R
```diff
@@ -133,6 +133,8 @@ causal_forest <- function(X, Y, W,
                           num.fit.reps = 50,
                           num.optimize.reps = 1000) {
     validate_X(X)
+    validate_Y(Y)
+    validate_W(W)
     if(length(Y) != nrow(X)) { stop("Y has incorrect length.") }
```
jtibshirani: It would be nice to move these checks into the validate methods as well, to unify all the validation logic in one place.

r-package/grf/tests/testthat/test_validation.R
```diff
@@ -0,0 +1,67 @@
+library(grf)
+library(testthat)
+set.seed(1234)
+
+test_that("example data passes validation"){
```
jtibshirani: I don't think we need this test method, as the behavior it tests (good data passing validation) is extensively covered by other tests.

r-package/grf/R/causal_forest.R
ras44: @jtibshirani Thanks for the suggestion.  Did you mean move the `if(length(Y))...` and `if(length(W)...` tests into the validate_Y/W functions?  That would require passing X, or its length, to the validate_Y/W functions.  Is that ok?  Another alternative would be to create a wrapper for these 5 lines as `validate_XYW(X,Y,W)`.  Let me know which you would prefer.  Thanks.

r-package/grf/R/input_utilities.R
ras44: agreed!  This will be addressed.

r-package/grf/tests/testthat/test_validation.R
ras44: removed.

r-package/grf/R/causal_forest.R
ras44: I've modified this to use `validate_YX`, `validate_WX` which then call `validate_Y` and `validate_W` to keep the logic separate.  I also added a `validate_ZX` and `validate_Z` for the instrumental_forest.  This way it offers both YX and Y-only validation.

ras44:
@jtibshirani I've added commits that should address all of the comments.  I also added validation for all other forests I could find 👍

r-package/grf/R/input_utilities.R
```diff
@@ -11,6 +11,55 @@ validate_X <- function(X) {
...
+  if(length(W) != nrow(X)) { stop("length(W) does not equal nrow(X).") }
+}
+
+validate_W <- function(W) {
+  if(!is.vector(W) || !is.numeric(W)) {
+    stop(paste("The treatment W must be numeric vector. GRF does not", 
+               "currently support non-numeric or non-vector treatments."))
+  }
+
+  if(length(unique(W))!=2) {
```
jtibshirani: I think we should omit this check, as we actually allow for continuous treatments.

Since `validate_Y`, `validate_W` will contain the same logic after removing the check, I think it'd be nice to have a single function `validate_observations(...)` that takes all the observation vectors and validates each one. It would also be nice to simplify and just have one version of method instead of two (right now we have `validate_WX` and `validate_W` for example, when just one method would suffice).

r-package/grf/R/input_utilities.R
```diff
@@ -11,6 +11,55 @@ validate_X <- function(X) {
...
+  }
+}
+
+validate_ZX <- function(Z,X) {
+  validate_Z(Z)
+  if(length(Z) != nrow(X)) { stop("length(Z) does not equal nrow(X).") }
+}
+
+validate_Z <- function(Z) {
+  TRUE
```
jtibshirani: Why do we always return `TRUE` when validating the instrument?

ras44: I didn't find any prior constraints on Z and didn't want to impose the same conditions as those on W and Y.  Judging from the [prior comment](https://github.com/grf-labs/grf/pull/378#discussion_r271132460), it sounds like the Z, W, and Y can all be validated under the same constraints.  I'll follow the suggestion and implement that in the next commits.

ras44:
I've added `validate_observations` which should validate W,Y, and Z all in the same way while also ensuring they have the same length as X.  It can be passed either a list of observation vectors or a single vector.

ras44:
You're welcome @jtibshirani! Thanks for your helpful reviews.

PR #386, opened 2019-04-01
XPtr R interface

### LINKED ISSUE #187: Avoid making extra copies of the forest when serializing to and from R.
After C++  training completes, we serialize the contents of the forest to a byte stream, and pass this up to R. This serialized forest is then passed back to subsequent C++ functions for prediction, analysis, etc.

We currently make several copies of forest during serialization. One of these extraneous copies is likely due to an inefficiency in our C++ code, and can be fixed. It will be more challenging to remove all unnecessary copies, because of limitations in R/ Rcpp's support of variable-sized byte streams.

Looking at memory profiles, these extra copies causs total memory usage to spike when training has completed, and results are being piped up into R. This is likely a big contributor to the memory issues observed in #179.

davidahirshberg:
Avoid serialization/deserialization of the forest on each R call by passing a pointer to the C++ forest object. 

The default behavior allows 'save' to work as it has in the past. To do this, whenever an R forest object
is created we store a serialized version of it, and whenever a method (e.g. predict)  is called on a forest object that does not have a valid pointer to a C++ forest (one that has been saved and then loaded), we deserialize, storing the pointer so that we do not have to deserialize again.  Passing 'serialize=FALSE' to forest constructors ( regression_forest, instrumental_forest, etc. ) saves time and space by not serializing; if such an object is saved without first making an explicit call to `forest$structure$serialize()', loading it yields an invalid forest object. If a method requiring a valid pointer is called on an invalid forest object, an error is thrown.

Closes  #187.


jtibshirani:
I'm going to close this out, as we've decided to take a slightly different approach (and have started on that new path with PR #408). It was great you opened this though, it brought `Rcpp::XPtr` to everyone's attention and showed it was a viable path.

PR #388, opened 2019-04-01
Add boosting option for causal forest

evanmunro:
Causal forest predictions rely on predictions of outcome Y and treatment W given the covariates X. Improved prediction of these nuisance parameters can improve the treatment effect predictions of the causal forest. The following two features have been added: 

 - The option `boosting` has been added to `causal_forest`. If set to True, then E.Y and E.W are selected by boosting a `regression_forest` using a number of boosting steps selected by cross-validation in the first boosting step, and default maximum steps.  
- `boosted_regression_forest` is also exposed in case the user prefers to estimate E.Y and E.W themselves first, using some custom number of steps. `print` and `predict` methods are implemented for the boosted forest. 

Experiments have been run with a few data generating processes evaluating the accuracy of the causal forest predictions. I find that boosting helps prediction of treatment effects, and rarely hurts performance more than a marginal amount. The summery of these simulations is [here](https://github.com/grf-labs/grf/files/3031777/grf_simulation.pdf).


r-package/grf/NAMESPACE
```diff
@@ -13,8 +13,10 @@ S3method(predict,instrumental_forest)
 S3method(predict,ll_regression_forest)
 S3method(predict,quantile_forest)
 S3method(predict,regression_forest)
+S3method(predict,boosted_regression_forest)
```
halflearned: These should be in alphabetic order (sorry this must be done manually)

r-package/grf/R/boosted_regression_forest.R
```diff
@@ -0,0 +1,248 @@
...
+                              samples_per_cluster = NULL,
+                              tune.parameters = FALSE,
+                              num.fit.trees = 10,
+                              num.fit.reps = 100,
+                              num.optimize.reps = 1000,
+                              num.steps = NULL,
+                              max.steps = 6,
+                              num.trees.tune = 10) {
+
+    boosted.forest = NULL
```
halflearned: These two lines could be combined into one

r-package/grf/R/boosted_regression_forest.R
```diff
@@ -0,0 +1,248 @@
...
+    boosted.forest[["forests"]] = list()
+    forest.Y <- regression_forest(X, Y, sample.fraction = sample.fraction,
```
halflearned: I think this could be simplified a good deal. More on this on my overall comment.

r-package/grf/R/causal_forest.R
```diff
@@ -144,24 +147,37 @@ causal_forest <- function(X, Y, W,
 
     reduced.form.weight <- 0
 
-    if (is.null(Y.hat)) {
+    if (is.null(Y.hat) && !boosting) {
```
halflearned: There's already a lot going on in `causal_forest`, and I don't know if we should be adding more to it right now.

Is there a reason for not creating a separate function `boosted_causal_forest`, just like we have `boosted_regression_forest`?

r-package/grf/R/boosted_regression_forest.R
```diff
@@ -0,0 +1,248 @@
...
+        }
+    }
+    else {
+        step <- 2
+        tolerance <- 0.99
+        still_boosting <- TRUE
+        while(still_boosting & step <= max.steps) {
+            E.hat <- Y - Y.hat
+            forest.small <- regression_forest(X,E.hat,
+                                    sample.fraction = as.numeric(tunable.params["sample.fraction"]),
```
swager: Nitpick: Can you double-check the indentation here? In Rstudio, you can automatically fix indentation with CMD-I.

r-package/grf/R/boosted_regression_forest.R
```diff
@@ -0,0 +1,248 @@
...
+                                        clusters = clusters, samples_per_cluster = samples_per_cluster);
+            current.pred <- predict(forest.E)
+            Y.hat <- Y.hat + current.pred$predictions
+            error.debiased <- current.pred$debiased.error
+            boosted.forest[["forests"]][[step]] <- forest.E
+        }
+    }
+    else {
+        step <- 2
+        tolerance <- 0.99
```
swager: This should be pluggable (i.e., there should be an argument in the API where the user can change this). Also, a tolerance of 0.99 is really high. I think we may want a slightly lower default, for example 0.9 or 0.95 (the ideal would be to guide this choice with some simulation experiments).

r-package/grf/R/boosted_regression_forest.R
```diff
@@ -0,0 +1,248 @@
...
+                                           min.node.size = tunable.params["min.node.size"],
+                                           honesty = honesty,
+                                          honesty.fraction = honesty.fraction,
+                                           seed = seed, ci.group.size = ci.group.size,
+                                          alpha = tunable.params["alpha"],
+                                          imbalance.penalty = tunable.params["imbalance.penalty"],
+                                        clusters = clusters, samples_per_cluster = samples_per_cluster);
+            current.pred <- predict(forest.E)
+            Y.hat <- Y.hat + current.pred$predictions
+            error.debiased <- current.pred$debiased.error
```
swager: It might be worth storing + returning the debiased errors from each step, so the user can see how much tuning helped in each step.

r-package/grf/R/boosted_regression_forest.R
```diff
@@ -0,0 +1,248 @@
...
+                                           min.node.size = as.numeric(tunable.params["min.node.size"]),
+                                           honesty = honesty,
+                                          honesty.fraction = honesty.fraction,
+                                           seed = seed, ci.group.size = ci.group.size,
+                                          alpha = as.numeric(tunable.params["alpha"]),
+                                          imbalance.penalty = as.numeric(tunable.params["imbalance.penalty"]),
+                                        clusters = clusters, samples_per_cluster = samples_per_cluster);
+            step.error.approx <- predict(forest.small)$debiased.error
+            if((mean(step.error.approx,na.rm=TRUE) < tolerance*mean(error.debiased,na.rm=TRUE))) {
+                forest.E <- regression_forest(X,E.hat,
```
swager: This is exactly the same thing as lines 116-130. It'd be good to consolidate these two code chunks into a single flow. Something like:
- first compute residuals
- check if we should still do a step of boosting (trivial if num.steps specified, more involved if we're tuning)
- perform a step of boosting if appropriate

r-package/grf/R/boosted_regression_forest.R
```diff
@@ -0,0 +1,248 @@
...
+#'
+#' }
+#'
+#' @method predict boosted_regression_forest
+#' @export
+predict.boosted_regression_forest <- function(object,
+                                        newdata=NULL,
+                                        num.steps=NULL,
+                                      num.threads = NULL
+                                      ) {
```
swager: There shouldn't be a line break before the ).

r-package/grf/R/boosted_regression_forest.R
```diff
@@ -0,0 +1,248 @@
...
+    forest.Y.p <- predict(forest.Y)
+    #save tuned parameters for use on future boosting iterations
+    tunable.params <- forest.Y$tunable.params
+    Y.hat <- forest.Y.p$prediction
+    error.debiased <- forest.Y.p$debiased.error
+    boosted.forest$forests[[1]] <- forest.Y
+
+    if(!is.null(num.steps)) {
+        for (step in 2:num.steps) {
+            E.hat <- Y - Y.hat
```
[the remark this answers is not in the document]
swager: Actually, maybe `residual.error` is also not good, since it sounds like a MSE-type quantity. `Y.residual` may be clearer.

r-package/grf/R/boosted_regression_forest.R
evanmunro: Not clear on which two lines you are referring to here - the two lines with the input parameters from the function?

r-package/grf/R/boosted_regression_forest.R
evanmunro: I chose a tolerance of 0.95 - Although, I have been finding that the low noise DGP results in only a single step being taken, whether the tolerance is 0.95 or 0.99. Seems fairly rare to take more than a couple of steps.

evanmunro:
Thanks for the comments @halflearned  and @swager. I've attempted to address them in the latest commits. Major changes include
- Streamlining the logic of the boosting termination checks
- Adding a set of test cases 

Vitor, in response to your questions about the somewhat weak results in the simulations, I'm running some additional simulations with low noise linear DGPs where boosting is likely to outperform. I'll edit this comment in the next 48 hours when they are done running to include some updated results.

r-package/grf/tests/testthat/test_boosted_regression_forest.R
```diff
@@ -0,0 +1,52 @@
...
+
+ test_that("Under cross-validation, errors decrease each step", {
+   n=200; p=6
+   X = matrix(runif(n * p), n, p)
+   mu = 2 * X[,1]^2 * X[,2] + 3 * X[,3] + 4 * X[,4]
+   Y = mu + rnorm(n)
+   forest.boost <- boosted_regression_forest(X,Y,num.steps=2)
+   errors <- unlist(lapply(forest.boost$debiased.errors,mean))
+   improves <- errors[2:length(errors)] - errors[1:(length(errors)-1)]
+   expect_true(sum(improves>0)==0)
```
jtibshirani: Small comment: I think this can just be `expect_true(all(improves > 0))`.

r-package/grf/tests/testthat/test_boosted_regression_forest.R
```diff
@@ -0,0 +1,52 @@
...
+test_that("Boosted forest takes user specified number of steps",  {
+  n = 100; p = 6
+  X = matrix(runif(n * p), n, p)
+  mu = 2 * X[,1] * X[,2] + 3 * X[,3] + 4 * X[,4]
+  Y = mu + rnorm(n)
+  forest.boost <- boosted_regression_forest(X,Y,num.steps=2)
+  expect_equal(2,length(forest.boost$forests))
+})
+
+ test_that("Tolerance above 1 increases length of boosting", {
```
jtibshirani: Should we actually validate in `boosted_regression_forest` that `tolerance` is always less than 1? It doesn't seem to make a lot of sense to allow a number greater than 1 to be supplied.

r-package/grf/R/boosted_regression_forest.R
```diff
@@ -0,0 +1,229 @@
...
+#'                            of a given cluster to the final forest scales with the minimum of
+#'                            the number of observations in the cluster and samples_per_cluster.
+#' @param tune.parameters If true, NULL parameters are tuned by cross-validation; if false
+#'                        NULL parameters are set to defaults.
+#' @param num.fit.trees The number of trees in each 'mini forest' used to fit the tuning model.
+#' @param num.fit.reps The number of forests used to fit the tuning model.
+#' @param num.optimize.reps The number of random parameter values considered when using the model
+#'                          to select the optimal parameters.
+#' @param num.steps The number of boosting iterations. If NULL, selected by cross-validation
+#' @param tolerance If num.steps is NULL, the percentage of previous steps' error that must be estimated
```
jtibshirani: Small comment, a better name might be `boost.error.reduction`.

r-package/grf/R/boosted_regression_forest.R
```diff
@@ -0,0 +1,229 @@
...
+                                alpha = alpha,
+                                imbalance.penalty = imbalance.penalty,
+                                clusters = clusters, samples_per_cluster = samples_per_cluster);
+  current.pred <- predict(forest.Y)
+  #save tuned parameters for use on future boosting iterations
+  tunable.params <- forest.Y$tunable.params
+  Y.hat <- current.pred$predictions
+  error.debiased <- current.pred$debiased.error
+  boosted.forest[["forests"]][[1]] <- forest.Y
+  boosted.forest[["debiased.errors"]][[1]] <- error.debiased
```
jtibshirani: I think we should store the mean debiased error (and call it `error`) instead of the pointwise errors. This is more consistent with what we return from other tuning methods like `regression_tuning`.

r-package/grf/R/boosted_regression_forest.R
```diff
@@ -0,0 +1,229 @@
...
+#' @param num.optimize.reps The number of random parameter values considered when using the model
+#'                          to select the optimal parameters.
+#' @param num.steps The number of boosting iterations. If NULL, selected by cross-validation
+#' @param tolerance If num.steps is NULL, the percentage of previous steps' error that must be estimated
+#'                  by cross validation in order to take a new step, default 0.95
+#' @param max.steps The maximum number of boosting iterations to try when num.steps NULL
+#' @param num.trees.tune If num.steps is NULL, the number of trees used to test a new boosting step when tuning
+#'        num.steps
+#'
+#' @return A trained regression forest object.
```
jtibshirani: Let's update this description of the return value to let users know that they can access the individual boosted forests in `boosted.forest$forests`, and also the debiased errors at each step.

r-package/grf/R/boosted_regression_forest.R
```diff
@@ -0,0 +1,229 @@
...
+#' @method predict boosted_regression_forest
+#' @export
+predict.boosted_regression_forest <- function(object, newdata=NULL,
+                                              num.steps=NULL) {
+
+  # If not on new data, use pre-computed predictions
+  if (is.null(newdata)) {
+    return(data.frame(predictions=object$predictions))
+  }
+  else {
```
jtibshirani: Small comment, this should be on the same line as the previous brace `}`.

r-package/grf/R/boosted_regression_forest.R
```diff
@@ -0,0 +1,229 @@
...
+#' r.pred = predict(r.boosted.forest, X.test)
+#'
+#' # Predict on out-of-bag training samples.
+#' r.pred = predict(r.boosted.forest)
+#'
+#' }
+#'
+#' @method predict boosted_regression_forest
+#' @export
+predict.boosted_regression_forest <- function(object, newdata=NULL,
```
jtibshirani: We should make sure the `num.threads` parameter is available here as well. I agree that it makes sense to omit the other parameters like `estimate.variance`.

r-package/grf/R/boosted_regression_forest.R
```diff
@@ -0,0 +1,229 @@
...
+  # If not on new data, use pre-computed predictions
+  if (is.null(newdata)) {
+    return(data.frame(predictions=object$predictions))
+  }
+  else {
+    forests <- object[["forests"]]
+    if(is.null(num.steps)) {
+      num.steps <- length(forests)
+    }
+    else {
```
jtibshirani: Same comment here.

r-package/grf/R/boosted_regression_forest.R
```diff
@@ -0,0 +1,229 @@
...
+      num.steps <- min(num.steps,length(forests))
```
jtibshirani: Small comment, maybe `boost.max.steps` would be a better name for this parameter since you are taking a min over it and the number of forests.

r-package/grf/R/boosted_regression_forest.R
```diff
@@ -0,0 +1,229 @@
...
+    }
+    Y.hat <- predict(forests[[1]],newdata)$predictions
```
jtibshirani: A more common way to do this would be

```
Y.hat <- 0
for (forest in 1:num:steps) {
    Y.hat <- Y.hat + predict(...)
}
```

r-package/grf/R/boosted_regression_forest.R
```diff
@@ -0,0 +1,229 @@
+#' Boosted regression forest
```
jtibshirani: Let's add a note that this is experimental so our users aren't surprised if they're a bit slow compared to what they're used to. We could change this heading to `Boosted regression forest (experimental)` and also add a sentence saying that this functionality is experimental and will likely change in future releases.

r-package/grf/tests/testthat/test_boosted_regression_forest.R
```diff
@@ -0,0 +1,52 @@
+library(grf)
+
+set.seed(1)
+
+test_that("Boosted regression forest improves performance vs regular forest", {
```
jtibshirani: It would be nice to add one test around OOB prediction, since precomputing those predictions is a little trickier in the case of boosted forests.

r-package/grf/tests/testthat/test_boosted_regression_forest.R
evanmunro: Agreed, I'll also remove this test.

r-package/grf/R/boosted_regression_forest.R
evanmunro: I named it predict.steps - does that make sense? It is meant to allow the user to specify the number of forests from the boosted object to use for prediction (they may not want to use all of them, if they didn't use the cross-validation procedure). But you take the min since they can't use more than the number of forests in the boosted object.

r-package/grf/tests/testthat/test_boosted_regression_forest.R
evanmunro: What sort of test would you suggest to test the OOB prediction?

evanmunro:
I left a few questions/responses unresolved on your comments. 

For the timing, without tuning, boosting scales linearly in the number of forests that it trains. I have set the default number of trees for boosted forests to 500, vs regression forests where they are 2000. So with default settings, boosting tends to be faster.  I've changed the boost.error.reduction default value to 0.97, since the 5% cutoff per step seems to select only a single step of boosting too often, when another step will still improve prediction by 3-4% (is that ok @swager?). 

With N=50,000, and p = 10, on my local laptop: 
boosted_regression_forest runtime (2 steps): 92 seconds
regression_forest runtime: 170 seconds

If I manually specify the same number of trees (500 for regression forest as well), then the regression forest is faster: 
boosted_forest runtime (5 steps): 255 seconds
regression_forest runtime: 48 seconds 

Tuning parameters doesn't make much of a difference in the timing order, since boosting only tunes on the first step. 

Happy to add a short section to the algorithm reference, will work on that along with closing the last few comments before merging.

r-package/grf/tests/testthat/test_boosted_regression_forest.R
jtibshirani: In the case you describe, could the user just use `boost.steps` to select the number of steps they want?

r-package/grf/R/boosted_regression_forest.R
jtibshirani: Yes I find `predict.steps` a lot clearer, thanks!

jtibshirani:
> I have set the default number of trees for boosted forests to 500, vs regression forests where they are 2000.

I missed this while reviewing, but I think that boosted forests should have the same default as standard forests (`num.trees = 2000`). This will be the least surprising behavior for users.

>  I've changed the boost.error.reduction default value to 0.97, since the 5% cutoff per step seems to select only a single step of boosting too often, when another step will still improve prediction by 3-4% (is that ok @swager?).

I checked with @swager and he is good with this change.

> Happy to add a short section to the algorithm reference, will work on that along with closing the last few comments before merging.

Great, thanks for doing that. Maybe we could update the reference in a follow-up PR though, since we need to think through some details around when to recommend boosting for orthogonalization.

r-package/grf/tests/testthat/test_boosted_regression_forest.R
jtibshirani: We could draw a test set from the same distribution as the training data, and check that the mean prediction error is not too different. We'll need to train a decent number of trees to make up for the fact that OOB prediction is only able to use the out-of-bag trees for each datapoint.

r-package/grf/tests/testthat/test_boosted_regression_forest.R
evanmunro: You're right, added validation check to the input_utilities file

r-package/grf/tests/testthat/test_boosted_regression_forest.R
evanmunro: Added this

evanmunro:
Addressed the remaining comments and will wait for now on updating the reference.

r-package/grf/tests/testthat/test_boosted_regression_forest.R
```diff
@@ -21,22 +21,46 @@ test_that("Boosted regression forest improves performance vs regular forest", {
...
+
+test_that("OOB prediction is close to actual out of sample error", {
+  n<-3000; p<-6
+  X <- matrix(runif(n * p), n, p)
+  mu <- 2 * X[,1]^2 * X[,2] + 3 * X[,3] + 4 * X[,4]
+  Y <- mu + rnorm(n)
+  test <- 2000:3000
+  train <- 1:2000
+  forest.boost <- boosted_regression_forest(X[train,],Y[train],boost.steps=2)
+  OOB_error <- unlist(forest.boost$error)[length(forest.boost$error)]
```
jtibshirani: Could we use the actual OOB predictions here (what's returned from `predict(forest.boost)`), as that is what users will primarily be working with?

Also small comment: variable names in R should use dots instead of underscores.

evanmunro: This is fixed now.

evanmunro:
For the conflict, it looks like there are a few more arguments to causal_forest than when I had forked? Seems like the version with the additional arguments is correct, but could someone who is more familiar with the code take a look?

PR #389, opened 2019-04-02
Fix quantile forest bindings

### LINKED ISSUE #387: Quantile regression forest broken by commit 996f3b3fd4d89761b7b9ed6b0da31dbbba6895ac
**Description of the bug**
predict returns nondeterministic junk --- probably the contents of uninitialized memory.

**Steps to reproduce**
The following code is a stripped-down version experiments/quantile_examples/quantile_simu_2.R. 

p = 10
n = 500
i = 5
X = matrix(2 * runif(n * p) - 1, n, p)
Y = rnorm(n)  + 100 * (X[,i] > 0)
X.new = matrix(2 * runif(n * p) - 1, n, p)

forest = quantile_forest(X, Y)
all(predict(forest,X.new) == predict(forest,X.new)) # true pre-commit, false afterward
predict(forest,X[1:4, ])                                                # reasonable pre-commit, junk afterward

**GRF version**
Development, commit 996f3b3fd4d89761b7b9ed6b0da31dbbba6895ac and after

davidahirshberg:
Fixes two issues with quantile forest introduced with the Data/Observations rewrite in commit 996f3b3fd4d89761b7b9ed6b0da31dbbba6895ac.
1. Nondeterministic junk output from predict caused by
    a. bindings failing to subtract 1 from outcome_index to go from R to C indexing conventions
    b. R code passing [X, null] = X rather than [X, Y] to the C code 
2. predict(forest) giving an error, caused by disagreement between R and C code on argument order.

Closes #387.

### DIFF
--- a/r-package/grf/tests/testthat/test_quantile_forest.R
+++ b/r-package/grf/tests/testthat/test_quantile_forest.R
@@ -54,6 +54,6 @@ test_that("quantile forest predictions for 90th percentile are strongly positive
     Y = runif(n) + 100 * (X[,i] > 0)
 
     qrf = quantile_forest(X, Y, quantiles = c(0.1, 0.5, 0.9), mtry = p, min.node.size = 10, sample.fraction = 0.632)
-  expect_true(cor(predict(qrf, quantiles=.9), X[,i]) > .5)
+    expect_true(cor(predict(qrf, quantiles=.9), X[,i]) > .5)
     expect_true(cor(predict(qrf, X.new, quantiles=.9), X.new[,i])  > .5)
 })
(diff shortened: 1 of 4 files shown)


davidahirshberg:
Yeah, I think it might make sense to replace create_data_matrices with something that returns a pointer to a Data object. That way, the slots containing w,x,y,z, etc are known by that Data object and we don't need to pass around indices.

r-package/grf/tests/testthat/test_quantile_forest.R
```diff
@@ -30,3 +30,30 @@ test_that("quantile forests with regression splitting are identical to regressio
...
+					 with covariate strongly positively correlated with Y", {
+	p = 10
+	n = 500
+	i = 5
+	X = matrix(2 * runif(n * p) - 1, n, p)
+	X.new = matrix(2 * runif(n * p) - 1, n, p)
+	Y = runif(n) + 100 * (X[,i] > 0)
+
+	qrf = quantile_forest(X, Y, quantiles = c(0.1, 0.5, 0.9), mtry = p, min.node.size = 10, sample.fraction = 0.632)
+  expect_true(cor(predict(qrf, quantiles=.9), X[,i]) > .5)
```
jtibshirani: Super small comment: spacing is off here.

PR #398, opened 2019-04-16
Added preliminary structure for local linear causal tuning

rinafriedberg:
This PR adds a function, tune_ll_causal_forest, which selects a regularization parameter lambda for the local linear regression step by minimizing estimated OOB errors. We estimate these errors by using the R-learner from Nie and Wager (2017). The PR also includes a basic test analogous to the LL regression forest tuning test.


r-package/grf/R/causal_forest.R
```diff
@@ -308,7 +310,9 @@ predict.causal_forest <- function(object, newdata = NULL,
                                   linear.correction.variables = NULL,
                                   ll.lambda = 0.1,
                                   ll.weight.penalty = FALSE,
-                                  num.threads = NULL, estimate.variance = FALSE, ...) {
+                                  num.threads = NULL,
+                                  estimate.variance = FALSE,
+                                  tune.parameters = FALSE, ...) {
```
jtibshirani: I'm wondering why we have this new parameter `tune.parameters`, whereas in `regression_predict` we don't? It would be good to make these two methods consistent. I also noticed that here `ll.lambda` defaults to `0.1`, whereas for `regression_predict` it defaults to `NULL`.

r-package/grf/R/local_linear_causal_tuning.R
```diff
@@ -0,0 +1,68 @@
...
+  # Subtract 1 to account for C++ indexing
+  linear.correction.variables = linear.correction.variables - 1
+
+  # Find sequence of predictions by lambda
+  prediction.object = ll_causal_predict_oob(forest.short, data$default, data$sparse, outcome.index,
+      treatment.index, instrument.index, lambda.path, ll.weight.penalty, linear.correction.variables, num.threads)
+  predictions = prediction.object$predictions
+
+  # Use R-learner to estimate tau
+  fit = rlearner::rboost(X, W, Y)
```
jtibshirani: I don't think we need a dependency on the `rlearner` package -- could we just compute the r-learner loss here directly?

r-package/grf/tests/testthat/test_causal_forest_tuning.R
```diff
@@ -38,3 +38,28 @@ test_that("causal forest tuning only cross-validates null parameters", {
     expect_equal(as.numeric(tunable.params["min.node.size"]), min.node.size)
     expect_equal(as.numeric(tunable.params["imbalance.penalty"]), imbalance.penalty)
 })
+
+test_that("local linear causal forest tuning returns lambda and decreases error", {
```
jtibshirani: It would also be nice to add a test analogous to "output of tune local linear forest is consistent with prediction output" (in `test_local_linear_forest`). This test ended up helping us catch an important bug.

r-package/grf/R/local_linear_causal_tuning.R
swager: Yes agreed, there must have been a miscommunication here. The idea is to minimize the R-loss \hat{L} defined in (6) from https://arxiv.org/pdf/1712.04912.pdf to choose lambda. The nuisance components \hat{e} and \hat{m} used to form this loss function are the same as those used to grow the forest in the first place -- there's no need to train any new predictor here.

rinafriedberg: Ohh good point- this is doing lots of extra work. Replaced, thanks for the find!

r-package/grf/NAMESPACE
```diff
@@ -31,5 +31,6 @@ export(split_frequencies)
 export(test_calibration)
 export(tune_causal_forest)
 export(tune_ll_regression_forest)
+export(tune_ll_causal_forest)
```
jtibshirani: These should be alphabetized, unfortunately there's an issue with the build and we need to do it by hand :(

rinafriedberg:
@jtibshirani - added that test, which has a few failures. Not sure what's going on and trying to take a look- let me know if you see anything clear, otherwise will keep you posted! Thanks!

jtibshirani:
@rinafriedberg I did some debugging and it looks like we aren't centering the data in `tune_ll_causal_forest`. Here's an example of centering from `predict.causal_forest`: https://github.com/grf-labs/grf/blob/master/r-package/grf/R/causal_forest.R#L222-L225. This part of the R code could really use some work, it is so easy to miss this data processing step.

r-package/grf/NAMESPACE
jtibshirani: I just fixed the issue in #423, so now you shouldn't need to touch the `NAMESPACE` file at all!

rinafriedberg:
@jtibshirani Hopefully this last commit fixes that issue! Although I agree we should put some time into these chunks of R code for clarity moving forward.

PR #405, opened 2019-04-22
added variable interaction count

ginward:
Count the variable interaction frequencies. For example, if age was split in the first node, and education was split in the second node, I count `age  education` as one interaction.


ginward:
0 means leaf node, numbers starting from 1 indicates the variables.

swager:
Thanks a lot, @ginward, for this contribution! Some high-level comments before diving into the code review:

First, could you please include a short code snippet that shows example use + output of this function?

Second, this kind of interaction analysis seems very interesting. Is there are reference in the literature on anyone doing something like this? Some questions I had when looking at this:
- I'd imagine I might be interested in the depth at which different interactions occur. If on tree first splits on feature A and then on feature B, whereas a second tree does its third split on A and fifth one on B, then qualitatively it seems like the first tree has a stronger A-B interaction than the second one. The R function definition suggests that this information is returned to the user; what exactly is the display strategy here?
- In order for there to actually be an A-B interaction, it's not enough to have both a split on A and a split on B in the same tree; rather, we'd need splits on both A and B on a path leading to a specific leaf. It's not clear to me how to best reflect this.

ginward:
Hi @swager thanks for your reply! 

This is the code that you can use for the interaction frequencies ([the data is here](https://www.dropbox.com/s/h5q2eozxmtrr7op/bug.csv?dl=0)): 

```
#compile and include the grf package
library(dplyr)
select <- dplyr::select
TREES_CPS=100
SEED=6000
set.seed(SEED)
dat=read.csv("bug.csv")
forest=causal_forest(seed=SEED,num.trees = TREES_CPS, X=as.matrix(select(dat, re74, re75, age, education, black, hispanic, married, nodegree)), Y=dat$re78, W=dat$treat)

interact=interaction_frequencies(forest)
interact=interact[order(interact$count, decreasing = TRUE),]
```
The output you will get looks like this: 

> interact
```
   node1 node2 count
3      0     2  1149
2      0     1  1135
1      0     0   634
4      0     3   578
11     1     2   464
19     2     3   334
12     1     3   270
10     1     1   205
18     2     2   182
20     2     4   165
25     3     3   147
5      0     4   119
7      0     6   111
13     1     4   111
8      0     7    86
26     3     4    85
23     2     7    49
16     1     7    39
29     3     7    37
6      0     5    36
28     3     6    36
15     1     6    27
24     2     8    25
21     2     5    24
14     1     5    22
22     2     6    21
27     3     5    20
17     1     8    17
34     4     7    17
33     4     6    12
9      0     8    11
30     3     8    11
31     4     4    10
39     6     8     4
40     7     7     4
32     4     5     3
35     4     8     2
37     5     7     2
36     5     6     1
38     5     8     1
```

Of course, if I replace the variable names, it will look like this: 

```
> interact
       node1     node2 count
3       leaf      re75  1149
2       leaf      re74  1135
1       leaf      leaf   634
4       leaf       age   578
11      re74      re75   464
19      re75       age   334
12      re74       age   270
10      re74      re74   205
18      re75      re75   182
20      re75 education   165
25       age       age   147
5       leaf education   119
7       leaf  hispanic   111
13      re74 education   111
8       leaf   married    86
26       age education    85
23      re75   married    49
16      re74   married    39
29       age   married    37
6       leaf     black    36
28       age  hispanic    36
15      re74  hispanic    27
24      re75  nodegree    25
21      re75     black    24
14      re74     black    22
22      re75  hispanic    21
27       age     black    20
17      re74  nodegree    17
34 education   married    17
33 education  hispanic    12
9       leaf  nodegree    11
30       age  nodegree    11
31 education education    10
39  hispanic  nodegree     4
40   married   married     4
32 education     black     3
35 education  nodegree     2
37     black   married     2
36     black  hispanic     1
38     black  nodegree     1
```

Basically, it counts the interaction frequencies.

ginward:
@swager  I will reply to your other questions tomorrow. Thanks!

ginward:
@swager Thanks for your comment. Here are my replies to your questions. 

I would like to answer your question 2 first: 

`In order for there to actually be an A-B interaction, it's not enough to have both a split on A and a split on B in the same tree; rather, we'd need splits on both A and B on a path leading to a specific leaf. It's not clear to me how to best reflect this.
`
I believe there is a misunderstanding here. I only count the interaction of two consecutive nodes, like the image shown below: 

image

Therefore, I am not counting random combinations of variables in the forest. I am counting on the variables that are directly linked together in a tree. For example, if var1 x var2 appeared once in a tree, I add the frequency by adding 1 to it. If I found another var1 x var2 in the same tree or another tree, I add another one to it. 

Does that help? 

Now question 1: 

`I'd imagine I might be interested in the depth at which different interactions occur. If on tree first splits on feature A and then on feature B, whereas a second tree does its third split on A and fifth one on B, then qualitatively it seems like the first tree has a stronger A-B interaction than the second one. The R function definition suggests that this information is returned to the user; what exactly is the display strategy here?`

I agree with you. So far I only implemented the count function where each frequency is weighted equally, regardless of its depth. 

There are two ways I see that can do this: 

- Weight by MSE Reduction 
- Weight by Tree Depth 

You can also have a look at [this](https://onlinelibrary.wiley.com/doi/full/10.1002/jae.1046) paper. It is similar to counting the occurrence of coefficients. 

Did I answer your question? Do you think I need to implement the weighting for it to be merged into the package? Or is it possible to merge this preliminary version (after code review, of course). 

Jinhua

swager:
Thanks @ginward, this is very helpful -- I now understand better what you're doing. I think adding a way to summarize variable interactions in forests would be very useful. However, before adding such a feature to `grf`, we'd need to make sure that whatever we're adding in fact summarizes interactions in a way that meets a need from applied researchers.

My first intuition had been that the most natural measure of A-B interactions (for features A and B) is the fraction of leaves who have upstream splits along both A and B. What you're doing here is also interesting but different, i.e., computing the number of back-to-back splits on first A then B. Do you have a theoretical reason to back this latter measure of interaction frequency?

Overall, I think it'd be helpful to support a PR like this using either a literature review on different approaches to summarizing feature interactions in a forest, or a technical white-paper that compares and discusses various approaches to the problem. We should also think through how practitioners could use the output from this type of a function to produce useful diagnostics to accompany applications of causal forests.

Issue #396, opened 2019-04-12
test_calibration differential.forest.prediction p-value very often significant 


bakered:
**Description**
When there is no heterogeneity in the treatment effect, the differential.forest.prediction comes out highly significant. e.g. in the code below significant at 0.05 33% of the time.

**Steps to reproduce**
```
n = 800; p = 5; alpha = 0.05
sim<-vector(length=100)
for(s in 1:100){
  X = matrix(rnorm(n*p), n, p)
  W = rbinom(n, 1, 0.5)
  Y = W*5
  forest = causal_forest(X, Y, W)
  sim[s]<-test_calibration(forest)[2,4]<alpha
}
table(sim)
```

**GRF version**
Version 0.10.2

swager:
Very interesting, @bakered, thanks for sharing. I tried replicating a closely related simulation (with some noise to make sure we're not hitting a zero-noise pathology), and got 26% rejections. Now what's curious is that the z-values for the heterogeneity coefficient are systematically negative. And, of course, with a properly calibrated forest we'd want the coefficient `differential.forest.prediction` to be positive, and preferably close to 1.

As a short-term fix, it might be helpful to consider using one-sided p-values in the `test_calibration` function (since, really, we're hoping to find positive, rather than non-zero, coefficients). Beyond that, it'd be super interesting to get a deeper understanding of what's going on. If anyone has clean formal results -- ideally that work for leave-one-out cross-fitting as is most computationally practical with forest -- please chime in!

```
n = 800; p = 5
zvalue = replicate(1000, {
    X = matrix(rnorm(n*p), n, p)
    W = rbinom(n, 1, 0.5)
    Y = W*5 + rnorm(n)
    forest = causal_forest(X, Y, W)
    test_calibration(forest)[2,1]
})
```

![image](https://user-images.githubusercontent.com/1275702/56253102-dd2f1f00-606f-11e9-83f1-25b84fad77fe.png)

ras44:
Using the suggestion from @swager, running the following produces a false positive rate within the 5% as set by alpha:

```
n = 800; p = 5; alpha = 0.05
sim<-vector(length=100)
for(s in 1:100){
  X = matrix(rnorm(n*p), n, p)
  W = rbinom(n, 1, 0.5)
  Y = W*5
  forest = causal_forest(X, Y, W)
  tc <- test_calibration(forest)
  sim[s]<-tc[2,4]<alpha & tc[2,1]>0
}
table(sim)
```

[...]

PR #410, opened 2019-04-26
Use one-sided p-values for calibration test

ras44:


r-package/grf/R/forest_summary.R
```diff
@@ -70,6 +70,8 @@ test_calibration = function(forest) {
           "as well as the mean forest prediction as regressors, along",
           "with heteroskedasticity-robust (HC3) SEs",
           sep="\n")
+  dimnames(blp.summary)[[2]][4] <- gsub("[|]", "", dimnames(blp.summary)[[2]][4])
+  blp.summary[,4] <- blp.summary[,4]/2
```
swager: This won't give the one-sided p-value in the case where the t-stat is negative.

ras44: Thanks for your valuable comments.  Responding in the same order:
- re: mean.forest.prediction consistency, I believe `blp.summary[,4] <- blp.summary[,4]/2` should modify both the mean.forest.prediction and the differential.forest.prediction values.  Please let me know if I misunderstood.
- Re pretty printing, two issues:
  - "< 2e-16" decorator is dynamically created when the result is printed (verify with `res <-test_calibration(forest); res[1,4] <- 4e-16; res; res[1,4] <- 1e-16; res`
  - *-based significance decorators, these are dynamically created when the result is printed and should correspond to p-values modified after the result (verify with `res <- test_calibration(forest); res[1,4]<-0.999`)
  - In short, I think both of these are handled by the dynamic decoration that happens on print
- how about something like "... along with one-sided heteroskedasticity-robust (HC3) SEs:"?
- I've added a commit to address the negative t-stat case. Would it be helpful to have a short comment in the code here?  Something like "#convert to one-sided p-values"?

swager: This all sounds good, thanks! The formula for going from a 2-sided p-value to a 1-sided one is `ifelse(t < 0, 1 - pval / 2, pval / 2)`. Also, nitpick: we don't need to use the `sign` command on the t-stat since we're only checking if it's smaller than 0.

ras44: agreed!  the `sign()` was an unfortunate relic of a first attempt to do the calculation without an ifelse.  Also, previously it might have looked like I was using the formula like ifelse(..., 1-pval, pval), but that was because I was setting pval to pval/2 in the prior line in the code.  It should now be a little more clear, all in one line.  I've also added the update to the function description and a comment.

swager:
In response to #396

ras44:
Great!  Thanks for your helpful reviews, @swager

PR #413, opened 2019-05-03
Fix the naming style of samples.per.cluster.

jtibshirani:
In our R code, method parameters always contain dots instead of underscores.


jtibshirani:
Thanks @halflearned ! 🌴

PR #414, opened 2019-05-03
Clean-up the causal forest bindings and related logic.

jtibshirani:
* Add a binding for `causal_train`, so that the  R `causal_forest` methods consistently call into `CausalForestBindings` (and not `InstrumentalForestBindings` as well).
* Remove the `instrument_index` parameter from all methods in `CausalForestBindings`.


jtibshirani:
* Add a binding for `causal_train`, so that the  R `causal_forest` methods
consistently call into `CausalForestBindings` (and not
`InstrumentalForestBindings` as well).
* Remove the `instrument_index` parameter from all methods in
`CausalForestBindings`.

PR #415, opened 2019-05-04
Ensure that nodes aren't split unless it improves the objective.

jtibshirani:
Previously, we stopped splitting if all outcomes in a leaf are the same. This purity requirement didn't make sense for forest types like causal forest that used observations besides the outcome, so we removed it in #362. However, we did not accordingly adjust the termination condition to make sure we stop splitting once the objective can no longer be approved.

This PR updates the splitting termination condition so that nodes do not split unless they improve the objective. It's important to note that after this change, `regression_forest` will behave very similarly to how it behaves in 0.10.2, but **`causal_forest` will likely split less aggressively**. We should clearly communicate this change in behavior when we announce the 0.10.3 release.


jtibshirani:
Thanks a lot @swager for the review! 🌲

PR #418, opened 2019-05-07
added sample weights for causal forest and boosted regression forest

davidahirshberg:
Added sample weights for training causal forests and boosted regression forests. A new test in [test_causal_forest.R](https://github.com/davidahirshberg/grf/blob/7ee28513e1458b338470234d57beea643657dffc/r-package/grf/tests/testthat/test_causal_forest.R#L159) shows that inverse probability of complete case weighting improves the performance of that causal forest in terms of MSE on the complete and incomplete cases.


swager:
Thanks, @davidahirshberg! One thing is: once we add sample weights to causal forest, we also need to account for them in all the causal forest post-processing functions (including `average_treatment_effect`, `average_partial_effect`, and `test_calibration`). Also, functions like `average_treatment_effect` with sample weights should have dedicated tests (including on the interaction with cluster-robust standard errors with unevenly sized clusters).

core/src/prediction/InstrumentalPredictionStrategy.cpp
```diff
@@ -171,19 +171,26 @@ PredictionValues InstrumentalPredictionStrategy::precompute_prediction_values(
...
     double sum_WZ = 0;
 
+    double sum_weight = 0.0;
     for (auto& sample : leaf_samples[i]) {
-      sum_Y += data->get_outcome(sample);
-      sum_W += data->get_treatment(sample);
-      sum_Z += data->get_instrument(sample);
-      sum_YZ += data->get_outcome(sample) * data->get_instrument(sample);
-      sum_WZ += data->get_treatment(sample) * data->get_instrument(sample);
+      sum_Y +=  data->get_weight(sample) * data->get_outcome(sample);
```
jtibshirani: Maybe we could call `data->get_weight(sample)` once and store it in a variable. I think it would help readability, and wouldn't hurt performance-wise.

r-package/grf/R/regression_forest.R
```diff
@@ -5,6 +5,10 @@
 #'
 #' @param X The covariates used in the regression.
 #' @param Y The outcome.
+#' @param sample.weights Weights defining the population on which we want our estimator of m(x) to perform well
```
jtibshirani: I'm not sure this comment is precise, since we don't incorporate sample weights into splitting?

davidahirshberg: I updated the post-processing functions and added tests for them as advised, made the recommended style change in InstrumentalPredictionStrategy, and changed the function signature docs on sample weights to explain how they are used [now sample weighting in causal forest changes the estimand for the ATE/APE, whereas for an HTE they just change the loss] and why they are used [ inverse probability of complete case weighting in censored data]. 

I think that given that weights are pretty conceptually heavy, it may be best to leave the fact that weights aren't used in splitting as an implementation detail. We're still minimizing an estimate of the sample-weighted R-loss, and I think that could be lost on users if we tell them we don't use sample weights in splitting. I'd want to talk about it in a context in which I can say that, and as it stands the function signature docs don't talk about R-loss. Is there a place for that?  If not, maybe future grf.Rd?

core/src/prediction/InstrumentalPredictionStrategy.cpp
davidahirshberg: Done.

davidahirshberg:
I updated the post-processing functions and added tests for them as advised, made the recommended style change in InstrumentalPredictionStrategy, and updated the function signature docs on sample weights to explain how they are used [now sample weighting in causal forest changes the estimand for the ATE/APE, whereas for an HTE they just change the loss] and why they are used [ inverse probability of complete case weighting in censored data].

r-package/grf/R/regression_forest.R
jtibshirani: That sounds good to me -- I think a deeper explanation of weights would make a nice addition to our [algorithm reference](https://github.com/grf-labs/grf/blob/master/REFERENCE.md), it would be great to do that as part of this PR.

davidahirshberg: Done

davidahirshberg:
I think I've addressed everything now.

r-package/grf/tests/testthat/test_causal_forest.R
```diff
@@ -125,12 +125,56 @@ test_that("local linear causal forests with large lambda are equivalent to causa
...
+   e.cc = 1/(1+exp(-3*X[,1]))
+   cc = as.logical(rbinom(n, 1, e.cc))
+   sample.weights = 1/e.cc
+
+   forest = causal_forest(X[cc,], Y[cc], W[cc])
+   boosted.forest = causal_forest(X[cc,], Y[cc], W[cc], orthog.boosting = TRUE)
+   weighted.forest = causal_forest(X[cc,], Y[cc], W[cc], sample.weights = sample.weights[cc])
+   boosted.weighted.forest = causal_forest(X[cc,], Y[cc], W[cc], sample.weights = sample.weights[cc], orthog.boosting = TRUE)
+   mse = function(f) { sum((predict(f,X) - tau)^2) }
+   expect_true(mse(weighted.forest) < mse(forest))
```
swager: Can we get these tests to pass with non-trivial separation? E.g.,  `mse(weighted.forest) < 0.9 * mse(forest)`.

REFERENCE.md
```diff
@@ -210,6 +211,14 @@ When clustering is enabled during training, all subsampling procedures operate o
...
 
+### Sample-Weighting
+
+When the distribution of data that you observe is not representative of the population you are interested in scientifically, it can be important to adjust for this. We pass `sample.weights` to specify that in our population of interest, we observe Xi with probability proportional to `sample.weights[i]``. By default, these weights are constant, meaning that our population of interest is the population from which X1 ... Xn are sampled. For causal validity, the weights we use should not be confounded with the potential outcomes --- typically this is done by having them be a function of Xi. One common example is inverse probability of complete case weighting to adjust for missing data, which allows us to work only the complete cases [the units with nothing missing] to estimate properties of the full data distribution [all units as if nothing were missing].
```
jtibshirani: A couple additional comments:
- There's a stray backtick in "we observe Xi with probability proportional to ... " that will cause all text after this to not be rendered correctly.
- "We pass `sample.weights`" -> "We can pass `sample.weights`"
- In line with @swager's point, I think we should remove the line "One common example is inverse probability...". I don't think there's enough detail for this point to really be understandable to a general user. In general I think we should omit detailed discussions of use cases from the reference, and prefer to put those in a tutorial.

REFERENCE.md
```diff
@@ -210,6 +211,14 @@ When clustering is enabled during training, all subsampling procedures operate o
...
+
+When our estimand is a function of x, e.g. `r(x) = E[Y | X=x]` in `regression_forest` or `tau(x)=E[Y(1)-Y(0)| X=x]` in `causal_forest`, passing weights that are a function of x does not change our estimand. It instead prioritizes fit on our population of interest. In `regression_forest`, this means minimizing weighted mean squared error, i.e. mean squared error over the population specified by the sample weights. In `causal_forest`, this means minimizing weighted R-loss (Nie and Wager (2017), Equations 4/5). When our estimand is an average of such a function, as in `average_treatment_effect` and `average_partial_effect`, it does change the estimand. Our estimand will be the average treatment/partial effect over the population specified by our sample weights.
```
jtibshirani: Could we add a sentence at the start to explain what the purpose of this paragraph is? Maybe "As a clarification, sample weights only sometimes affect the function we are estimating."

REFERENCE.md
```diff
@@ -210,6 +211,14 @@ When clustering is enabled during training, all subsampling procedures operate o
...
+
+Our current implementation does not use the sample weights when in our splitting rules. In terms of Equation 2 of the GRF paper, it replaces `alpha(X_i)` with `alpha(X_i)` = sample.weight[i] * alpha(X_i)` without changing `alpha` itself. This does make what we minimize an estimate of mean squared error / R-loss over the population specified by our sample weights. However, it is likely that we would learn better neighborhood weights `alpha' for the task of estimating this weighted loss if we used our sample weights in splitting as well, by sample-weighting the terms in Equation 4 of the GRF paper. Future implementations may do this.
```
jtibshirani: A few other notes:
- "Our current implementation does not use the sample weights when in our splitting rules." -> "Our current implementation does not use the sample weights during tree splitting."
- The sentence "This does make what we minimize an estimate of mean squared error / R-loss over the population specified by our sample weights." is hard to grok.
- Can we remove 'Future implementations may do this'? It's better to track all speculative work in GitHub issues, so if we decide against it we don't need to update the reference.

REFERENCE.md
```diff
@@ -210,6 +211,14 @@ When clustering is enabled during training, all subsampling procedures operate o
 
 Note that when clusters are provided, standard errors from `average_treatment_effect` and `average_partial_effect` estimation are also cluster-robust. Moreover, if clusters are specified, then each cluster gets equal weight. For example, if there are 10 clusters with 1 unit each and per-cluster ATE = 1, and there are 10 clusters with 19 units each and per-cluster ATE = 0, then the overall ATE is 0.5 (not 0.05).
 
+### Sample-Weighting
```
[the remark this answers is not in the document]
davidahirshberg: I'll change the docs for the individual functions, but I could use some advice on rewriting the REFERENCE section. We use weights to do two conceptually different things that are related by the idea of change of measure: to minimize a weighted loss function when estimating a function of x, and to specify the population we average over in the ATE/APE/forest summary. If we were using other weights, e.g. the inverse of Var[Y|X,W], for estimating a HTE, then our ATE estimate wouldn't be what you'd want.

r-package/grf/tests/testthat/test_causal_forest.R
```diff
@@ -125,12 +125,56 @@ test_that("local linear causal forests with large lambda are equivalent to causa
...
+   p = 6
+   X = matrix(rnorm(n*p), n, p)
+   e = 1/(1+exp(-3*X[,1]+3*X[,2]))
+   W = rbinom(n, 1, e)
+   tau = 2*X[,1] + X[,2]
+   Y = W * tau + rnorm(n)
+   e.cc = 1/(1+exp(-3*X[,1]))
+   sample.weights = 1/e.cc
+
+   forest.1 = causal_forest(X, Y, W, sample.weights = sample.weights)
```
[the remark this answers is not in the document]
davidahirshberg: At least some of these sample sizes are roughly calibrated so that we get a meaningful test. When sample sizes are much smaller, I have to increase overlap between the censored and uncensored populations, so that the ones dealing with censoring aren't so variable that the tests are meaningless. At some point, this makes the bias due to censoring small enough that the weights don't have much to correct for. I'll see what I can do. Definitely fewer trees.

r-package/grf/R/forest_summary.R
```diff
@@ -76,3 +67,20 @@ test_calibration = function(forest) {
   blp.summary
 
 }
+
+
+observation_weights_for_predictions = function(forest) {
```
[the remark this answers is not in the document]
davidahirshberg: Good catch. It isn't, really. observation_weights was just specific to causal forest in that it used Y.hat to get the size of something, so I wrote this without thinking. I'll fix that.

r-package/grf/tests/testthat/test_causal_forest.R
davidahirshberg: Usually we get improvement better than that, but even this test would probably fail 1/50 times. 
I'll rewrite it with a fixed seed.

davidahirshberg: And a smaller forest. But with a smaller forest, behavior is a little more unpredictable, so the result may be a little more sensitive to choice of seed.

davidahirshberg:
I've added the duplication vs weighting test, reduced forest size in causal forest tests, changed the in-function documentation, and made a few of the changes Julie recommended to the REFERENCE. I'll defer a larger rewrite of the section on sample weights in the REFERENCE until I have a better understanding of how our leadership wants to present them.

r-package/grf/tests/testthat/test_causal_forest.R
```diff
@@ -125,12 +125,105 @@ test_that("local linear causal forests with large lambda are equivalent to causa
...
+   e.cc = 1/(1+exp(-2*X[,1]))
+   cc = as.logical(rbinom(n, 1, e.cc))
+   sample.weights = 1/e.cc
+
+   num.trees = 400
+   forest = causal_forest(X[cc,], Y[cc], W[cc], num.trees=num.trees, seed = seed)
+   boosted.forest = causal_forest(X[cc,], Y[cc], W[cc], orthog.boosting = TRUE, num.trees = num.trees, seed = seed)
+   weighted.forest = causal_forest(X[cc,], Y[cc], W[cc], sample.weights = sample.weights[cc], num.trees = num.trees, seed = seed)
+   boosted.weighted.forest = causal_forest(X[cc,], Y[cc], W[cc], sample.weights = sample.weights[cc], orthog.boosting = TRUE, num.trees = num.trees, seed = seed)
+   mse = function(f) { sum((predict(f,X) - tau)^2) }
```
swager: For non-censored observations, these predictions aren't out-of-bag (so their MSE may not reflect the actual accuracy of the forest).

Can we create a separate test set here, and do clean MSE evaluation on test set predictions, so there's no risk of weirdness?

swager: Also, if we define an MSE function, it should compute the MSE :)

r-package/grf/tests/testthat/test_causal_forest.R
```diff
@@ -125,12 +125,105 @@ test_that("local linear causal forests with large lambda are equivalent to causa
...
+  forest.rep = regression_forest(X.rep, W.rep, num.trees=num.trees, seed = seed)
+  forest.weight = regression_forest(X[rep(1:n, 2),], c(W.a,W.b),
+    sample.weights=sample.weights, num.trees=num.trees, seed = seed)
+  p.rep = predict(forest.rep,X)$predictions
+
+  p.weight = predict(forest.weight,X)$predictions
+  expect_true(mean((p.rep - p.weight)^2)/ mean(p.rep^2) < .1)
+
+  ## causal forest
+  forest.rep = causal_forest(X.rep, Y.rep, W.rep, num.trees=num.trees, seed = seed)
```
swager: We should re-use the propensity estimates computed above (by passing them in via the `W.hat` argument) to save on computation.

r-package/grf/tests/testthat/test_causal_forest.R
```diff
@@ -125,12 +125,105 @@ test_that("local linear causal forests with large lambda are equivalent to causa
...
+  ## regression forest for propensity score
+  forest.rep = regression_forest(X.rep, W.rep, num.trees=num.trees, seed = seed)
+  forest.weight = regression_forest(X[rep(1:n, 2),], c(W.a,W.b),
+    sample.weights=sample.weights, num.trees=num.trees, seed = seed)
+  p.rep = predict(forest.rep,X)$predictions
+
+  p.weight = predict(forest.weight,X)$predictions
+  expect_true(mean((p.rep - p.weight)^2)/ mean(p.rep^2) < .1)
+
+  ## causal forest
```
swager: Let's not re-use variable names in this section. That's just asking for a bug to pop up from someone who doesn't notice this editing the code later on.

r-package/grf/tests/testthat/test_causal_forest.R
```diff
@@ -125,12 +125,105 @@ test_that("local linear causal forests with large lambda are equivalent to causa
...
+  sample.weights = c(rep(4,n), rep(1,n))
+
+  ## regression forest for propensity score
+  forest.rep = regression_forest(X.rep, W.rep, num.trees=num.trees, seed = seed)
+  forest.weight = regression_forest(X[rep(1:n, 2),], c(W.a,W.b),
+    sample.weights=sample.weights, num.trees=num.trees, seed = seed)
+  p.rep = predict(forest.rep,X)$predictions
+
+  p.weight = predict(forest.weight,X)$predictions
+  expect_true(mean((p.rep - p.weight)^2)/ mean(p.rep^2) < .1)
```
swager: Did you mean to divide by `var(p.rep)` instead?

r-package/grf/tests/testthat/test_causal_forest.R
```diff
@@ -125,12 +125,105 @@ test_that("local linear causal forests with large lambda are equivalent to causa
...
+   weighted.forest = causal_forest(X[cc,], Y[cc], W[cc], sample.weights = sample.weights[cc], num.trees = num.trees, seed = seed)
+   boosted.weighted.forest = causal_forest(X[cc,], Y[cc], W[cc], sample.weights = sample.weights[cc], orthog.boosting = TRUE, num.trees = num.trees, seed = seed)
+   mse = function(f) { sum((predict(f,X) - tau)^2) }
+   expect_true(mse(weighted.forest) < .9 * mse(forest))
+   expect_true(mse(boosted.weighted.forest)  < .9 * mse(boosted.forest))
+
+   set.seed(old.seed)
+})
+
+test_that("Weighting is roughly equivalent to replication of samples.", {
```
swager: Can you try running this test with a "wrong" forest (i.e., that uses the small dataset but without weighting), and confirm that the test fails?

r-package/grf/tests/testthat/test_forest_summaries.R
```diff
@@ -39,13 +39,41 @@ test_that("causal forest calibration is reasonable with no heterogeneous effect"
   X = matrix(rnorm(n*p), n, p)
   W = rbinom(n, 1, 0.25 + 0.5 * (X[,1] > 0))
   Y = pmax(X[,2], 0) + W + rnorm(n)
-  
+
   cf = causal_forest(X, Y, W,
-                     W.hat = 0.25 + 0.5 * (X[,1] > 0),
-                     Y.hat = 0.25 + 0.5 * (X[,1] > 0) + pmax(X[,2], 0),
+  W.hat = 0.25 + 0.5 * (X[,1] > 0),
```
swager: Could you please fix the indentation here?

r-package/grf/tests/testthat/test_causal_forest.R
```diff
@@ -125,12 +125,105 @@ test_that("local linear causal forests with large lambda are equivalent to causa
...
+   e.cc = 1/(1+exp(-2*X[,1]))
+   sample.weights = 1/e.cc
+
+   forest.1 = causal_forest(X, Y, W, sample.weights = sample.weights, seed=1)
+   forest.2 = causal_forest(X, Y, W, sample.weights = 40*sample.weights, seed=1)
+   expect_true(all(abs(forest.2$predictions - forest.1$predictions) < 1e-10))
+})
+
+test_that("IPCC weighting in the training of a causal forest with missing data improves its complete-data MSE.", {
+   old.seed = .Random.seed
```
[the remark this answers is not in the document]
davidahirshberg: It's either this, large sample sizes, or tests that fail on a single platform with fairly small probability.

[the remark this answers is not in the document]
swager: Of course we need to set the seed. What I'm asking about is why we need to do `old.seed = .Random.seed`, and everything that follows. As opposed to just putting `set.seed(1)` at the top of the test.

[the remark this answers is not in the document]
davidahirshberg: Oh. I wanted to avoid setting the seed for every other test.  Maybe testthat takes care of that automatically, but I couldn't find any documentation suggesting that.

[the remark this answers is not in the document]
swager: Actually, just remembered, we currently set the seed at the top of every test file. So there shouldn't be a need to set the seed further inside any specific test.

r-package/grf/tests/testthat/test_average_effect.R
```diff
@@ -289,13 +289,39 @@ test_that("cluster robust average effects do weighting correctly", {
   
 })
 
-test_that("average effect estimation doesn't error on data with a single feature", {
```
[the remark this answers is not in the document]
davidahirshberg: It was a mistake. It'll be back.

r-package/grf/tests/testthat/test_average_effect.R
```diff
@@ -289,13 +289,39 @@ test_that("cluster robust average effects do weighting correctly", {
...
   X = matrix(rnorm(n * p), n, p)
-  Y = rnorm(n)
-  W = rbinom(n, size=1, prob=0.5)
+  e = 1/(1+exp(-1*X[,1]+1*X[,2]))
+  W = rbinom(n, 1, e)
+  Y = tau * W + 2 * rnorm(n)
+
+  e.cc = 1/( 1+exp(-X[,1]*(cluster.sizes[clust] >= median(cluster.sizes)))  )
+  cc = as.logical(rbinom(n, 1, e.cc))
+  sample.weights = 1/e.cc
```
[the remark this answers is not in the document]
davidahirshberg: It looks to me like true.ate is the ate over the units in large clusters. Am I missing something?

r-package/grf/tests/testthat/test_causal_forest.R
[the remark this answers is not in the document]
davidahirshberg: Thanks, hadn't noticed.

davidahirshberg:
I made the recommended changes. 

Travis is reporting an invalid memory read in CausalForestTest.cpp. I don't get that locally,  so it could take some time to track it down. It tests an invariant that we also test in the R tests, so it may be more practical to disable/remove CausalForestTest.cpp. Thoughts?

REFERENCE.md
jtibshirani: Pinging on this again.

REFERENCE.md
```diff
@@ -210,6 +211,14 @@ When clustering is enabled during training, all subsampling procedures operate o
...
+
+When the distribution of data that you observe is not representative of the population you are interested in scientifically, it can be important to adjust for this. We can pass `sample.weights` to specify that in our population of interest, we observe Xi with probability proportional to `sample.weights[i]`. By default, these weights are constant, meaning that our population of interest is the population from which X1 ... Xn are sampled. For causal validity, the weights we use should not be confounded with the potential outcomes --- typically this is done by having them be a function of Xi. One common example is inverse probability of complete case weighting to adjust for missing data, which allows us to work only the complete cases [the units with nothing missing] to estimate properties of the full data distribution [all units as if nothing were missing].
```
jtibshirani: We should use standard parentheses `(` instead of square ones.

core/test/forest/CausalForestTest.cpp
```diff
@@ -0,0 +1,78 @@
...
+#include "forest/ForestTrainers.h"
+#include "utilities/ForestTestUtilities.h"
+
+#include "catch.hpp"
+
+
+TEST_CASE("causal forests are invariant to rescaling of the sample weights", "[causal, forest]") {
+  // Run the original forest.
+  // we'll overwrite a covariate in the original data with sample weights so we needn't resize the data.
+  uint weight_index = 9;
```
jtibshirani: These should be `size_t` instead of `uint`.

core/test/forest/CausalForestTest.cpp
```diff
@@ -0,0 +1,78 @@
...
+  uint treatment_index = 11;
+  Data* data = load_data("test/forest/resources/causal_data.csv");
+  data->set_weight_index(weight_index);
+  data->set_outcome_index(outcome_index);
+  data->set_treatment_index(treatment_index);
+  data->set_instrument_index(treatment_index);
+
+  bool error;
+  for(size_t r = 0; r < data->get_num_rows(); r++) {
+    double weight = 1.0 / (1.0 + exp(- data->get(1, r)));
```
jtibshirani: We're indexing by `(column, row)`, but the method signature is `double get(size_t row, size_t col)`, so this is going to produce an invalid read.

davidahirshberg: I hadn't noticed, thanks. Isn't it odd that the order for get and set are reversed?

```
double DefaultData::get(size_t row, size_t col) const {
  return data[col * num_rows + row];
}
void DefaultData::set(size_t col, size_t row, double value, bool& error) {
  data[col * num_rows + row] = value;
}
```

davidahirshberg: I flipped the arguments for the call to get in the test. Good to merge?

PR #423, opened 2019-05-20
Allow roxygen to autogenerate the NAMESPACE file.

jtibshirani:
Previously, we were modifying the NAMESPACE file by hand, despite using roxygen
in the build procedure. This commit adds the necessary roxygen statements to
the documentation to allow the NAMESPACE to be automatically generated.


jtibshirani:
Thanks @swager :)

PR #425, opened 2019-05-22
Update algorithm reference for boosting

evanmunro:


REFERENCE.md
```diff
@@ -199,6 +200,26 @@ The cross-validation procedure works as follows:
...
 - Finally, given the debiased error estimates for each set of parameters, we apply a smoothing function to determine the optimal parameter values.
 
+
+### Boosting Random Forests
+
+**Note:** this feature is currently marked 'experimental'. Its implementation may change in future releases.
+
+Ghosal and Hooker (2018), show how a boosting step can reduce bias in random forest predictions. We have included a  `boosted_regression_forest` feature, which trains a series of forests, each of which are trained on the OOB residuals from the previous step. `boosted_regression_forest` includes the same parameters as `regression_forest`, which are passed directly to the forest trained in each step. It also includes control over the step selection procedure:
+
+- If `boost.steps` is not specified, then a cross-validation procedure selects the number of boosting steps automatically, where the next step is only taken if the estimated OOB error for the next step, computed using a small forest of size `boost.trees.tune`, has an estimated OOB error reduction from  the previous step of more than `boost.error.reduction`. For computational considerations only up to `boost.max.steps` are considered.
```
jtibshirani: I think this may be too concise to be easily understandable to someone new to the algorithm. A suggestion for how to restructure this section (starting from the phrase 'It also includes control over the step selection procedure:'):

"By default, the number of boosting steps is automatically chosen through a cross-validation procedure. This procedure works as follows:
- First, a single `regression_forest` is trained.
- To decide if another step should be taken, we train a small forest ...
- If another step should be taken, then we train a full-sized `regression_forest` on the residuals and add it to the boosted forest.
- This process continues until `boost.error.reduction` cannot be met when training the mini-forest, or if the total number of steps exceeds the limit `boost.max.steps`.

If you would like to skip the cross-validation procedure and specify the number of steps directly, then you can pass the parameter `boost.steps`."

Also, can we add a sentence discussing how `tune.parameters` is handled with boosting?

evanmunro: thanks, this is much better

REFERENCE.md
```diff
@@ -205,12 +205,18 @@ The cross-validation procedure works as follows:
...
 
-- If `boost.steps` is not specified, then a cross-validation procedure selects the number of boosting steps automatically, where the next step is only taken if the estimated OOB error for the next step, computed using a small forest of size `boost.trees.tune`, has an estimated OOB error reduction from  the previous step of more than `boost.error.reduction`. For computational considerations only up to `boost.max.steps` are considered.
-- `boost.steps` can also be specified by the user.
+For computational considerations, if `tune.parameters=TRUE`, then parameters are selected by the `regression_forest` cross-validation only once, in the first step. The selected parameters are then applied to forests in any further steps, without repeating cross-validation. `boosted_regression_forest` also includes control over the step selection procedure. By default, the number of boosting steps is automatically chosen through a cross-validation procedure. This procedure works as follows:
```
jtibshirani: One last comment: could we move discussion of `tune.parameters` after the description of the boosting procedure? Tuning is not actually enabled by default and I think it is a detail that's best left to the end. Also do you mean "selected only using the first boosting step, when a complete regression forest is trained" as opposed to "selected by the `regression_forest` cross-validation only once"? Because this fact applies to boosting no matter whether cross-validation or a fixed number of boosting steps is used.

REFERENCE.md
```diff
@@ -205,12 +205,18 @@ The cross-validation procedure works as follows:
...
 
-`causal_forest` also provides a `orthog.boosting` flag, which, if set to `TRUE`, orthogonalizes  the causal forest using `boosted_regression_forest` instead of `tune_regression_forest` to estimate `e(x) = E[W|X=x]` and `m(x) = E[Y|X=x]`, with the number of steps set by cross-validation.
+- First, a single regression_forest is trained and added to the boosted forest.
```
jtibshirani: regression_forest -> `regression_forest`

REFERENCE.md
```diff
@@ -205,12 +205,18 @@ The cross-validation procedure works as follows:
...
+- To decide if another step should be taken, we train a small forest ...
```
jtibshirani: Oops, I didn't actually mean that we should put an ellipsis here, I was just giving a sense of the structure. We should mention here the parameter that controls the size of the small forest, and also how we check the OOB predictions of the small forest against `boost.error.reduction`.

REFERENCE.md
evanmunro: I meant the tuning procedure in regression_forest, not the cross validation for boosting. Changed the wording and hopefully is more clear now.

Issue #424, opened 2019-05-22
regression_forest does not accept a data frame as X


swager:
It appears that we are no longer able to handle design matrices that are data frames, as seen from the R-output below:
```R> n = 100; p = 5
> X = data.frame(matrix(rnorm(n*p), n, p))
> Y = rnorm(n)
> rf = regression_forest(X, Y)
Error in data.frame(..., check.names = FALSE) : 
  arguments imply differing number of rows: 100, 0

> rf = regression_forest(as.matrix(X), Y)
> print(rf)
GRF forest object of type regression_forest 
Number of trees:  2000 
Number of training samples: 100 
Variable importance: 
    1     2     3     4     5 
0.351 0.171 0.136 0.186 0.156 
```

halflearned:
I was just staring at this bug in my branch, thinking it was something I had done.

The culprit is [this line](https://github.com/grf-labs/grf/blob/62d925cbe3f2ebfdc225e6d014fadc62d02ba030/r-package/grf/R/input_utilities.R#L229). Apparently, we can `cbind` a matrix and a NULL (in which case the NULL is ignored), but not a data.frame and a NULL (in which case an error is raised). 

```
X1 <- data.frame(matrix(rnorm(n*p), n, p))
X2 <- as.matrix(X1)
sample.weights <- NULL
cbind(X1, sample.weights) # Bad!
cbind(X2, sample.weights) # Good (?!)
```

PR #426, opened 2019-05-23
create_data_matrices should allow for NULL inputs

halflearned:
Solves #424 by ensuring that the argument to `cbind` is indeed a matrix.


halflearned:
Added short test!

r-package/grf/tests/testthat/test_input_utilities.R
```diff
@@ -0,0 +1,28 @@
...
+  n = 100
+  p = 2
+  Xm = matrix(rnorm(n*p), n, p)
+  Xd = as.data.frame(Xm)
+  Y = rnorm(n)
+  W = rbinom(0.5, size=1, n=n)
+  Z = runif(n)
+  
+  # Matrix input
+  rfm = regression_forest(Xm, Y, num.trees=5, seed=1234)
```
jtibshirani: I was thinking we could add a unit (not integration) test, and just check the output of `create_data_matrices` directly instead of training forests. I'm hoping to move more towards unit tests, to make debugging test failures easier and cut down on the long R build times we've been seeing.

halflearned: Makes sense! I'll update this.

jtibshirani:
@halflearned just a friendly ping that it would be great to get this merged. I manually included in the fix in the 0.10.3 release, so I am eager to make sure it's properly incorporated into master.

Issue #427, opened 2019-05-26
Function 'enterRNGScope' not provided by package 'Rcpp'


halflearned:
On my machine (macOS, some recent clang), the current `grf` version on the master branch gives me this error:

```
> library(grf)
> n = 100
> p = 6
> X = matrix(rnorm(n*p), n, p)
> Y = rnorm(n)
> rf = regression_forest(X, Y)
Error in regression_train(data$default, data$sparse, outcome.index, sample.weight.index,  : 
  function 'enterRNGScope' not provided by package 'Rcpp'
```

This doesn't seem to affect all machines. I can't get it to reproduce on Stanford's Sherlock cluster (CentOS, gcc 4.8.5), or on my Kali vm (with gcc 8.3.0). The Rcpp version was the same for all these machines.

It [seems](http://lists.r-forge.r-project.org/pipermail/rcpp-devel/2014-February/007120.html) that this may affect systems sometimes, and that the solution is to have at least on Rcpp symbol imported in the NAMESPACE file. Indeed, the problem goes away if I add `importFrom(Rcpp, evalCpp)` to it.

PR #428, opened 2019-05-26
Adds an Rcpp symbol to NAMESPACE

halflearned:
Including any Rcpp symbol in NAMESPACE file seems to avoid the issue discussed in #427.

Following [this](http://lists.r-forge.r-project.org/pipermail/rcpp-devel/2014-February/007120.html) suggestion by Rcpp's author.


jtibshirani:
Thanks a lot for catching and fixing this. Because we now automatically generate the `NAMESPACE` file, I think we should add an `importsFrom` directive as a roxygen comment instead. The place I have been putting singleton comments is `regression_forest` (I just had to pick some method).

Issue #379, opened 2019-03-15
Stability of grf routines across machines when setting seed


tcovert:
I am having trouble writing R code which calls grf routines and generates the same estimates across different machines.

Here is an MWE, inspired by the basic code in the GRF README.md:
```
library(grf)

rfseed <- 19310821
set.seed(rfseed)

temp <- runif(1)

# Generate data.
n = 2000; p = 10
X = matrix(rnorm(n*p), n, p)
X.test = matrix(0, 101, p)
X.test[,1] = seq(-2, 2, length.out = 101)
Z = (runif(n) > 0.5) * rlnorm(n, sdlog = 2)

# Train a causal forest.
W = rbinom(n, 1, 0.4 + 0.2 * (X[,1] > 0))
Y = pmax(X[,1], 0) * W + X[,2] + pmin(X[,3], 0) + rnorm(n) + Z

tau.forest = causal_forest(X, Y, W, seed = rfseed)

ow_ate <- average_treatment_effect(tau.forest, target = "overlap")
``` 

All I've done here is (1) fix the R random number seed, (2) run a random number generator one time to verify that I get the same R-based random numbers across machines (this is the `temp` line), (3) add quite a bit of noise to the outcomes, so that this data is similarly skewed as the real data I am working with, (4) set the C++ random number seed in the `causal_forest` routine to a fixed value, and (5) using the overlap-weighted average treatment effect option to get point estimates and standard errors, to pick something easy.

On my machine (2018 MacBook Pro, running High Sierra, R 3.5.1 and grf version 0.10.2), I get 
```
> ow_ate
  estimate    std.err 
-0.9212784  1.3445339 
```

On my co-author's machine (PC, Windows 7 x64 (build 7601), R 3.5.1, grf version 0.10.2), we get
```
> ow_ate
 estimate    std.err
-0.8316843  1.2665160
```

Note, my co-author also has a Windows 10 laptop, and gets the same answer there (which is not the same as what my research assistant gets below).

On my research assistant's machine (PC, Windows 10 pro version 1803, R 3.5.1, grf version 0.10.2), we get
```
> ow_ate
 estimate    std.err
-0.8432834  1.2506999
```

Any suggestions about how to make this example run equivalently across machines?

tcovert:
[...]

PR #435, opened 2019-06-13
Platform-independent random number seeds

halflearned:
Hopefully solves #379. 

As @jtibshirani pointed out, although `std::mt19973` produces random numbers that are stable across platforms, the implementations of several other dependencies were platform-dependent. 

Highlight of major changes: 

+ Replacing most of the random header, we are now using [kthohr/stats](https://github.com/kthohr/stats) which has been copied to `third_party/`. 

+ Our version of `stats` has been slightly changed. For uniform real numbers it actually ended up calling the `random` [here](https://github.com/kthohr/stats/blob/8545bf8713d9529baa7b2e0c8eed23e6e93b9039/include/stats_incl/rand/runif.ipp#L43), defeating the purpose. So in our version that bit is replaced by the corresponding code in [boost](https://www.boost.org/doc/libs/1_64_0/boost/random/uniform_real_distribution.hpp).

+ `stats` also didn't support discrete distributions for vectors, so [draw_weighted](https://github.com/halflearned/grf/blob/675d93f424d35b2546d5ca8a6dec7389a38b7372/core/src/sampling/RandomSampler.cpp#L201) has some custom code based on [numpy.random.choice](https://github.com/numpy/numpy/blob/ff242911601f4b815f572fd105ca4f35fb2ec49b/numpy/random/mtrand.pyx#L679)

+ The implementation of `shuffle` from the `algorithm` header was also platform-dependent, so now we have method [`RandomSampler::shuffle`](https://github.com/halflearned/grf/blob/675d93f424d35b2546d5ca8a6dec7389a38b7372/core/src/sampling/RandomSampler.cpp#L229) also based on the corresponding method in boost.

+ Finally in order to stabilize seeding across threads some changes [`ForestTrainer::train_batch`](https://github.com/halflearned/grf/blob/675d93f424d35b2546d5ca8a6dec7389a38b7372/core/src/forest/ForestTrainer.cpp#L64) were also necessary

I tested this with a few versions of gcc and clang and the results are sensible, but I'll try to do some additional testing before merging.

Update: now all tests are seeded.


halflearned:
Retracting this PR. I tested on different machines and found discrepancies in results. Will revise my approach and submit a new PR.

PR #436, opened 2019-06-14
Update input parameter to samples_per_cluster

jakobzeitler:
Seems to have been changed in commit 12a179f6136c7bb6146f0b9fa828ac3e02c8a280 (parent c0bf593), but was not reflect in this example script. Will throw a "unused argument (samples.per.cluster = 50)" otherwise.


halflearned:
Thanks @jakobzeitler!

jtibshirani:
@jakobzeitler the parameter `samples_per_cluster` was renamed to `samples.per.cluster` in 0.10.3. If you try updating to the latest version of grf, the script should work as-is. Feel free to open a new issue or PR if you encounter a problem with the script after updating.

jakobzeitler:
My bad, seems like I had an older grf loaded!

PR #437, opened 2019-06-14
Add lintr to CI and bring existing r-source up to standard

erikcs:
This PR adds [lintr](https://github.com/jimhester/lintr) as a CI step

The style is default "tidyverse" R standard with the exception of 5 options

* line_length_linter(120)
    Stick to lines no longer than 120 characters

* object_name_linter = NULL
    lintr cannot do this kind of semantic analysis out of the box and check if an object is a variable or a function, so one cannot use this to check the grf style with variable.names.with.dots and_function_names_with_underscores. This style is very evident from existing code, and should not be hard to keep in mind. 

* commented_code_linter = NULL
   Misc. default linter that will produce false positives.

* object_usage_linter = NULL
    Default linter which tries to do semantic analysis but most often produces false positives (i.e: is a local variable used). Disabled for less clutter.

*  spaces_left_parentheses_linter = NULL
   Can give annoying false positives.

(more checks can be ignored here in the future (for potentially annoying complaints), or one could add a `# nolint` directly in the source code  (lintr and styler are integrated with RStudio and will automatically replace `=` with `<-`). A note on this is added to [DEVELOPING.md](https://github.com/grf-labs/grf/blob/3514716ce070c925a64f9ea60ebd389218ffb0f5/DEVELOPING.md))

The linter is run as the first step in [build_package.R](https://github.com/grf-labs/grf/blob/3514716ce070c925a64f9ea60ebd389218ffb0f5/r-package/build_package.R), and and quits immediately if this fails. 

This PR also brings the existing R source up to this standard by touching a large number of files:
(original lint output for [master 15ee993](https://gist.github.com/erikcs/93a3c6644ff06d5cbf40c6b84f657dd1))

* dfd4546
    Does most of the automatic code formatting using [styler](https://github.com/r-lib/styler)
    Mostly consistently using `<-` as the assignment operator and removing superfluous whitespace (will make diffs easier to read in the future), consistent indentation, etc.

* 3bec174
    Manually fixes the few > 120 line length cases

* 9a608a5
   Manually fixes the few remaining ones (2)
    (trailing whitespace, a comment after opening curly bracket)

* 3d05408
   One final `styler::style_pkg()` (intermediate modifications above that pass the linter are reformatted)


.travis.yml
```diff
@@ -34,10 +36,13 @@ before_install:
   - sudo apt-get install -y libopencv-dev valgrind
 
 script:
+  # Checks for core C++
   - cd ../../core/
   - mkdir build && cd build
   - cmake -DCMAKE_CXX_COMPILER=$COMPILER .. && make
   - cd ..
   - valgrind --leak-check=full --error-exitcode=1 ./build/grf exclude:[characterization]
+  # Checks for R package
```
halflearned: Small potential improvement here: would it be possible to `lint` before `build`ing?

r-package/lint_package.R
```diff
@@ -0,0 +1,15 @@
+Sys.setenv(LINTR_COMMENT_BOT='false')
```
jtibshirani: 👍 I haven't found those bots to be useful in the past, and they can make PRs harder to read.

r-package/build_package.R
```diff
@@ -6,15 +6,31 @@
 # To prepare a CRAN build:
 #   `Rscript build_package.R --as-cran`
 
+Sys.setenv(LINTR_COMMENT_BOT="false")
```
jtibshirani: I forgot to mention this last time -- I think we should move this to the travis script  through something like `export LINTR_COMMENT_BOT=false`, because it's not relevant when building locally, and it's nice to avoid touching local environment variables when possible.

erikcs:
> * Does the linter verify that only 2 spaces are used for indentation, as opposed to 4? I couldn't find this in the `lintr` documentation.

No, sadly not.

> * Would it be hard to get `spaces_left_parentheses_linter` to work, or does it seem too broken? It is just a nice check to have, as it is not something that is obvious to newcomers.

It can give annoying false positives, here is an example:
Screen Shot 2019-06-25 at 08 35 31
(the check could be disabled with `# nolint` after that line)

r-package/build_package.R
```diff
@@ -11,10 +11,25 @@ library(Rcpp)
...
+# Check code style consistency
+linters <- with_defaults(
+  line_length_linter(120), # Max line length = 120
+  object_name_linter = NULL, # To allow variable.names + function_names
+  commented_code_linter = NULL, # Misc. false positives
+  object_usage_linter = NULL, # Misc. false positives
+  spaces_left_parentheses_linter = NULL, # Misc. false positives
+  default = default_linters);
+
+lint_res <- lint_package(path = "grf/", linters = linters, exclusions = "grf/R/RcppExports.R")
```
jtibshirani: Perhaps we could use `package.name` here instead of `"grf/"`.

PR #439, opened 2019-06-16
Allowing number of trees in local centering step to vary

### LINKED ISSUE #270: number of trees in local centering step
Hey guys,

I was thinking that we may want to give users an option to specify the number of trees used in the local centering step. I thought the parameter could num.lc.trees. The default could be still be min(num.trees, 500)), keeping the current number of trees.

```r
causal_forest <- function(X, Y, W,
                          Y.hat = NULL,
                          W.hat = NULL,
                          sample.fraction = 0.5,
                          mtry = NULL,
                          num.trees = 2000,
                          num.threads = NULL,
                          min.node.size = NULL,
                          honesty = TRUE,
                          ci.group.size = 2,
                          alpha = NULL,
                          imbalance.penalty = NULL,
                          stabilize.splits = TRUE,
                          compute.oob.predictions = TRUE,
                          seed = NULL,
                          clusters = NULL,
                          samples_per_cluster = NULL,
                          tune.parameters = FALSE,
                          num.lc.trees = min(num.trees, 500), 
                          num.fit.trees = 200,
                          num.fit.reps = 50,
                          num.optimize.reps = 1000) 
```

Should I submit a PR? Thanks

halflearned:
Closes #270, allowing the number of trees in the "local centering" step to be an increasing function of `num.trees`.

Thank you to @klosins for simulations. We'll update the issue with her simulation results, but we have found basically no detectable difference between the four functions suggested in the issue.

### DIFF
--- a/r-package/grf/R/causal_forest.R
+++ b/r-package/grf/R/causal_forest.R
@@ -152,18 +152,19 @@ causal_forest <- function(X, Y, W,
     clusters <- validate_clusters(clusters, X)
     samples.per.cluster <- validate_samples_per_cluster(samples.per.cluster, clusters)
     honesty.fraction <- validate_honesty_fraction(honesty.fraction, honesty)
+    num.trees.orthog <- max(50, num.trees / 4)
 
     reduced.form.weight <- 0
 
     if (is.null(Y.hat) && !orthog.boosting) {
       forest.Y <- regression_forest(X, Y, sample.weights = sample.weights, sample.fraction = sample.fraction, mtry = mtry, tune.parameters = tune.parameters,
-                                    num.trees = max(50, num.trees / 4), num.threads = num.threads, min.node.size = NULL, honesty = TRUE,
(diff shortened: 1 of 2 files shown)


r-package/grf/R/causal_forest.R
```diff
@@ -157,13 +157,13 @@ causal_forest <- function(X, Y, W,
 
     if (is.null(Y.hat) && !orthog.boosting) {
       forest.Y <- regression_forest(X, Y, sample.weights = sample.weights, sample.fraction = sample.fraction, mtry = mtry, tune.parameters = tune.parameters,
-                                    num.trees = min(500, num.trees), num.threads = num.threads, min.node.size = NULL, honesty = TRUE,
```
jtibshirani: It would be nice to pull this out into a variable, to avoid repeating the transformation in several different places.

jtibshirani:
@halflearned great that we got this resolved! Are you or @klosins still planning to upload the simulation results? It would be nice to have those as a reference for our decision.

PR #441, opened 2019-06-16
CI: compile R extension with clang

erikcs:
@halflearned pointed out that in some cases his core C++ compilation would fail, but the R build would pass, that was because the C++ failed with clang, but the R package was built with GNU g++ 

With the R Makevars R CMD runs `g++`:
Screen Shot 2019-06-15 at 21 43 48

Which in both build matrices is the GNU compiler
Screen Shot 2019-06-15 at 21 43 33

Solution: alias g++ to clang++-3.7 in the clang matrix:
Screen Shot 2019-06-15 at 22 03 14

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,6 +27,7 @@ matrix:
           packages:
             - clang-3.7
       env: COMPILER=clang++-3.7
+      before_script: shopt -s expand_aliases && alias g++=clang++-3.7
 
 before_install:
   - cd r-package/grf


erikcs:
tiny aside:

* [core Makefile](https://github.com/grf-labs/grf/blob/master/core/CMakeLists.txt)
    core c++ is buildt with some modified linking arguments for pthreads
    compiled shared object is around 1 mb.

* [R Makevars](https://github.com/grf-labs/grf/blob/master/r-package/grf/bindings/Makevars)
   does not have anything for pthreads specified
   however, threading works fine 
   compiled shared object is around 750 kb
   (is dynamically linked to the same things as the above (`otool -L`) with the exception of Rcpp)

jtibshirani:
We compile the core C++ with both `g++` and `clang`, so for an error to pass through CI, it must (1) only be in the C++ bindings file, and (2) compile under `g++` but not `clang`. To me this seems like an edge case, and I'm not sure it's worth complicating our build to address it (to me it's fairly hacky to be aliasing `g++` to `clang` on the build machines).

erikcs:
Ok, that makes sense, I'll go ahead and close this.

PR #442, opened 2019-06-20
Add ll causal variance

rinafriedberg:
This PR includes a basic variance estimate for local linear causal forests. The variance estimate follows the local linear forest method, with algebra adjusted for the local linear causal equation (see LLF paper equation 11 and section 4.1). There is also a simple test to verify that the standardized errors are reasonable, following tests on instrumental and regression forests.


r-package/grf/tests/testthat/test_confidence_intervals.R
```diff
@@ -93,3 +93,27 @@ test_that("instrumental CIs are invariant to scaling Z", {
...
+   X = matrix(rnorm(n*p), n, p)
+   W = rbinom(n, 1, 0.5)
+   TAU = 2*X[,1] + X[,2]
+   Y = W * TAU + 0.5 * rnorm(n)
+
+   forest = causal_forest(X, Y, W)
+   tau.hat = predict(forest, linear.correction.variables = 1:ncol(X), estimate.variance = TRUE)
+   error.standardized = (tau.hat$predictions - TAU) / sqrt(tau.hat$variance.estimates)
+
+   expect_true(mean(abs(error.standardized) > qnorm(0.975)) <= 0.16)
```
halflearned: [Super minor comment] Could we use `expect_lt(a, b)` instead of `expect_true`? I've been trying to use that more often, because the test fails it's useful to see how wrong it really was.

rinafriedberg: Thanks @halflearned ! Committed that change.

core/src/prediction/LLCausalPredictionStrategy.cpp
```diff
@@ -147,8 +150,152 @@ std::vector<double> LLCausalPredictionStrategy::compute_variance(
         size_t sampleID,
         std::vector<std::vector<size_t>> samples_by_tree,
         std::unordered_map<size_t, double> weights_by_sampleID,
-        const Data* train_data,
-        const Data* data,
+        const Data* original_data,
```
jtibshirani: Small comment, maybe we could stick with `train_data` and `data`, since that's what we use in `LocalLinearPredictionStrategy` (and fix the naming in the other parts of this file)?

core/src/prediction/LLCausalPredictionStrategy.cpp
```diff
@@ -92,8 +93,10 @@ std::vector<double> LLCausalPredictionStrategy::predict(
     for (size_t j = 0; j < num_variables; ++j){
       size_t current_predictor = linear_correction_variables[j];
       // X - x0 column
-      X(i,j+1) = test_data->get(sampleID, current_predictor)
-                 - original_data->get(index, current_predictor);
+      //X(i,j+1) = test_data->get(sampleID, current_predictor)
```
jtibshirani: It'd be good to remove this commented out line. Also, was the calculation here incorrect before, or does the sign turn out not to matter? If it was actually incorrect, it'd be good to make sure we have a test to cover the behavior.

core/src/prediction/LLCausalPredictionStrategy.cpp
```diff
@@ -147,8 +150,152 @@ std::vector<double> LLCausalPredictionStrategy::compute_variance(
...
+
+  for (size_t i = 0; i < num_nonzero_weights; ++i) {
+    // Index of next neighbor with nonzero weights
+
+    size_t index = indices[i];
+    double treatment = original_data->get_treatment(index);
+
+    // Intercept
+    X(i, 0) = 1;
+    //X_alone(i, 0) = 1;
```
jtibshirani: Another commented out line that we could remove.

core/src/prediction/LLCausalPredictionStrategy.cpp
```diff
@@ -147,8 +150,152 @@ std::vector<double> LLCausalPredictionStrategy::compute_variance(
...
+    X(i, treatment_index) = treatment;
+
+    // Outcome (just copied)
+    Y(i) = original_data->get_outcome(index);
+  }
+
+  // find ridge regression predictions
+  Eigen::MatrixXd M_unpenalized (dim_X, dim_X);
+  M_unpenalized.noalias() = X.transpose() * weights_vec.asDiagonal() * X;
+  double normalization = M_unpenalized.trace() / dim_X;
```
jtibshirani: Small comment, we could move this into the `if (!weight_penalty) { ... }` block to avoid calculating the normalization when using a weighted penalty.

core/src/prediction/LLCausalPredictionStrategy.cpp
```diff
@@ -147,8 +150,152 @@ std::vector<double> LLCausalPredictionStrategy::compute_variance(
...
+
+  Eigen::MatrixXd M;
+  M = M_unpenalized;
+  if (!weight_penalty) {
+    // standard ridge penalty
+    for (size_t j = 1; j < treatment_index; ++j){
```
jtibshirani: Small comment, but I think it'd be clearer to have a single loop, and then add an if statement inside (`if (j != treatment_index) { ... }`). When reading over it, I didn't immediately understand why there were two loops.

core/src/prediction/LLCausalPredictionStrategy.cpp
rinafriedberg: You can formulate local linear regression either way; I changed it to be consistent with local linear prediction strategy.

jtibshirani: Got it, thanks.

core/src/prediction/LLCausalPredictionStrategy.cpp
```diff
@@ -92,8 +93,8 @@ std::vector<double> LLCausalPredictionStrategy::predict(
     for (size_t j = 0; j < num_variables; ++j){
       size_t current_predictor = linear_correction_variables[j];
       // X - x0 column
-      X(i,j+1) = test_data->get(sampleID, current_predictor)
-                 - original_data->get(index, current_predictor);
+      X(i,j+1) = train_data->get(index, current_predictor) -
```
swager: Do we mean to be flipping the sign here? Was this a bug before (or does this not matter)?

r-package/grf/tests/testthat/test_confidence_intervals.R
```diff
@@ -93,3 +93,27 @@ test_that("instrumental CIs are invariant to scaling Z", {
...
+   X = matrix(rnorm(n*p), n, p)
+   W = rbinom(n, 1, 0.5)
+   TAU = 2*X[,1] + X[,2]
+   Y = W * TAU + 0.5 * rnorm(n)
+
+   forest = causal_forest(X, Y, W)
+   tau.hat = predict(forest, linear.correction.variables = 1:ncol(X), estimate.variance = TRUE)
+   error.standardized = (tau.hat$predictions - TAU) / sqrt(tau.hat$variance.estimates)
+
+   expect_lt(mean(abs(error.standardized) > qnorm(0.975)), 0.16)
```
swager: This is not very good coverage... The DGP in use here looks pretty difficult. Can we use an easier DGP for the test that gets better coverage? (Also, can we add a test that replicates what we found in the paper, i.e., with smoothness, LLCF can get better coverage with shorter CIs relative to a basic CF.)

core/src/prediction/LLCausalPredictionStrategy.cpp
```diff
@@ -148,7 +148,149 @@ std::vector<double> LLCausalPredictionStrategy::compute_variance(
...
+  Eigen::VectorXd e_trt = Eigen::VectorXd::Zero(dim_X);
+  e_trt(treatment_index) = 1.0;
+  Eigen::VectorXd zeta = M.ldlt().solve(e_trt);
+
+  Eigen::VectorXd X_times_zeta = X * zeta;
+  Eigen::VectorXd local_prediction = X * theta;
+  Eigen::VectorXd pseudo_residual = Eigen::VectorXd::Zero(num_nonzero_weights);
+
+  for (size_t i = 0; i < num_nonzero_weights; i++) {
+    pseudo_residual(i) = X_times_zeta(i) * (Y(i) - local_prediction(i,1) - X(i, treatment_index)*local_prediction(i, treatment_index));
```
swager: The local linear regression adjustment appears to be missing here?

core/src/prediction/LLCausalPredictionStrategy.cpp
rinafriedberg: Doesn't matter as long as you're consistent (LLR can be written both ways)- I flipped it so it looks the same as the code in local linear prediction strategy to avoid confusion, and double checked that the output is the same!

core/src/prediction/LLCausalPredictionStrategy.cpp
rinafriedberg: I may be thinking through the math incorrectly; but I think it's analogous to the pseudo-residuals for LLF, which are:

pseudo_residual(i) = X_times_zeta(i) * (Y(i) - local_prediction(i));

Here, we just de-mean Y(i) by the local_prediction(0) term - the 1 is a typo I'll fix, forgot about C++ indexing- and then subtract out (W-W.hat)*tau.hat.local. The linear correction goes into our prediction for tau.hat.local, but I think the pseudo-residual should just have the actual prediction.

r-package/grf/tests/testthat/test_confidence_intervals.R
rinafriedberg: Yup addressing this now! Coverage is awesome with an easier DGP and I'm adding a test for coverage/CI length now; will push once it runs.

core/src/prediction/LLCausalPredictionStrategy.cpp
rinafriedberg: @swager let me know what you think! All other comment responses are now in :)

swager: Ok I see -- now I understand what's up here. Just to make sure I'm not missing anything, though, could you please write up a short tex document that spells out the algebra going on here (including the dimension of all the objects)? In particular, what is the dimension of "local_prediction"?

r-package/grf/tests/testthat/test_confidence_intervals.R
swager: Great, thanks!

core/src/prediction/LLCausalPredictionStrategy.cpp
swager: Ok, got it.

core/src/prediction/LLCausalPredictionStrategy.cpp
rinafriedberg: Ok good check - this code was actually a bit more complicated than it should have been! Doc is reticent to attach, but I'll email it to you and try to add it to this conversation for future reference.

rinafriedberg:
Encountering problems with the lintr package in the build (did not encounter them locally); I'll look into this issue now. Otherwise all R tests and C++ tests are passing locally.

rinafriedberg:
@jtibshirani or @halflearned , do you have any feedback on the lintr error? All tests (R and C++) pass locally with no problems, but the build keeps failing with error "The command "RScript -e 'remotes::install_github(c("jimhester/lintr"))'" failed and exited ..." I checked for merges from master and don't see any missing.

core/src/prediction/LLCausalPredictionStrategy.cpp
swager: Got it, thanks! I had been looking at the outdated version of the code above -- that's why I got confused.

rinafriedberg:
File upload: PDF giving a basic derivation for the local linear causal forest variance estimation formula, and linking it to the values in LLCausalPredictionStrategy.cpp file. 

[llcf_variance.pdf](https://github.com/grf-labs/grf/files/3333041/llcf_variance.pdf)

PR #444, opened 2019-06-27
added code to replicate simulations in LLF paper

rinafriedberg:
This PR has a set of simulations to replicate the LLF simulations (and to explain usage for interested users). All of the experiments together take a long time to run, so there is a "main" file that calls all of the others, and individual files to just test one experiment.


experiments/local_linear_examples/friedman_table.R
```diff
@@ -0,0 +1,131 @@
...
+args = expand.grid(n = ns, p = ps, sigma = sigmas)
+full.results = t(apply(args, MARGIN = 1, FUN = function(arguments){
+  print(arguments)
+  mses = simulation.run(arguments[1], arguments[2], arguments[3], num.reps = num.reps)
+  sqrt(mses)
+}))
+full.results = cbind(args, round(full.results, 3))
+full.results
+
+write.csv(full.results,"Dropbox/Local Linear Forests/Simulations/for-repo-eventually/Results/friedman_roundnplusone.csv", row.names = FALSE)
```
jtibshirani: I think this should just create the csv in the current directory, as opposed to `"Dropbox/..."`.

rinafriedberg:
Thanks @jtibshirani ! Not looking for a super detailed code review; just wanted to make sure style is ok and that overall it's a readable PR. Made your suggested changes- thank you as always :)

Issue #309, opened 2018-10-01
Error in t.default(T) : argument is not a matrix


ZhangMengxia:
Dear grf developers,

I keep get the following error:
**Error in t.default(T) : argument is not a matrix**
after I run:
`tau.forest.centered = causal_forest(X, Y_centered, W_centered,tune.parameters = TRUE)
`
I checked all the inputs X, Y_centered, W_centered. They all look normal and do not have NaN. The dimension of X is 57914 by 48. Some columns of X are continuous real numbers. Some columns of X are binary dummy variables.  Dimension of Y_centered is 57914 by 1. Dimension of W_centered is 57914 by 1. Y_centered and W_centered are dependent variable and treatment, respectively. And they are centered by Y_centered=Y-Y.hat and W_centered=W-W.hat.  And Y.hat is computed using the following code, similarly for W.hat.
`forest.Y = regression_forest(X, Y, tune.parameters = TRUE)
Y.hat = predict(forest.Y)$predictions`

The following is the output for sessionInfo() function.

> sessionInfo()
R version 3.5.1 (2018-07-02)
Platform: x86_64-apple-darwin15.6.0 (64-bit)
Running under: OS X El Capitan 10.11.6

Matrix products: default
BLAS: /System/Library/Frameworks/Accelerate.framework/Versions/A/Frameworks/vecLib.framework/Versions/A/libBLAS.dylib
LAPACK: /Library/Frameworks/R.framework/Versions/3.5/Resources/lib/libRlapack.dylib

locale:
[1] en_US.UTF-8/en_US.UTF-8/en_US.UTF-8/C/en_US.UTF-8/en_US.UTF-8

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] survival_2.42-3 plm_1.6-6       Formula_1.2-3   grf_0.10.0     

loaded via a namespace (and not attached):
 [1] bdsmatrix_1.3-3   Rcpp_0.12.18      lattice_0.20-35   zoo_1.8-3         lmtest_0.9-36     MASS_7.3-50       grid_3.5.1        nlme_3.1-137      miscTools_0.6-22  Matrix_1.2-14     sandwich_2.4-0   
[12] splines_3.5.1     tools_3.5.1       compiler_3.5.1    DiceKriging_1.5.5 maxLik_1.3-4   

Best regards,
Mengxia.

ZhangMengxia:
[...]

Issue #278, opened 2018-08-05
Solidify the parameter tuning implementation.


jtibshirani:
To round out the parameter tuning implementation, we should try out the approach on the set of simulations developed in #274, and make sure the following points hold:
- The approach is reasonably fast, and doesn't have a severe impact on running time. A reasonable rule of thumb here might be that it doesn't slow down training by more than 2x.
- The tuning algorithm is solid, and will never cause training to crash or fail to complete. Tuning should also never hurt the forest's statistical performance.
- It performs reasonably well compared to other viable approaches, and we are not missing any obvious low-hanging fruit.

The overall goal is for tuning to be fast and stable enough that we feel comfortable enabling it by default (`tune.parameters = TRUE`).

Note that as part of this work, we'll need to address the instability that @halflearned identified in Dice-Kriging (see #252, #309). We can either debug and patch the current approach, or switch to a simpler implementation (the preferred option) that doesn't require the Dice-Kriging dependency.

Issue #252, opened 2018-06-10
Fail gracefully if there is not sufficient treatment overlap.


jtibshirani:
For the causal forest algorithm to work, there must be sufficient overlap among similar samples in terms of their treatment. Currently when given data with very little overlap, `causal_forest` may fail with an error along the lines of `Error in t.default(T) : argument is not a matrix`.

We should aim to detect the case where there is not sufficient overlap, and fail with an informative error message. By default before training a causal forest, we calculate treatment propensities (`W.hat` in the code), and this may be a natural point to test for low overlap.

halflearned:
Interesting. I got this error message before.

But is this really due to insufficient overlap? [This example](https://gist.github.com/halflearned/30a1e1f5b81cb95b97fb7687252e725b) has zero overlap, but causal forest still runs. (The predictions are incorrect, but the code doesn't fail).

jtibshirani:
I agree this needs further fleshing out -- I'll try to dig up the example where we first encountered the issue. Do you remember any details on when you've seen the error message before?

halflearned:
Ok, at least one source of errors can be attributed to `DiceKriging`. [Here](https://gist.github.com/halflearned/eb54423187ee3d2d9e15a0230dd4688e)'s how it gives weird errors depending on what it receives for `response` and `noise.var`.

These errors are triggered whenever we try to use tuning. [Here](https://gist.github.com/halflearned/1ed2180ce790d2bb63040b38a31f5369)'s a minimal example of this playing out when we try to tune a `regression_forest` but the outcome changes very little.

In `causal_forest`, we will see this error during the `W.hat` computations: if `W` is mostly constant and `tune.parameters=TRUE` then [this](https://github.com/swager/grf/blob/df90bfb5a4cd269e7e92831b8a1c6fb121dbd7f1/r-package/grf/R/causal_forest.R#L148) step will fail with that error.

swager:
That's surprising -- thanks! I guess this is rather concerning if it suggests that DiceKriging may just unexpectedly crash on us...

swager:
Closing this as it looks like the core problem is with DiceKriging being unstable, and we're tracking that in #278

PR #455, opened 2019-07-17
Ensuring DiceKriging does not break forest training

halflearned:
Related to [#278 - Solidify the parameter tuning implementation](https://github.com/grf-labs/grf/issues/278).

When a small number of observations is passed, DiceKriging fails with cryptic error messages. This behavior was documented in #252 and #309.

Here's a fun minimal example:

```r
set.seed(1)
X <- matrix(runif(20), 10, 2)
Y <- rep(0, 10)
good <- grf::regression_forest(X, Y, tune.parameters=F)
fail <- grf::regression_forest(X, Y, tune.parameters=T)
# Yields the error message:
# Error in chol.default(C) :
#   the leading minor of order 1 is not positive definite
```

**Major changes**
The tuning procedure is now as follows:
1. Fit small forests with random parameters (with 20 trees for regression, 100 trees for causal)
2. Predict the parameter combination that minimizes the average debiased error
3. Retrain a moderate-size forest using those parameters (with 80 trees for regression, 400 trees for causal)
4. Train another moderate-size forest using default parameters
5. Output the parameter combination that yields the smallest error
6. Train the original forest with selected parameters

If in any of steps `1-5` the code encounters an error, we throw a warning and 

**Minor changes**

As an experimental feature, the user can also check tuning status as follows:
```r
cf <- causal_forest(X, Y, W, tune.parameters=TRUE)
cf$tuning_output$status
#  'tuned'    if tuning was completed and successfully used
#  'default'  if tuning was completed but default parameters yielded smaller mse
#  'failure'  if tuning did not complete (e.g. because Y was constant)
```
The function `print.tuning_output` also shows the tuning status.

**Performance**

Code: [gist](https://gist.github.com/halflearned/e6b3129fea2b5d5e81e96c2f0c9200b3)

(I'm running additional simulations and will update this with a more precise set of bars)

At least for the DGPs here, the results are favorable. The bar colors indicate tuning or not, and the height is (infeasible) mse on oob samples. We want `tuned or default` to be lower than `notune`.

simple
aw1
aw2
amle1
amle2
ai1
ai2

---

**Possible future PRs**

This PR is a small improvement to the tuning procedure and its main objective is to avoid tuning from breaking. However, as discussed in previous meetings:

+ The attribute `tuning_output` does not include the status or the chosen parameters for the tuning that happened during the orthogonalization step.

+ There's a lot of redundancy in the code. In particular, the latter half of `causal_tuning` and `regression_tuning` are nearly identical. 

+ A cleaner API for tuning could be something like
```r
> my.tuning.options = list(parameters = c("mtry", "min.node.size"), num.fit.trees=200)
> causal_forest(X, W, Y, tuning.options = my.tuning.options)
```
this would also make it clearer which parameters are supposed to be tuned. For example, a first-time user wouldn't know that to force `sample.fraction` to be tuned, we need to explicit pass `causal_forest(X, W, Y, tune.parameters=T, sample.fraction = NULL)`. 

+ `honesty.fraction` and `prune` could also be tunable parameters

----

EDITS: 

Aug 13: updated performance plots


experiments/tuning_examples/figures.ipynb
```diff
@@ -0,0 +1,362 @@
+{
```
swager: So far we've used to "experiments" directory only for replicating experiments from published papers. For this kind of thing, I'd prefer just documenting the gains with a simple test that both illustrates the feature and helps guarantee that we don't break it in the future.

r-package/grf/R/causal_tuning.R
```diff
@@ -156,34 +156,97 @@ tune_causal_forest <- function(X, Y, W, Y.hat, W.hat,
...
-  optimal.draw <- which.min(grid[, "error"])
-  optimal.param <- grid[optimal.draw, ]
 
-  out <- list(
-    error = optimal.param[1], params = c(fixed.params, optimal.param[-1]),
-    grid = grid
+  grid <- cbind(error = c(model.surface), tuned.params)
+  small.forest.optimal.draw <- which.min(grid[, "error"])
+  small.forest.optimal.params <- grid[small.forest.optimal.draw, -1]
+  small.forest.optimal.error <- grid[small.forest.optimal.draw, 1]
```
swager: To avoid the possibility selection bias, it'd be better to re-train a small forest at the learned value, and then use that for the comparison with defaults.

experiments/tuning_examples/figures.ipynb
halflearned: Oops. Removed.

r-package/grf/R/causal_tuning.R
```diff
@@ -165,15 +165,29 @@ tune_causal_forest <- function(X, Y, W, Y.hat, W.hat,
     mean(prediction$debiased.error, na.rm = TRUE)
   })
 
+  if (all(is.na(small.forest.errors))) {
+    warning(paste0("Could not tune causal forest because all small forest error estimates were NA.\n",
+                   "Consider increasing argument num.fit.trees."))
+    out <- get_tuning_output(params = c(all.params), status = "failure")
```
swager: In case of failure, shouldn't we be resorting to defaults here?

r-package/grf/R/causal_forest.R
```diff
@@ -216,36 +216,42 @@ causal_forest <- function(X, Y, W,
...
-      stabilize.splits = stabilize.splits,
-      num.threads = num.threads,
-      honesty = honesty,
-      honesty.fraction = honesty.fraction,
-      prune.empty.leaves = prune.empty.leaves,
-      seed = seed,
-      clusters = clusters,
-      samples.per.cluster = samples.per.cluster
-    )
+    tuning.output <- tryCatch({
```
swager: The tryCatch statement should be at the most granular level possible. So for example if there's a risky call to DiceKriging, we should surround that in a tryCatch, but nothing else.

My intuition is that we should keep this part of the code roughly as it was before, and put all the error checking inside `tune_causal_forest` (so that `tune_causal_forest` is smart enough to always return an answer, and to return default parameter values if it failed to run DiceKriging).

halflearned: That makes a lot of sense. Done!

halflearned:
Thanks for the first review, @swager! All comments should be resolved now.

r-package/grf/R/print.R
```diff
@@ -94,30 +94,47 @@ print.boosted_regression_forest <- function(x, ...) {
...
-  cat("Average error by ", length(tuning.quantiles) - 1, "-quantile:\n", sep = "")
-  for (i in out) {
+  if (x$status == "failure") {
+    cat("Tuning status: failure.\n")
+    cat("This indicates tuning was attempted but failed, and we fell back to default parameters: \n\n")
+    params <- x$params
+    cat(paste0(names(params), ": ", params, "\n"))
+  } else if (x$status == "default") {
+    cat("Tuning status: default.\n")
+    cat("This indicates tuning was not attempted, and we used default parameters: \n\n")
```
erikcs: But tuning was attempted..? Will only print if tune.parameters = TRUE and tuning was attempted but params.default was best

Tuning means finding the best parameters, if the best was the default, then that's still a tuning result? Is some code bloat arising from all the different categories (in an already large tuning module) but see you must have a branch here in the current form because otherwise x$grid etc is empty...

halflearned: Ah, good point. I'll fix that immediately.

>  Is some code bloat arising from all the different categories (in an already large tuning module) but see you must have a branch here in the current form because otherwise x$grid etc is empty...

Sorry, I don't understand what you mean.

r-package/grf/R/print.R
```diff
@@ -94,30 +94,48 @@ print.boosted_regression_forest <- function(x, ...) {
...
-  cat("Average error by ", length(tuning.quantiles) - 1, "-quantile:\n", sep = "")
-  for (i in out) {
+  if (x$status == "failure") {
+    cat("Tuning status: failure.\n")
+    cat("This indicates tuning was attempted but failed due to an error, and we fell back to default parameters: \n\n")
+    params <- x$params
+    cat(paste0(names(params), ": ", params, "\n"))
+  } else if (x$status == "default") {
+    cat("Tuning status: default.\n")
+    cat("This indicates tuning was attempted.\n")
```
jtibshirani: Super small comment, don't need extra newline here?

halflearned:
:) 

How about the following compromise: add a test and fix style issues in this PRs, submit another PR this weekend with the refactoring?

swager:
Yes sounds good. And yeah, just because @jtibshirani and I are working next to each other at the same table doesn’t mean we don’t accidentally forget to coordinate ;)

halflearned:
@jtibshirani and I discussed the tests a bit further and decided that the set of simulations above (newly updated Aug 14) should give us enough confidence to merge this PR in. A more comprehensive set of tests would be nice, although it would require more thought to implement it in a useful way.

PR #456, opened 2019-07-18
Add option to disable pruning for honest trees

erikcs:
Ref #273 (Address poor performance of honest forests on small datasets.)

This PR extends the honesty parameters with a boolean flag `prune_empty_leaves` / `prune.empty.leaves` that toggles whether the honest tree is pruned or not. The parameter is placed after `honesty.fraction` in all C++ Option() signatures, and R signatures.

Logically it only entails adding an if statement in [TreeTrainer::repopulate_leaf_nodes](https://github.com/grf-labs/grf/blob/master/core/src/tree/TreeTrainer.cpp#L118), however this little innocuous change percolates through many function signatures. The following commits keeps track of all the files touched:

* 1a80452 Updates all C++ files

* 1af1fa5 Updates all R signatures

* 4beb58c Updates all R doc signatures

#### Tests:

* Test that a non-pruned honest regression forest has lower MSE than a pruned honest regression forests (65 % less for the specific setup)

* Test that pruning is executed: a non-pruned honest forest has trees with empty leaves, while the converse holds with prune=TRUE.


r-package/grf/tests/testthat/test_regression_forest.R
```diff
@@ -213,3 +213,19 @@ test_that("inverse propensity weighting in the training of a regression forest w
   ipw.mse.forest.weighted <- sum((predict(forest.weighted, X) - Y)^2)
   expect_true(ipw.mse.forest.weighted < ipw.mse.forest)
 })
+
+test_that("a non-pruned honest regression forest has lower MSE than a pruned honest regression forests
```
swager: Looks good overall. A few minor things for the tests:
- It'd be good to have explicit separation in the MSE, e.g., something like expect_true(mse.notpruned < 0.6 * mse.pruned). This gives more confidence that the feature is actually helping.
- Could you please also add a test for causal_forest, just to make sure that code path in fact gets exercised during testing?

r-package/grf/R/causal_forest.R
```diff
@@ -38,6 +38,10 @@
 #' @param honesty.fraction The fraction of data that will be used for determining splits if honesty = TRUE. Corresponds
 #'                         to set J1 in the notation of the paper. When using the defaults (honesty = TRUE and
 #'                         honesty.fraction = NULL), half of the data will be used for determining splits
+#' @param prune (experimental) If true, prune the estimation sample tree such that no leaves are empty.
+#'              If false, keep the same tree as determined in the splits sample (internally, NaNs are returned for
```
jtibshirani: Is this true? I thought that instead of returning NaN, we make sure to skip over that tree when computing predictions.

erikcs: Sorry, I think I am wording this wrong. Yes, there is a `continue` if `leaf_node.empty()` in RegressionPredictionStrategy.cpp, I meant what percolates back to R in that case, which is a NaN:

```
library(grf)
X = matrix(rnorm(100*5), 100, 5)
Y = rnorm(100)
f = regression_forest(X, y, honesty.fraction = 0.9, prune = FALSE, num.trees = 1)
p = predict(f)$predictions
sum(is.nan(p)) / 100
[1] 0.65

```

With honesty = false, and sample.fraction=0.5, the above fraction will always be 50%

jtibshirani: But I think you only get `NaN` if `num.trees = 1`? I think this wording may confuse users because it could suggest that if any empty leaf is encountered, the entire estimate becomes `NaN`. We could instead say something like "(if an empty leave is encountered, that tree is skipped and does not contribute to the estimate)".

PR #468, opened 2019-08-04
Improving some tests

halflearned:
These tests were badly calibrated and failing too often. These modifications ensure that they pass at least 95% of the time.

(In fact, they should pass >98% of the time, with the exception of [this](https://github.com/halflearned/grf/blob/c4e0b00497787f14623a91db48620e82cfb24599/r-package/grf/tests/testthat/test_causal_forest.R#L183) test that I could not improve)

### DIFF
--- a/r-package/grf/tests/testthat/test_causal_forest.R
+++ b/r-package/grf/tests/testthat/test_causal_forest.R
@@ -160,19 +160,19 @@ test_that("IPCC weighting in the training of a causal forest with missing data i
   X <- matrix(rnorm(n * p), n, p)
   e <- 1 / (1 + exp(-X[, 1] + 2 * X[, 2]))
   W <- rbinom(n, 1, e)
-  tau <- 2 * (X[, 1] > 0 & X[,5] > 0) -
+  tau <- 2 * (X[, 1] > 0 & X[, 5] > 0) -
       0.5 * (X[, 2] > 0) - 0.5 * (X[, 3] > 0) - 0.5 * (X[, 4] > 0)
   Y <- W * tau + rnorm(n)
 
-  e.cc <- 1 - 0.9 * (X[, 1] > 0 & X[,5] > 0)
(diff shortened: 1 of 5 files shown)


halflearned:
Most of these changes should be fairly uncontroversial, with the except of [this](https://github.com/grf-labs/grf/blob/c4e0b00497787f14623a91db48620e82cfb24599/r-package/grf/tests/testthat/test_causal_forest.R#L157) test, which was modified a bit more heavily than the others.

r-package/grf/tests/testthat/test_causal_forest.R
```diff
@@ -155,13 +155,13 @@ test_that("predictions are invariant to scaling of the sample weights.", {
 })
 
 test_that("IPCC weighting in the training of a causal forest with missing data improves its complete-data MSE.", {
```
swager: So here -- I'm not sure it's really possible to estimate anything with a forest n = 40 and p = 20. So I'm rather skeptical of this change. If this one is tricky, can we leave this one test as is, and follow-up with a separate PR?

r-package/grf/tests/testthat/test_merge_forests.R
```diff
@@ -43,26 +43,29 @@ test_that("Merged causal forest attributes are sensible", {
...
-  error <- mean(preds$debiased.error) + excess.error
+  excess.error <- mean(preds$excess.error, na.rm = TRUE)
+  error <- mean((Tau - preds$predictions)^2, na.rm = TRUE)
 
   big.preds <- predict(big.rf)
   big.excess.error <- mean(big.preds$excess.error)
-  big.error <- mean(big.preds$debiased.error) + big.excess.error
+  big.error <- mean((Tau - big.preds$predictions)^2, na.rm = TRUE)
 
   expect_lt(big.error, error)
```
swager: For some reason my previous comment here got dropped.

It'd be better to have a quantitative check here, e.g.,  `expect_lt(big.error, 0.95 * error)`

PR #469, opened 2019-08-04
Platform-independent random numbers [using stdlib copy]

halflearned:
Third -- and hopefully final -- attempt at  #379.

In our last meeting, we decided that we should try once again to copy from boost a way that 1) would not bloat the third party dependencies with hundreds of boost files, but also 2) be easily reproducible. Unfortunately, I wasn't able to achieve both goals simultaneously. Boost has a very complex type system, and I wasn't able to extricate the necessary functions without bringing a few hundred boost files along.

Instead, I decided to take a final look at standard library's `<random>` and noticed that most of the dependencies came from `operator<<`. Having removed those methods, it was not hard to pull the necessary classes from the standard library.

**Major changes**

+ Two new files in `third_party`: `grfstd/algorithm.hpp` and `grfstd/random.hpp`, which are nearly exact copies of the relevant parts of `algorithm` and `random` headers, with minor differences such as adding prefix `std::` to some of the function calls.

+ Calls to `std::shuffle` and `std::*_distribution` are now `grfstd::shuffle` and `grfstd::*_distribution`.

### DIFF
--- a/core/third_party/random/README.md
+++ b/core/third_party/random/README.md
@@ -9,7 +9,7 @@ They are copies of `random` and `algorithm` headers from the [llvm](https://gith
 
 Users complained about stability of random numbers across machines when setting seed across different platforms. See issue [#379](https://github.com/grf-labs/grf/issues/379).
 
-As [pointed out](https://github.com/grf-labs/grf/issues/379#issuecomment-480641123) by @jtibshrani:
+As [pointed out](https://github.com/grf-labs/grf/issues/379#issuecomment-480641123) by @jtibshirani:
 
 > the mersenne twister has the same implementation across platforms, the other random methods may differ from compiler to compiler
 
@@ -45,7 +45,7 @@ Find and remove the following `_LIBCPP` macros:
 + `_LIBCPP_HAS_NO_PRAGMA_SYSTEM_HEADER`
 + `_LIBCPP_INLINE_VISIBILITY`
 + `_LIBCPP_MSVCRT`
-+ `_LIBCPP_POP_MACROS`s
++ `_LIBCPP_POP_MACROS`
 + `_LIBCPP_PUSH_MACROS`
 + `_LIBCPP_RANDOM`
 + `_LIBCPP_TEMPLATE_VIS`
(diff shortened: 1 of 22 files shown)


halflearned:
Thanks for the review, @jtibshirani! I'm also pretty happy with how this turned out.

On `nonstd`: that's a good point, I should have checked what was done for `optional`.

On the `README`: will do.

I'm a little tied up today and tomorrow, but by Friday I should have this done.

halflearned:
Thank for the review, @jtibshirani! It should be ready for a second round whenever you have the time.

core/third_party/random/README.md
```diff
@@ -0,0 +1,73 @@
...
+## What are these files?
+
+They are copies of `random` and `algorithm` headers from the [llvm](https://github.com/llvm-mirror/libcxx/blob/master/include/) standard library.
+
+
+## Motivation
+
+Users complained about stability of random numbers across machines when setting seed across different platforms. See issue [#379](https://github.com/grf-labs/grf/issues/379).
+
+As [pointed out](https://github.com/grf-labs/grf/issues/379#issuecomment-480641123) by @jtibshrani:
```
jtibshirani: jtibshrani -> jtibshirani :)

core/third_party/random/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+> the mersenne twister has the same implementation across platforms, the other random methods may differ from compiler to compiler
+
+
+In PR [#469](https://github.com/grf-labs/grf/pull/469), @halflearned included this reduced copy of the relevant headers, ensuring random number generation is done in a consistent way across compilers.
+
+
+## How to reproduce
```
jtibshirani: This is great, thanks!

erikcs:
Making a note of that:

After #1006 core GRF produces the same result with three different compilers (latest version), using the same seed and number of threads on

- GCC
- Clang
- MSVC

Using Intel's C++ compiler ICC does not give the exact same results, but close. Evidently its compiler optimizations are more "aggressive", if they are set to -O0 or -O1 the ForestCharacterizationTest matches exactly on the linux machine I tried out ICC 19.0.0.117 on.

PR #472, opened 2019-08-04
Solidify IPCC test.

swager:
This test checks that inverse-probability of censoring weighting improves complete data MSE. Across 100 replications, the median improvement in MSE was a factor 0.75; in 97/100 tests, the improvement was better than a factor 0.9.

Addresses issue discussed in #468.

### DIFF
--- a/r-package/grf/tests/testthat/test_causal_forest.R
+++ b/r-package/grf/tests/testthat/test_causal_forest.R
@@ -155,32 +155,38 @@ test_that("predictions are invariant to scaling of the sample weights.", {
 })
 
 test_that("IPCC weighting in the training of a causal forest with missing data improves its complete-data MSE.", {
-  n <- 800
-  p <- 2
+  n <- 1000
+  p <- 6
   X <- matrix(rnorm(n * p), n, p)
-  e <- 1 / (1 + exp(-3 * X[, 1] + 2 * X[, 2]))
+  e <- 1 / (1 + exp(-X[, 1] + 2 * X[, 2]))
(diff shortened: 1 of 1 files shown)


r-package/grf/tests/testthat/test_causal_forest.R
```diff
@@ -155,32 +155,38 @@ test_that("predictions are invariant to scaling of the sample weights.", {
...
+      X[cc, ], Y[cc], W[cc],
+      sample.weights = sample.weights[cc],
+      orthog.boosting = TRUE, num.trees = num.trees
   )
-  mse <- function(f) {
-    mean((predict(f, X)$predictions - tau)^2)
-  }
-  expect_true(mse(weighted.forest) / mse(forest) < .975)
-  expect_true(mse(boosted.weighted.forest) / mse(boosted.forest) < .975)
+  expect_true(mse(boosted.weighted.forest) / mse(boosted.forest) < .9)
```
halflearned: I find it just a little more convenient to use `expect_lt` here instead of `expect_true`, because it lets the user know the difference. This is especially useful when one of the elements being compared happens to be NA (likely due to a bug), eg.
```r
> x <- NA  # a bug
> y <- 0
> expect_true(x < y)
# Error: x < y isn't true.
> expect_lt(x, y)
# Error: `x` is not strictly less than `y`. Difference: NA
```

swager: good point, thanks!

swager:
Test failure appears to be due to linter error.

PR #475, opened 2019-08-05
Add simple collection of DGPs to experiments/

erikcs:
Adds a "source"-able R script under `experiments/benchmarking/dgps.R` to easily simulate comparable data for use when tuning grf parameters or running experiments.

Each DGP is from @halflearned's [script collection](https://github.com/halflearned/grf/blob/tuning-earth/experiments/tuning_examples/tuning.R#L18) and standardized as suggested by @swager.


experiments/dgps.R
```diff
@@ -0,0 +1,125 @@
...
+  } else if (dgp == "ai2") {
+    X <- matrix(rnorm(n, p), n, p)
+    nu_x <- 0.5 * X[, 1] + 0.5 * X[, 2] + X[, 3] + X[, 4] + X[, 5] + X[, 6]
+    tau <- 0.5 * ((X[, 1] > 0) * X[, 1] + (X[, 2] > 0) * X[, 2])
+    e <- 0.5
+    W <- rbinom(n = n, size = 1, prob = e)
+    m <- nu_x + e * tau
+    V <- 0.1^2
+  } else if (dgp == "kunzel") {
+    X <- MASS::mvrnorm(n = n, mu = rep(0, p), Sigma = toeplitz(0.5^seq(0, p - 1)))
```
swager: We should quietly check that the package MASS is installed before running this and, if not, throw an error.

experiments/dgps.R
```diff
@@ -0,0 +1,125 @@
...
+#'
+#' @examples
+#' gen_data(100, 5, sigma.m = 1, sigma.tau = 0.1, sigma.noise = 1, dgp = "simple")
+#' @note:
+#' To add an additonal DGP, fill in the template below and add an entry to `dgp` and `.minp`.
+#'
+gen_data <- function(n, p, sigma.m = 1, sigma.tau = 0.1, sigma.noise = 1,
+                     dgp = c("simple", "aw1", "aw2", "aw3", "ai1", "ai2", "kunzel"),
+                     ...) {
+  .minp <- c(3, 2, 2, 1, 2, 6, 2)
```
swager: It'd be more robust to store this as a named list: c(simple=3, ...)

Then, you can access it as minp = .minp[dgp]

experiments/dgps.R
```diff
@@ -0,0 +1,125 @@
+# dgps.R - convenience script for generating simulation data for grf.
```
swager: Could you please create a new directory experiments/benchmarking, and put this in there? I don't think we want this to be at the top level of the experiments directory.

experiments/dgps.R
```diff
@@ -0,0 +1,125 @@
...
+  # Scale and return data (rescale if `m` and `tau` is not constant)
+  if (!is.na(sd(m))) {
+    m <- m / sd(m) * sigma.m
+  }
+  if (!is.na(sd(tau))) {
+    tau <- tau / sd(tau) * sigma.tau
+  }
+  V <- V / mean(V) * sigma.noise^2
+  Y <- m + (W - e) * tau + sqrt(V) * rnorm(n)
+  mean(Y)
```
erikcs: sorry about these artifacts (a force-push messed something up after doing git rebase master instead of merge)

PR #478, opened 2019-08-10
Add example of how to save a grf plot with DiagrammeRsvg

ginward:
The DiagrammeR plot function only works in RStudio's GUI environment. Add an example to `plot.grf_tree` of how to save a  plot programmatically (requires the "DiagrammeRsvg" package).


erikcs:
It seems this is limitation in DiagrammeR, adding a [workaround](https://github.com/rich-iannone/DiagrammeR/issues/70) by creating another package (DiagrammeRsvg) just adds another requirement to grf. Sounds better if the problem was fixed by DiagrammeR directly.

A current [workaround](https://github.com/ropensci/plotly/issues/294) that does not involve extra packages

```R
g = plot(get_tree(forest, 1))
htmlwidgets::saveWidget(g, 'plot.html')
```
then convert that to pdf/jpg/svg form the command line.

(saving to something else than html would ideally be fixed by Rstudio/diagrammer/htmlwidgets

jtibshirani:
I'm in agreement with @erikcs -- we try to avoid adding dependencies (even 'suggested' ones) to grf unless they are really critical.

Here's an alternative proposal: we could make the function `export_graphviz` public by removing `@keyword internal` and clean up its name. That way users can get a hold of the dot file directly, and then create a file from it using `DiagrammeRsvg` or their method of choice. We could give an example of this usage pattern in the method documentation. @ginward would this do a good enough job to address your use case?

ginward:
@jtibshirani @erikcs Thank you for your kind comments. I agree with you and I have changed the commit to exporting the export_graphviz function instead. 

Jinhua

r-package/grf/R/plot.R
```diff
@@ -67,8 +67,9 @@ size = ", num_samples, '"];')
 #' Export a tree in DOT format.
 #' This function generates a GraphViz representation of the tree,
 #' which is then written into `dot_string`.
+#' Can be used to save the plot with the following command: cat(DiagrammeRsvg::export_svg(DiagrammeR::grViz(dot_string)), file=path)
```
erikcs: Could maybe be added as a comment in `plot.grf_tree ` instead?

And why does this function have to be exported? is it not enough to do:

```R
tree.plot = plot(tree)
cat(DiagrammeRsvg::export_svg(tree.plot), file='plot.svg')
```

jtibshirani: Thanks @erikcs, I read too quickly and didn't see that you could call `export_svg` on the 'graph viz' object itself. Your suggested approach of keeping `export_graphviz` private and documenting this in `plot.grf_tree` makes sense to me.

r-package/grf/R/plot.R
```diff
@@ -80,6 +80,10 @@ export_graphviz <- function(tree) {
 }
 
 #' Plot a GRF tree object.
+#' 
+#' If you would like to save the plots, you can do so by installing the DiagrammeRsvg package and running the command:
```
erikcs: Maybe add this to the @examples section as an example of how to save the plot from the command line?

ginward:
@erikcs Thanks for your kind comments. I have added the examples section.

r-package/grf/R/plot.R
```diff
@@ -80,10 +80,15 @@ export_graphviz <- function(tree) {
 }
 
 #' Plot a GRF tree object.
+#' 
 #' @param x The tree to plot
 #' @param ... Additional arguments (currently ignored).
 #'
 #' @method plot grf_tree
+#' @examples
```
erikcs: Great, but see here how to use the examples section: https://github.com/grf-labs/grf/blob/master/r-package/grf/R/causal_forest.R#L79

i.e. please add a \dontrun and a full example that can be copy pasted and run

r-package/grf/R/plot.R
```diff
@@ -80,10 +80,15 @@ export_graphviz <- function(tree) {
 }
 
 #' Plot a GRF tree object.
+#'
```
erikcs: and you forgot to remove a white space 👀

ginward:
@erikcs Thanks for your kind suggestions. I have changed the formatting and also added the example code according to the format you provided. Please let me know if you need anything else fixed. 

Jinhua

r-package/grf/R/plot.R
```diff
@@ -88,6 +88,20 @@ export_graphviz <- function(tree) {
 #' @param ... Additional arguments (currently ignored).
 #'
 #' @method plot grf_tree
+#' @examples
```
erikcs: Great! there was a \dontrun missing so the final version could look like:

```
#' \dontrun{
#' Saving a tree plot programmatically requires the "DiagrammeRsvg" package:
#' # install.packages("DiagrammeRsvg")
#' n <- 500
#' p <- 10
#' X <- matrix(rnorm(n * p), n, p)
#' W <- rbinom(n, 1, 0.5)
#' Y <- pmax(X[, 1], 0) * W + X[, 2] + pmin(X[, 3], 0) + rnorm(n)
#' c.forest <- causal_forest(X, Y, W)
#' # Save the first tree in the forest as plot.svg
#' tree.plot = plot(get_tree(c.forest, 1))
#' cat(DiagrammeRsvg::export_svg(tree.plot), file = 'plot.svg')
#'}
```

Also, please remove the .Rapp.history committed file

Tip: if you have grf open with the grf.Rproj file you can build the documentation with CMD+Shift+D, then reload grf with CMD+Shift+L, the look at the generated documentation with ?plot.grf_tree

ginward:
@erikcs Thanks for your suggestions! I have revised further. Please have a look now and let me know if anything needs to be changed.

erikcs:
@ginward Thank you for the contribution!

tbates:
With current `DiagrammeR` version 1.0.6.1, `export_graphviz` appears not to be exported or accessible via ::: ?

Also, is "DiagrammeRsvg" a by-design add-on rather than a temporary way station for rolling into the main package? Either way, it seems that `DiagrammeR` could usefully require the SVG package as exporting seems like a common requirement.

erikcs:
Hi @tbates, [export_graphviz](https://github.com/grf-labs/grf/blob/master/r-package/grf/R/plot.R#L94) is an internal grf function if that is what you are referring to? I do not know why "DiagrammeRsvg" is an addon to "DiagrammeR", you could ask them?

PR #480, opened 2019-08-14
CI: Disable lintr

erikcs:
A bug in lintr hinders the CI tests on R 3.6. Disable lintr for now and revert this once https://github.com/jimhester/lintr/issues/384 has been resolved by the lintr developers.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,8 +7,6 @@ r_binary_packages:
   - devtools
   - testthat
   - roxygen2
-r_github_packages:
-  - jimhester/lintr
 
 matrix:
   include:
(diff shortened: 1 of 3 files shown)


russHyde:
Hi, I traced the lintr issue into xmlparsedata (and the recent changes to getParseData in R-3.6), and have pushed a change into xmlparsedata to fix this. If you'd like to use the dev version of lintr, you could try adding both `jimhester/lintr` and `r-lib/xmlparsedata` to the `r_github_packages` field of your travis.yml.

erikcs:
Thanks a lot @russHyde !

erikcs:
@jtibshirani @halflearned lintr does not break CI on R 3.6 anymore and with the current setup it never raises errors. Want it back in? 

Thought it was convenient mostly for <- instead of = which is impossible to remember

PR #481, opened 2019-08-14
Fix a typo in the `@keywords` roxygen directive.

jtibshirani:
In #342 we tried to exclude private helper methods from being included in the
package documentation. However we accidentally used `@keyword` instead of the
valid directive `@keywords`.


jtibshirani:
Thanks @halflearned for the review.

PR #487, opened 2019-08-20
Update boosting and bart implementations in simulations

rinafriedberg:
This PR updates the cross-validation for  boosting and the parameters for BART in the larger comparison simulations. It also includes example code for running Table and Figure 1 for wage regressions and adds that call to main.R.


experiments/local_linear_examples/boundary_bias_table.R
```diff
@@ -86,20 +66,12 @@ simulation.run = function(n, p, sigma, num.reps = 100, ntest = 2000){
...
-    cv.variables = boosting.cv(dtrain, Y)
-    cv.variables = as.numeric(cv.variables)
-    eta = cv.variables[1]
-    nrounds = cv.variables[2]
-    max_depth = cv.variables[3]
-    
-    model.xgb = xgboost(dtrain, nrounds = nrounds, params = list(objective = "reg:linear"),
-                        eval_metric = "rmse", eta = eta, max_depth = max_depth)
-    xgb.preds = predict(model.xgb, newdata = X.test)
+    boost.cv.fit = cvboost(as.matrix(X), Y)
```
swager: Since this is a non-standard package, it can't hurt to be explicit, and call `rlearner::cvboost`. Otherwise, it'll be impossible to figure out what's going on down the road (since googling may not help).

rinafriedberg: Good call!

PR #495, opened 2019-08-24
Added section on cross platform seeds to REFERENCE

halflearned:
A short section on how to ensure that forest predictions are the same across different platforms (see issue #379 for some discussion)

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -274,6 +274,24 @@ For further discussion of the overlap assumption, please see Imbens and Rubin (2
 
 While the algorithm in `regression_forest` is very similar to that of classic random forests, it has several notable differences, including 'honesty', group tree training for variance estimates, and restrictions during splitting to avoid imbalanced child nodes. These features can cause the predictions of the algorithm to be different, and also lead to a slower training procedure than other packages. We welcome GitHub issues that shows cases where GRF does notably worse than other packages (either in statistical or computational performance), as this will help us choose better defaults for the algorithm, or potentially point to a bug.
 
+
+### Forests predict different values depending on the platform even though the seed is the same
+
+When it comes to cross-platform predictions, the output of `grf` forest will depend on a few factors beyond the forest seed.
+
+One such factor is the compiler that was used to build `grf`. Different compilers may have optimization procedures that perform more or less aggressive floating-point optimizations, and these could lead to slightly different forest splits if the data requires numerical precision. If your data contains variables have more than 8 significant digits, please consider rounding these variables, or multiplying them by a constant.
+
+Another factor is how the forest construction is distributed across different threads. Right now, our forest splitting algorithm can give different results depending on the number of threads that were used to build the forest.
+
+We recommend that the user try to satisfy the following three conditions to ensure that results are the same across different platforms.
+
++ Set arguments `seed` and `num.threads` to the same number
++ Compile `grf` with either `gcc` or `clang` compilers (at the moment we do not provide support for this functionality for Windows-based compilers such as the MSVC compiler)
++ The data is rounded to 8 digits (if it cannot be rounded, multiply it by a constant)
+
+If you still do not see consistent cross-platform results after ensuring the three conditions above, please submit an issue.
+
+
 ## References
 
 Athey, Susan, Julie Tibshirani, and Stefan Wager. Generalized Random Forests. *Annals of Statistics (forthcoming)*, 2018.


REFERENCE.md
```diff
@@ -274,6 +274,24 @@ For further discussion of the overlap assumption, please see Imbens and Rubin (2
...
+
+Another factor is how the forest construction is distributed across different threads. Right now, our forest splitting algorithm can give different results depending on the number of threads that were used to build the forest.
+
+We recommend that the user try to satisfy the following three conditions to ensure that results are the same across different platforms.
+
++ Set arguments `seed` and `num.threads` to the same number
++ Compile `grf` with either `gcc` or `clang` compilers (at the moment we do not provide support for this functionality for Windows-based compilers such as the MSVC compiler)
++ The data is rounded to 8 digits (if it cannot be rounded, multiply it by a constant)
+
+If you still do not see consistent cross-platform results after ensuring the three conditions above, please submit an issue.
```
jtibshirani: Super small suggestion: 'issue' -> 'Github issue', since many of our users are not regulars on Github.

REFERENCE.md
```diff
@@ -274,6 +274,24 @@ For further discussion of the overlap assumption, please see Imbens and Rubin (2
...
+
+When it comes to cross-platform predictions, the output of `grf` forest will depend on a few factors beyond the forest seed.
+
+One such factor is the compiler that was used to build `grf`. Different compilers may have optimization procedures that perform more or less aggressive floating-point optimizations, and these could lead to slightly different forest splits if the data requires numerical precision. If your data contains variables have more than 8 significant digits, please consider rounding these variables, or multiplying them by a constant.
+
+Another factor is how the forest construction is distributed across different threads. Right now, our forest splitting algorithm can give different results depending on the number of threads that were used to build the forest.
+
+We recommend that the user try to satisfy the following three conditions to ensure that results are the same across different platforms.
+
++ Set arguments `seed` and `num.threads` to the same number
```
jtibshirani: The usual list formatting in this file looks like the following:

```
Here is a list.
- First item.
- Second item.
```

REFERENCE.md
```diff
@@ -274,6 +274,24 @@ For further discussion of the overlap assumption, please see Imbens and Rubin (2
 
 While the algorithm in `regression_forest` is very similar to that of classic random forests, it has several notable differences, including 'honesty', group tree training for variance estimates, and restrictions during splitting to avoid imbalanced child nodes. These features can cause the predictions of the algorithm to be different, and also lead to a slower training procedure than other packages. We welcome GitHub issues that shows cases where GRF does notably worse than other packages (either in statistical or computational performance), as this will help us choose better defaults for the algorithm, or potentially point to a bug.
 
+
+### Forests predict different values depending on the platform even though the seed is the same
+
+When it comes to cross-platform predictions, the output of `grf` forest will depend on a few factors beyond the forest seed.
```
jtibshirani: I would be nice to add a couple opening sentences to give context: "Overall, GRF is designed to produce the same estimates across platforms when using a consistent value for the random seed through the training option `seed`. However, there are still some cases where GRF can produce different estimates across platforms."

REFERENCE.md
```diff
@@ -274,6 +274,24 @@ For further discussion of the overlap assumption, please see Imbens and Rubin (2
...
+
+One such factor is the compiler that was used to build `grf`. Different compilers may have optimization procedures that perform more or less aggressive floating-point optimizations, and these could lead to slightly different forest splits if the data requires numerical precision. If your data contains variables have more than 8 significant digits, please consider rounding these variables, or multiplying them by a constant.
```
jtibshirani: I don't think I would describe these as 'optimization procedures' that perform 'aggressive floating-point optimizations', but rather 'different default behaviors around floating point rounding'.

Also I'm not sure if we should recommend multiplying variables by a constant. I think you would only want to do this for the `X` values (and not `Y`, `W`, etc.), and in this case you could still end up with slightly different estimates because of the differences in the observation variables?

REFERENCE.md
```diff
@@ -274,6 +274,24 @@ For further discussion of the overlap assumption, please see Imbens and Rubin (2
...
+
+Another factor is how the forest construction is distributed across different threads. Right now, our forest splitting algorithm can give different results depending on the number of threads that were used to build the forest.
+
+We recommend that the user try to satisfy the following three conditions to ensure that results are the same across different platforms.
+
++ Set arguments `seed` and `num.threads` to the same number
++ Compile `grf` with either `gcc` or `clang` compilers (at the moment we do not provide support for this functionality for Windows-based compilers such as the MSVC compiler)
```
jtibshirani: Is it true that we don't support consistent estimates on Windows compilers? I thought they would also be consistent, given the approach we took to random number generation.

REFERENCE.md
```diff
@@ -274,6 +274,24 @@ For further discussion of the overlap assumption, please see Imbens and Rubin (2
...
++ The data is rounded to 8 digits (if it cannot be rounded, multiply it by a constant)
```
jtibshirani: Small comment, for parallel list structure this should be 'Round the data to 8 digits...'.

REFERENCE.md
halflearned: Sure, The reason I was hesitant to include such a statement is because the user _must_ set `num.threads` to the same number as well. Otherwise, we get different results even if the other configurations are the same.

REFERENCE.md
halflearned: Thanks, will edit accordingly.

REFERENCE.md
halflearned: We haven't tested it on Windows.

But you're right in that things _should_ work. I suppose we can have users submit issues if, contrary to our current expectations, they still encounter problems.

REFERENCE.md
halflearned: :)

REFERENCE.md
jtibshirani: To me it's still accurate + helpful to say 'Overall, GRF is designed to produce the same estimates...'. That way the user knows it should be possible to get consistent estimates after making a few adjustments (in contrast with some software where there is not even an attempt at consistency).

REFERENCE.md
jtibshirani: I agree with your overall principle "if it isn't tested, it doesn't work" -- perhaps we could be really straightforward and say 'we expect this to work on Windows but have not done extensive testing, and welcome any bug reports'?

REFERENCE.md
halflearned: Ah, I see where you're coming from then. That's fair. The sentence is in now!

REFERENCE.md
```diff
@@ -274,6 +274,16 @@ For further discussion of the overlap assumption, please see Imbens and Rubin (2
...
 
+
+### Forests predict different values depending on the platform even though the seed is the same
+
+Overall, GRF is designed to produce the same estimates across platforms when using a consistent value for the random seed through the training option seed. However, there are still some cases where GRF can produce different estimates across platforms. When it comes to cross-platform predictions, the output of GRF will depend on a few factors beyond the forest seed.
+
+One such factor is the compiler that was used to build GRF. Different compilers may different default behavior around floating-point rounding, and these could lead to slightly different forest splits if the data requires numerical precision. Another factor is how the forest construction is distributed across different threads. Right now, our forest splitting algorithm can give different results depending on the number of threads that were used to build the forest.
+
+Therefore, in order to ensure that equal results across platforms, we recommend that users make sure that both arguments `seed` and `num.threads` are the same in both platforms. If results are still not consistent, please submit a Github issue.
```
jtibshirani: It still seems useful to suggest rounding the data to 8 significant digits here?

REFERENCE.md
```diff
@@ -287,6 +287,16 @@ For further discussion of the overlap assumption, please see Imbens and Rubin (2
...
 
+
+### Forests predict different values depending on the platform even though the seed is the same
+
+Overall, GRF is designed to produce the same estimates across platforms when using a consistent value for the random seed through the training option seed. However, there are still some cases where GRF can produce different estimates across platforms. When it comes to cross-platform predictions, the output of GRF will depend on a few factors beyond the forest seed.
+
+One such factor is the compiler that was used to build GRF. Different compilers may different default behavior around floating-point rounding, and these could lead to slightly different forest splits if the data requires numerical precision. Another factor is how the forest construction is distributed across different threads. Right now, our forest splitting algorithm can give different results depending on the number of threads that were used to build the forest.
```
jtibshirani: 'may different' -> 'may have different'. Also I had one comment before that might've slipped through the cracks: it still seems useful to suggest rounding the data to 8 significant digits here, as you had in an earlier version.

Issue #281, opened 2018-08-07
Find the best tree in the random forest


swager:
The idea behind a random forest is that, by averaging predictions from many trees, we can do better than any single tree alone could. However, in some instances (e.g., for visualization), it would be helpful to have a single tree that represents the signal as well as it can. Two baseline approaches to this problem would be the following:

- Just grab a single tree out of the forest (e.g., the first one). This is easy, but the resulting tree may not be particularly good: In a random forest, we on purpose make the trees a little bit noisy to explore a larger fraction of the feature space, but this means that single trees in the forest may be less accurate.
- Fit a new ad hoc tree, e.g., via CART with pruning, or mixed integer programming as recommended by Bertsimas and Dunn (2017). The former, however, is still a greedy algorithm (so may not be optimal), whereas approaches based on MIP require specialized solvers and only scale to thousands of observations.

It may be worth exploring a 3rd idea that seeks to leverage the exploration done by a forest to find a good tree. Essentially, with a random forests, we already have 2000 trees that explore various ways of representing the data. If we pick the best of those 2000, we might recover a tree that is better than just running CART once, but more practical on large datasets than solving a MIP.

The proposed approach is as follows. Assuming honesty, each tree divides the training examples into a splitting set (S), and honest estimation set (H), and an out-of-bag set (O). Everything that's specific to a single tree b = 1, ..., B is subscripted with b.

1. Train a forest, as is already done. Denote the m-th leaf of the b-th tree by L_bm. Recall that the set {L_b.} is chosen only using S_b.
2. Compute the ''raw'' loss R_bm of node L_bm, using cross-validated error on the honest estimation set H_b; see details below.
3. Compute the ''pruned'' loss P_bm of a node recursively. If L_bm is a leaf, then simply P_bm = R_bm. If L_bm is not a leaf, then P_bm also depends on the purities of the two children of L_bm, called here L_bm' and L_bm'': P_bm = min{R_bm, c + P_bm' + P_bm''}. This definition reflects a choice to prune: Either we keep the split below L_bm, in which case we pay the loss of the children; or we prune it away, and just pay the loss of the node itself. Here, c > 0 is a tuning parameter that encourages simplicity by telling the tree to avoid splits in case of doubts. 
4. Select as b* the tree with the smallest root-node impurity L_b1 (note that this is a pruned version of the b*-th tree in the original forest, as reflected by the loss criterion in step 3).
[...]

PR #502, opened 2019-08-27
Added average treatment effect (ate) of each terminal leaf node to the plots

ginward:
This feature can help identify heterogeneity, especially after #281 is merged. We can show the treatment effects within each node in the plots (ate is simply the treated outcome variable minus the control outcome variable). 

Screenshot 2019-08-27 at 19 54 25


ginward:
Thanks Stefan. I will see if I can create a C++ implementation of #281 

In another note, do you know how to calculate the standard error of the CATE estimate in the leaf node? Does it involve infinitesimal jackknife, or can I just take the standard deviation of all the CATE estimates?

ginward:
> Thanks a lot @ginward for the initiative in preparing this. However, as you note, treatment effect estimates in small leaves are very unstable, to the extent that they don't give the practitioner much information at all. For this reason, I don't want to add per-leaf CATE estimates to the plotting function -- at least yet, without also providing good companion functionality for pruning trees so they actually give reliable CATE estimates within each leaf. I was actually originally even skeptical about even including \bar{Y} and \bar{W} to the visualization, but those are arguably more harmless (since they're relatively more stable than the CATE estimates, and they're also more clearly just diagnostic tools so it's unlikely that an unsuspecting user would misunderstand and report them in an analysis).
> 
> Regarding your question about why we haven't merged #281: This was more of a proof of concept, and is currently too slow to be released as part of the main package. In general, recursion in R is super slow, so anything having to do with tree search across many trees should probably be implemented in C++. If you're interested in prototyping a fast version of #281, we'd be happy to take a look.
> 
> Also, on another note, all treatment effect estimates in GRF are doubly robust, and need to work with continuous treatments. The natural CATE estimate in a leaf would be the residual-on-residual regression estimate hat{\tau} = sum (Yi - hat{m}(Xi)) (Wi - hat{e}(Xi)) / sum (Wi - hat{e}(Xi))^2.

Thanks @swager . I will see if I can create a C++ implementation of #281 

In another note, do you know how to calculate the standard error of the CATE estimate in the leaf node? Does it involve infinitesimal jackknife, or can I just take the standard deviation of all the CATE estimates?

ginward:
I understand that this pull request is closed, but i have implemented the doubly robust ate in tree leaves here: https://github.com/ginward/grf/tree/leaf_heterogeneity

Just in case it might come in handy one day.

Issue #203, opened 2018-04-03
Standardize C++ pointer style.


jtibshirani:
Because the package was initially based on ranger, it includes a mix of styles (both classic pointers and `shared_ptr`). We should settle on a single approach that's in line with current best C++11 practices.

jtibshirani:
In #508, #512, and #513, we refactored to prefer stack objects and `std::unique_ptr` whenever possible. There are still some references to raw pointers in the `*SplittingRule` classes, but these are well-isolated and can be cleaned up as part of #231.

PR #508, opened 2019-08-30
Remove most references to shared_ptr.

jtibshirani:
This commit replaces most uses of `shared_ptr` with `unique_ptr`, invoking
`move` when necessary to transfer ownership.

We still use `shared_ptr` to hold the list of trees -- this will be addressed in
a follow-up.

Relates to #203.


jtibshirani:
Thanks @swager !

PR #512, opened 2019-08-31
Stop wrapping tree objects in shared_ptr.

jtibshirani:
Now, tree objects are stored directly instead of always being wrapped in
`std::shared_ptr`. Ownership of trees is explicitly transferred through
`std::move`.

This design may be changed or partially reverted as part of our work on #187 

Relates to #203.


jtibshirani:
Thanks @swager !

PR #523, opened 2019-09-10
Dev energy

lorismichel:


swager:
@lorismichel thanks a lot for your contribution.

Usually, when adding larger features to the repo, we start by opening an issue where we discuss the feature, and then follow up with a pull request only after we've agreed on the plan for the feature. For some examples see #272 or #288.

If you could open an issue describing the feature + implementation, that'd be very helpful. If the feature involves a new statistical or a methodological idea, it'd also be good if you could link to relevant papers, numerical experiments, etc.

The reason for this process is that, with anything that gets added to `grf`, we're implicitly both endorsing the method and committing to maintaining it in the future. So that's why we want to understand new ideas well before merging anything.

lorismichel:
Sorry this is a false error from my side.

PR #524, opened 2019-09-10
Tuning: vary sample.fraction based on ci.group.size and tune by default

erikcs:
[Not mergeable - to be split up later]

Previously: if sample.fraction is tuned (=NULL) then the grid is on U(0.05, 0.5). For Y.hat and W.hat estimation (and other regression_forest use cases where confidence intervals are not desired) ci.group.size = 1 and the sample.fraction can go beyond 0.5.

1. Suggestion: sample.fraction tuning grid is set to U(0.5, 0.8) if ci.group.size = 1

```R
n <- 100
p <- 4
Y <- abs(X[, 1]) + 0.1 * rnorm(n)

# Previous behavior:
f1 <- regression_forest(X, Y, tune.parameters = TRUE, sample.fraction = NULL, ci.group.size = 1)
mean((predict(f1)$predictions - Y)^2)
#[1] 0.06579016
f1$tuning.output$params[2]
# sample.fraction 
# 0.4536291 

# Now:
f2 <- regression_forest(X, Y, tune.parameters = TRUE, sample.fraction = NULL, ci.group.size = 1)
mean((predict(f2)$predictions - Y)^2)
# [1] 0.04222663
f2$tuning.output$params[2]
# sample.fraction
# 0.7055207
```

2. Change defaults (?) to sample.fraction = NULL so this if tuned by default if tune.parameters is set to TRUE
    Perhaps instead: increase the default sample.fraction in Y.hat and W.hat estimation in causal_forest?

    Ans: see comment [below](https://github.com/grf-labs/grf/pull/524#issuecomment-534704118): make sense to keep as is

3. Before the lowest sample.fraction was 0.05, increase this to 0.1 as from experiments going this low is never (rarely) optimal (?)

--------------
TODO:  
* Full tuning benchmark run to settle on good defaults (have been preoccupied with policyTree regret so far where this gave an improvement)
* A tuning API change first? more path dependencies like this gets hairy in the current tuning setup

Making changes to the tuning grid will be easier after #534


erikcs:
Increasing the default `sample.fraction` in Y.hat and W.hat estimation inside causal_forest does not make much difference: it can give an improvement in mse both directions, suggestion perhaps it's better to tune than change defaults.

sample fraction

A larger sample.fraction is usually beneficial for [small n](https://user-images.githubusercontent.com/7185264/65541843-e516a900-dec2-11e9-83c3-90766cd59e28.png) when propensities are "hard" to estimate as in DGPs `simple` and `aw3` (in all others e=constant, 0.01 in Kunzel). For larger n it is also an unnecessary computational increase.

swager:
Thanks, this is very interesting! One small comment (this is not a full review): For now, I'd feel more comfortable keeping `sample.fraction = 0.5` as the default, and then one can manually set it to `NULL` if one wants to tune it.

erikcs:
1. On increasing sample.fraction grid in tuning if ci.group.size = 1 (easier to add after #534):

Surprisingly this can go both ways, one would assume U(0.05, 0.8) would subsume U(0.05, 0.5) (?), but there are practical constraints like tuning being sensitive and a larger parameter space to explore (all tuned number of forests = default).

tuned sample fraction
_Simulated tuned causal forest with the [above](https://github.com/grf-labs/grf/pull/524#issuecomment-534704118) DGPs. The tuned sample.fraction is allowed to take on values up to 0.8 in W.hat/Y.hat prediction tuning vs the default of 0.5)_

erikcs:
* These two modifications were not obvious wins for tuning: leaving it as it is and close for now

erikcs:
@swager should we close https://github.com/grf-labs/grf/issues/273 with this?

The suggestion above with increasing sample.fraction could have a small impact on small data (there was already a note on this in the Algorithm Reference), but its impact on tuning performance overall was ambiguous.

PR #529, opened 2019-09-20
Rename prune.empty.leaves -> honesty.prune.leaves

erikcs:
Towards harmonizing the tuning interface: done in separate PR for manageable diff.


halflearned:
Incidentally, I wouldn't mind having all the renaming a single PR

PR #531, opened 2019-09-27
create_data_matrices: handle outcome indices

erikcs:
Suggestion:

Instead of calling and setting the offsets afterwards

```R
data <- create_data_matrices(X, Y.centered, W.centered, sample.weights = sample.weights)
outcome.index <- ncol(X) + 1
treatment.index <- ncol(X) + 2
sample.weight.index <- ncol(X) + 3
```

do

```R
data <- create_data_matrices(X, outcome = Y.centered, treatment = W.centered, sample.weights = sample.weights)
# then pass on: data$outcome_index, data$treatment_index, data$sample_weight_index,data$use_sample_weights
```

All the entry names in the create_data_matrices output are exactly the same as the signature name of the bindings: train_matrix, outcome_index, etc: allowing one further down the line to make use of this simplifying idiom:

```R
data.arguments = create_data_matrices(X, outcome = Y)
arguments = c(data.arguments, "pair list of all the other arguments")
do.call(regression_train, arguments)
```

Edit: all the variable names could stay dotted for consistency with the rest of the R side, and in do.call one would first to `gsub("\\.", "_", names(arguments))`

### DIFF
--- a/r-package/grf/R/input_utilities.R
+++ b/r-package/grf/R/input_utilities.R
@@ -227,23 +227,23 @@ create_data_matrices <- function(X, outcome = NULL, treatment = NULL,
   out <- list()
   i <- 1
   if (!is.null(outcome)) {
-    out[["outcome_index"]] <- ncol(X) + i
+    out[["outcome.index"]] <- ncol(X) + i
   }
   if (!is.null(treatment)) {
     i <- i + 1
-    out[["treatment_index"]] <- ncol(X) + i
+    out[["treatment.index"]] <- ncol(X) + i
   }
   if (!is.null(instrument)) {
     i <- i + 1
-    out[["instrument_index"]] <- ncol(X) + i
+    out[["instrument.index"]] <- ncol(X) + i
   }
   if (!isFALSE(sample.weights)) {
     i <- i + 1
-    out[["sample_weight_index"]] <- ncol(X) + i
+    out[["sample.weight.index"]] <- ncol(X) + i
     if (is.null(sample.weights)) {
-      out[["use_sample_weights"]] <- FALSE
+      out[["use.sample.weights"]] <- FALSE
     } else {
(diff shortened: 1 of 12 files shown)


r-package/grf/R/input_utilities.R
```diff
@@ -220,18 +220,45 @@ validate_sample_weights <- function(sample.weights, X) {
...
+  }
+  if (!is.null(treatment)) {
+    i <- i + 1
+    out[["treatment_index"]] <- ncol(X) + i
+  }
+  if (!is.null(instrument)) {
+    i <- i + 1
+    out[["instrument_index"]] <- ncol(X) + i
+  }
+  if (!isFALSE(sample.weights)) {
```
halflearned: Why is `sample.weights=FALSE` by default, as opposed to `sample.weights=NULL`?
 
What's the desired behavior for `sample.weights=TRUE`? (Right now, it appends a column of `1`s)

erikcs: it is convoluted (a better way?) because of how sample.weights is handled

FALSE is to distinguish 1 vs 2:

1. passing on to regression_train and causal_train:

even when sample.weight = NULL: sample_weight_index is required in the bindings
(together with use_sample_weights)

1.1: If you are calling these: call create_data_matrices with NULL if no sample.weights, else with the sample.weights specified 

2. passing on to for ex. regression_predict: does not take sample.weight, and does
not require sample_weight_index as above: specify this with sample.weights = FALSE

erikcs: Maybe a simplification is to remove [bool use_sample_weights](https://github.com/grf-labs/grf/blob/master/r-package/grf/bindings/RegressionForestBindings.cpp#L36)

and instead inside do bool use_sample_weights = sample_weight_index > 0

and in create_data_matrices: set sample_weight_index = 0 if NULL

PR #534, opened 2019-10-01
Simplify function signatures and tuning interface

erikcs:
1. Function signatures are reordered for readability and defaults declared inline. The way tuning is specified is simplified by passing a string or vector of parameter names to tune. 

    Set `tune.parameters = "all"` to tune all of: `("sample.fraction", "mtry", "min.node.size", "honesty.fraction", "honesty.prune.leaves", "alpha", "imbalance.penalty")`, or pass the names of the parameters to tune. The default is "none" (no tuning).

    - [regression_forest.R](https://github.com/erikcs/grf/blob/tuning-interface-1/r-package/grf/R/regression_forest.R)

    - [causal_forest.R](https://github.com/erikcs/grf/blob/tuning-interface-1/r-package/grf/R/causal_forest.R)

    - [ll_regression_forest](https://github.com/erikcs/grf/blob/tuning-interface-1/r-package/grf/R/local_linear_forest.R): only modified thing was updating signature (and simplify the call)

2. The tuning interface is simplified: 

    - core tuning happens in [tuning.R](https://github.com/erikcs/grf/blob/tuning-interface-1/r-package/grf/R/tuning.R) (internal use)

    - [tune_regression_forest](https://github.com/erikcs/grf/blob/tuning-interface-1/r-package/grf/R/tune_regression_forest.R) and [tune_causal_forest](https://github.com/erikcs/grf/blob/tuning-interface-1/r-package/grf/R/tune_causal_forest.R) are wrappers of the above function.

    - two extra tests added [1](https://github.com/erikcs/grf/blob/tuning-interface-1/r-package/grf/tests/testthat/test_causal_forest_tuning.R#L21), [2](https://github.com/erikcs/grf/blob/tuning-interface-1/r-package/grf/tests/testthat/test_regression_forest_tuning.R#L21)

    - [Timings and output before/after](https://gist.github.com/erikcs/008d8bc1ce435bd533e68299a3b9a412): Since tuning is already time consuming: this change does not drastically reduce performance

3. Peripheral changes are made to accommodate the above additions.

No new functionality is added. Some small fixes to existing code: a) tuning: num.trees dangling variable, divide by zero if mean(small.forest.errors) = 0) b) regression_tests: name sample.weights arguments c) boosting did not pass on tuning fit parameters.


r-package/grf/R/boosted_regression_forest.R
```diff
@@ -89,29 +89,29 @@
...
+                                      tune.num.trees = 10,
+                                      tune.num.reps = 100,
+                                      tune.num.draws = 1000,
                                       boost.steps = NULL,
                                       boost.error.reduction = 0.97,
                                       boost.max.steps = 5,
-                                      boost.trees.tune = 10) {
+                                      boost.trees.tune = 10,
+                                      num.threads = 0,
+                                      seed = runif(1, 0, .Machine$integer.max)) {
```
halflearned: Feels funny to pass a double as a seed. How about:
```r
seed = sample(.Machine$integer.max, 1)
```

r-package/grf/R/causal_forest.R
```diff
@@ -136,59 +139,61 @@
...
 
   if (is.null(Y.hat) && !orthog.boosting) {
-    forest.Y <- regression_forest(X, Y,
-      sample.weights = sample.weights, sample.fraction = sample.fraction, mtry = mtry,
-      tune.parameters = tune.parameters, num.trees = num.trees.orthog, num.threads = num.threads, min.node.size = NULL,
-      honesty = TRUE, honesty.fraction = NULL, honesty.prune.leaves = honesty.prune.leaves, seed = seed, ci.group.size = 1,
-      alpha = alpha, imbalance.penalty = imbalance.penalty, clusters = clusters,
-      samples.per.cluster = samples.per.cluster
-    )
+    forest.Y <- do.call(regression_forest, c(Y = list(Y), args.orthog))
```
halflearned: ❤️

r-package/grf/R/causal_forest.R
```diff
@@ -197,88 +202,56 @@ causal_forest <- function(X, Y, W,
...
-    seed
-  )
+  if (tune.parameters[1L] != "none") {
+    if (tune.parameters[1L] == "all") {
+      tune.parameters <- tunable.params
+      if (!honesty) {
+        tune.parameters <- tune.parameters[!grepl("honesty", tune.parameters)]
+      }
+    } else {
+      tune.parameters <- unique(match.arg(tune.parameters, tunable.params, several.ok = TRUE))
```
halflearned: At first glance, I was momentarily confused by the meaning of `tunable.params` vs `tune.parameters`.

I'm not sure what's the best option here. How about at least changing `tunable.params` to `all.tunable.params`?

r-package/grf/R/input_utilities.R
```diff
@@ -276,3 +276,8 @@ observation_weights <- function(forest) {
   }
   observation.weight
 }
+
+do.call.grf = function(what, args, quote = FALSE, envir = parent.frame()) {
```
halflearned: For future generations, could you put a short comment here on why this function is necessary?

r-package/grf/R/tuning.R
```diff
@@ -0,0 +1,174 @@
...
+    fit <- regression_train
+    predict_oob <- regression_predict_oob
+    predict.data.args <- c("train.matrix", "sparse.train.matrix", "outcome.index")
+    defaults <- formals(regression_forest)
+  } else if (forest.type == "causal_forest") {
+    fit <- causal_train
+    predict_oob <- causal_predict_oob
+    predict.data.args <- c("train.matrix", "sparse.train.matrix", "outcome.index", "treatment.index")
+    defaults <- formals(causal_forest)
+  }
```
halflearned: What happens here if I forget to input a `forest.type`?

r-package/grf/R/tuning.R
```diff
@@ -0,0 +1,174 @@
...
+    small.forest <- do.call.grf(fit, c(data, fit.parameters, draw.parameters))
+    prediction <- do.call.grf(predict_oob, c(list(forest_object = small.forest), predict_oob.args))
+    error <- prediction$debiased.error
+
+    mean(error, na.rm = TRUE)
+  })
+
+  if (any(is.na(small.forest.errors))) {
+    warning(paste0(
+      "Could not tune causal forest because some small forest error estimates were NA.\n",
```
halflearned: This comment should match the passed `forest.type`

r-package/grf/tests/testthat/test_causal_forest_tuning.R
```diff
@@ -7,37 +7,33 @@ test_that("causal forest tuning decreases prediction error", {
...
   tuned.preds <- predict(tuned.forest)
   tuned.error <- mean((tuned.preds$predictions - TAU)^2)
 
   expect_true(tuned.error < error * 0.75)
 })
 
-test_that("causal forest tuning only cross-validates null parameters", {
-  p <- 6
+test_that("causal forest forest tuning tunes each parameter", {
+  p <- 4
```
halflearned: Just checking: are these test modifications on purpose?

r-package/grf/R/regression_forest.R
```diff
@@ -84,100 +87,83 @@
...
   class(forest) <- c("regression_forest", "grf")
   forest[["ci.group.size"]] <- ci.group.size
   forest[["X.orig"]] <- X
   forest[["Y.orig"]] <- Y
   forest[["sample.weights"]] <- sample.weights
   forest[["clusters"]] <- clusters
-  forest[["tunable.params"]] <- tunable.params
-  if (tune.parameters) {
+  forest[["tunable.params"]] <- args[tunable.params]
+  if (exists("tuning.output")) {
```
halflearned: Minor comment: this looks a bit inelegant. How idiomatic is this? Consider just defining `tuning.output` as `NULL` whenever `tune.parameters[1L] == "none"`.

r-package/grf/R/tuning.R
erikcs:
```
> match.arg(c("regression_forest", "causal_forest"), c("regression_forest", "causal_forest"))
[1] "regression_forest"
```

r-package/grf/R/causal_forest.R
erikcs: (っ◔◡◔)っ `do.call`

r-package/grf/tests/testthat/test_causal_forest_tuning.R
erikcs: Ah, yes this is basically just replacing the above test (only testing that a parameter is tuned), and I just set p=4 because it was faster

r-package/grf/R/boosted_regression_forest.R
erikcs: `runif(1, 0, .Machine$integer.max)` is always an integer, but surprisingly replacing it with `sample(.Machine$integer.max, 1)` caused some tests to fail (thresholds changed)

r-package/grf/R/input_utilities.R
```diff
@@ -131,7 +131,7 @@ validate_samples_per_cluster <- function(samples.per.cluster, clusters) {
   }
   cluster_size_counts <- table(clusters)
   min_size <- unname(cluster_size_counts[order(cluster_size_counts)][1])
-  if (is.null(samples.per.cluster)) {
+  if (samples.per.cluster == 0) {
```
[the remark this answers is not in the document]
erikcs: > * Could you explain the TODO "one design wart in tuning (X): lazy eval of function arguments and default for mtry"?

X is passed to tuning [here](https://github.com/erikcs/grf/blob/tuning-interface-1/r-package/grf/R/tuning.R#L17), even though the X needed for fitting already lives in the `data` also passed.

The tiny thing X is needed for is:

1. Clear: The nrow and ncol of X is [used](https://github.com/erikcs/grf/blob/tuning-interface-1/r-package/grf/R/tuning.R#L46) in [get_params_from_draw](https://github.com/erikcs/grf/blob/tuning-interface-1/r-package/grf/R/tuning.R#L55)

2. "Vague": requirement from [defaults](https://github.com/erikcs/grf/blob/tuning-interface-1/r-package/grf/R/tuning.R#L31) which includes this un-evaluated symbol for `mtry`: `min(ceiling(sqrt(ncol(X)) + 20), ncol(X))`

If that `mtry` (inside `args`) is to be passed to `do.call`, `X` has to live in the same environment as the expression is being evaluated in...

erikcs:
> * Could you explain the TODO "one design wart in tuning (X): lazy eval of function arguments and default for mtry"?

X is passed to tuning [here](https://github.com/erikcs/grf/blob/tuning-interface-1/r-package/grf/R/tuning.R#L17), even though the X needed for fitting already lives in the `data` also passed.

The tiny thing X is needed for is:

1. Clear: The nrow and ncol of X is [used](https://github.com/erikcs/grf/blob/tuning-interface-1/r-package/grf/R/tuning.R#L46) in [get_params_from_draw](https://github.com/erikcs/grf/blob/tuning-interface-1/r-package/grf/R/tuning.R#L55)

2. "Vague": requirement from [defaults](https://github.com/erikcs/grf/blob/tuning-interface-1/r-package/grf/R/tuning.R#L31) which includes this un-evaluated symbol for `mtry`: `min(ceiling(sqrt(ncol(X)) + 20), ncol(X))`

If that `mtry` (inside `args`) is to be passed to `do.call`, `X` has to live in the same environment as the expression is being evaluated in...

r-package/grf/R/input_utilities.R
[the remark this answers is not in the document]
erikcs: From the meeting on the proposed signature change: 

https://github.com/erikcs/grf/blob/tuning-interface-1/r-package/grf/R/regression_forest.R#L91

You mentioned it was inconsistent how now every parameters have defaults inlined in the signature, except a few which are now NULL.

This (and the 3 comments below) was an attempt (agree not very elegant) to remove the nulls and replace them with the default value that was passed to C++ (for sample.weight that did not work...)

Revert all these 3 back to NULL?

r-package/grf/R/tuning.R
```diff
@@ -0,0 +1,174 @@
...
+                        data,
+                        args,
+                        tune.parameters,
+                        forest.type = c("regression_forest", "causal_forest"),
+                        num.fit.trees = 10,
+                        num.fit.reps = 100,
+                        num.optimize.reps = 1000) {
+  forest.type <- match.arg(forest.type)
+
+  if (forest.type == "regression_forest") {
```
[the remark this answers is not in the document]
erikcs: Do you mean something like this?

```R
tune_regression_forest = function(X,
                                  data,
                                  args,
                                  tune.parameters,
                                  num.fit.trees,
                                  num.fit.reps,
                                  num.fit.draws) {
  fit <- regression_train
  predict_oob <- regression_predict_oob
  predict.data.args <- c("train.matrix", "sparse.train.matrix", "outcome.index")
  defaults <- formals(regression_forest)
  
  tuning.output <- tune_forest(X, data, args, tune.parameters, num.fit.trees, num.fit.reps, num.fit.draws,
                               fit, predict_oob, predict.data.args, defaults)
 
  tuning.output 
}
```

r-package/grf/R/tuning.R
```diff
@@ -0,0 +1,174 @@
...
+    fit <- regression_train
+    predict_oob <- regression_predict_oob
```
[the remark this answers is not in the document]
erikcs: It was just_with_underscores since all grf functions were with underscores! (and do.call.rcpp is "base R")

r-package/grf/R/tuning.R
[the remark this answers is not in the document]
jtibshirani: Yes, that's what I was thinking. It makes the signature of `tune_forest` less clean, but to me it's a good trade-off.

r-package/grf/R/tuning.R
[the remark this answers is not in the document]
jtibshirani: Ah, makes sense. I was thinking of this as a variable and thought it should have dots... but R doesn't distinguish between function declarations and function references.

This made me notice another super small thing, maybe `fit` should be renamed to `train`?

r-package/grf/NAMESPACE
```diff
@@ -27,8 +27,11 @@ export(quantile_forest)
 export(regression_forest)
 export(split_frequencies)
 export(test_calibration)
+export(tune_causal_forest)
+export(tune_forest)
```
jtibshirani: I don't think `tune_forest` should be exported, it seems like part of the internal implementation.

r-package/grf/tests/benchmarks/#causal_benchmark.R#
```diff
@@ -69,7 +69,7 @@ make_causal_forest = function(stabilize.splits, min.node.size,
 
 res.untuned.unstab = sapply(1:5, function(setup) {
   evaluate_method(function(X, Y, W) {
-    cf = causal_forest(X, Y, W, tune.parameters = FALSE, stabilize.splits = FALSE)
```
jtibshirani: No action on your part, but I noticed this file should've never been committed... I guess the `benchmarks` directory could use a clean-up.

erikcs: Yes, there is also a new benchmark dgp collection here: https://github.com/grf-labs/grf/tree/master/experiments/benchmarking

PR #535, opened 2019-10-09
Include <stdexcept> (gcc 10 compliance)

erikcs:
Addressing the upcoming gcc 10 which currently [fails](https://www.stats.ox.ac.uk/pub/bdr/gcc10/grf.log)

```
src/forest/ForestOptions.cpp:50:16: error: 'runtime_error' is not a member of 'std'
   50 |     throw std::runtime_error("When confidence intervals are enabled, the"
      |                ^~~~~~~~~~~~~
src/forest/ForestOptions.cpp: In static member function 'static grf::uint grf::ForestOptions::validate_num_threads(grf::uint)':
src/forest/ForestOptions.cpp:96:16: error: 'runtime_error' is not a member of 'std'
```

Solution: include `<stdexcept>` https://en.cppreference.com/w/cpp/error/runtime_error

### DIFF
--- a/core/src/commons/Data.cpp
+++ b/core/src/commons/Data.cpp
@@ -18,6 +18,7 @@
 #include <algorithm>
 #include <fstream>
 #include <iterator>
+#include <stdexcept>
 #include <sstream>
 
 #include "Data.h"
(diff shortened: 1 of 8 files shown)


erikcs:
Yes

Issue #109, opened 2017-07-23
Add support for factor variables.


jtibshirani:
We should likely implement the approach suggested in ESL where in each node, factor variables are ordered by their mean outcome before performing the split. This should be properly generalized to handle non-regression forests.

lwu9:
When there is a categorical variable X1 in X, it is possible that there is a child node where those observations contain only one category, i.e. all values of the variable X1 are the same. In this case, how can we get pseudo-outcomes? The pseudo-outcomes is obtained through the inverse of Ap which may be singular. Isn't it?

swager:
The X-values aren't used to compute the pseudo-outcomes in the leaf in the standard GRF formulation; rather, only the "outcomes" matter (e.g., W and Y for `causal_forest`). The features X enter into the problem by determining which leaf an observation falls into.

lwu9:
Thanks for you reply @swager! I understand in your causal_forest, pseudo-outcomes only have W and Y. But if we want to do the local linear regression, so our psi should be:
psi(Y_i)=Y_i-theta * X_i, (here theta is related to the query point x), shouldn't it? So when we calculate Ap, we need take derivative of psi w.r.t. theta, and the result will include feature X. Is there anything wrong in my understanding?

jtibshirani:
We added the [sufrep package](https://github.com/grf-labs/sufrep), which contains a collection of methods for handling categorical variables, and a [tutorial](https://grf-labs.github.io/grf/articles/categorical_inputs.html) for how to use sufrep with grf.

lwu9:
Thanks

PR #537, opened 2019-10-10
Documentation and vignette for categorical variables 

halflearned:
This addresses #109 (Add support for factor variables.)

Note that GRF doesn't have support factor variables per se, but we would like to teach users how to encode their categorical variables in different ways before calling `grf`. To that effect, this PR adds:

+ A section on categorical inputs to REFERENCE.md
+ A vignette with a short example, including [`sufrep`](https://github.com/grf-labs/sufrep) usage

Other changes:
+ ~Added `sufrep` to Suggests~
+ ~Travis install `sufrep`~

CRAN does not allow for github packages. Removed `sufrep` from DESCRIPTION and Travis. In the vignette, the package is just installed on the fly. I'll revert this once `sufrep` is on CRAN.


r-package/grf/vignettes/categorical_inputs.Rmd
```diff
@@ -0,0 +1,78 @@
+---
+title: "Categorical inputs"
+output: rmarkdown::html_vignette
+vignette: >
+  %\VignetteIndexEntry{ci_and_num.trees}
```
erikcs: New name!

And remember you have to add an entry here:

https://github.com/grf-labs/grf/blob/master/r-package/grf/_pkgdown.yml#L10

halflearned: Thanks!

r-package/grf/_pkgdown.yml
```diff
@@ -10,6 +10,8 @@ navbar:
     menu:
       - text: "Confidence intervals and the number of trees"
         href: articles/ci_and_num.trees.html
+      - text: "Categorical inputs"
```
erikcs: Keep it alphabetical and put this on top?

r-package/grf/DESCRIPTION
```diff
@@ -32,5 +32,8 @@ RoxygenNote: 6.1.1
 Suggests:
     DiagrammeR,
     testthat,
+    sufrep
+Remotes:
```
erikcs: R CMD check complains, and in this thread https://github.com/r-lib/devtools/issues/1717#issuecomment-368125654 it's mentioned that Remotes should not be there

r-package/grf/vignettes/categorical_inputs.Rmd
```diff
@@ -0,0 +1,78 @@
...
+x <- c("cyl", "qsec") # Continuous variables
+g <- c("maker")       # Categorical variable
+
+head(df[c(x, g)])
+```
+
+This code would raise an error, because data is not numerical.
+
+```{r problem}
+# rf <- regression_forest(X, Y)
```
erikcs: what is X and what is Y? It is evident from the context, just checking if you forgot to rename them

halflearned: Good point, thanks.

r-package/grf/vignettes/categorical_inputs.Rmd
```diff
@@ -0,0 +1,78 @@
...
+knitr::opts_chunk$set(
+  collapse = TRUE,
+  comment = "#>"
+)
+set.seed(123)
+```
+
+```{r setup}
+library(grf)
+library(sufrep)
```
erikcs: add a comment linking to https://github.com/grf-labs/sufrep?

r-package/grf/vignettes/categorical_inputs.Rmd
```diff
@@ -0,0 +1,92 @@
...
+x <- c("cyl", "qsec") # Continuous variables
+g <- c("brand")       # Categorical variable
+
+head(df[c(x, g)])
+```
+
+This code would raise an error, because data is not numerical.
+
+```{r problem}
+# rf <- regression_forest(X=df[c(x, g)], Y=df$mpg)
```
erikcs: nothing wrong here but just noticed `validate_X` being inconsistent: doing as.matrix(df[c(x, g)]) causes it to provide the error message instead of the call to the bindings

erikcs: And maybe we should not expose that inconsistency here? (i.e fix it in another pr)

erikcs: Maybe this error message could even be updated to refer to sufrep: https://github.com/grf-labs/grf/blob/master/r-package/grf/R/input_utilities.R#L8

halflearned: Yeah, I noticed it as well. I'll leave the error message as is for now, will fix it later.

PR #543, opened 2019-10-17
merge_forest: check that all forests are trained on the same data

### LINKED ISSUE #542: merge_forests segfaults if forests trained on samples of different size
**Issue**
If the user trains forest in datasets of different size, then `merge_forests` hits segfault when trying to retrieve oob predictions. 

**Example**
This causes a segfault
```r
library(grf)

X <- matrix(runif(400), 200, 2)
Y <- pmax(X[,1] > 0.5) + rnorm(200)

X1 <- X[1:80,]
X2 <- X[81:200,]
Y1 <- Y[1:80]
Y2 <- Y[81:200]

rf1 <- regression_forest(X1, Y1)
rf2 <- regression_forest(X2, Y2)
big_rf <- merge_forests(list(rf1, rf2), compute.oob.predictions = T) # Segfault 
```
If we don't compute oob predictions during merge, we can still hit a segfault later.

```r
big_rf <- merge_forests(list(rf1, rf2), compute.oob.predictions = F) 
yhat <- predict(big_rf) # Segfault
```

This, however, is fine.
```r
X <- matrix(runif(300), 150, 2)
big_rf <- merge_forests(list(rf1, rf2), compute.oob.predictions = F) 
yhat <- predict(big_rf, X_test)  # predicting on new dataset.
```

**Comments**
We might not be consider this a bug necessarily, since it involves using `merge_forests` in an unexpected way. The documentation states that `merge_forests`...

>  ...Merges a list of forests that were grown **using the same** data into one large forest

However, we might still want some users to be able to compute predictions on a test set as above. And in either case we should catch that error before it hits a segfault.

**GRF version**
0.10.4

erikcs:
* add check to validate_forest_list that all input forests are trained on the same data
* add test to test_merge_forests that this is caught

Fixes #542


swager:
Looks good! Just one thing -- as with any bugfix commit, could you please add a test that used to fail, but now passes. That ensures we never let the bug creep back into the codebase.

jtibshirani:
@erikcs I actually think we should error in this case instead of guessing which of the original forests to use. If a user tries to merge forests that were trained on different datasets, it's likely a mistake and they would want to be alerted to it.

Issue #488, opened 2019-08-20
Options for cluster weighting in cluster-robust forests


swager:
In the current implementation of causal forests, etc., when a user specifies clusters for samples, each cluster is given equal weight. For example, if we have data from a randomized trial clustered by village, and one village has 100 people and the other has 400 people, then both villages get equal weight in the forest. For details, see <a href="https://github.com/grf-labs/grf/blob/master/REFERENCE.md#cluster-robust-estimation">here</a>, and for an example in code, see <a href="https://github.com/grf-labs/grf/blob/584d85613ecded99f07275536b78d90f2f0c7d6f/r-package/grf/R/average_treatment_effect.R#L324">here</a>, and also #280 #393 #431.

In some applications, however, this choice may be inappropriate -- and it may be preferable not to re-weights clusters, and instead to give each cluster a weight proportionate to the number of samples it contains. We should thus add an option to turn off the cluster weighting if desired. Furthermore, we should add a transparent way for power users to manually specify the weight given to each cluster (potentially via the sample weights argument).

erikcs:
In the analysis functions computing all the different ATE/LATE/CATE etc:

swager:
Ok so this is simply a little tricky in terms of notation (one could probably write this more nicely with slightly less condensed code). The point is to estimate the variance of
```R
aclate.hat <- weighted.mean(Gamma.hat, subset.weights)
```
where data across clusters are independent. Write samples as i and clusters as j, the "Gamma.hat" as Gij and weights as Sij. So aclate.hat (theta) is
```
theta = sum_{cluser j} Aj / sum_{cluser j} Tj.
```
where Aj = sum_{sample i} Gij Sij and Tj = sum_{cluser j} sum_{sample i} Sij; notice that we grouped samples by cluster j because samples within cluster may be arbitrarily dependent. Assuming the clusters are IID, the variance of theta then scales like
```
(number of clusters) * Var[theta] -> Var[Aj] / E[Tj]^2.
```
In the code, we compute this as
```R
    clust.avg <- Matrix::sparse.model.matrix(
      ~ factor(subset.clusters) + 0,
      transpose = TRUE
    ) %*% (Gamma.hat * subset.weights) 
    sigma2.hat <- var(as.numeric(clust.avg)) * length(clust.avg) /
      sum(subset.weights)^2
```
[...]

PR #545, opened 2019-10-19
Change the cluster weighting for cluster robust forests

erikcs:
This addresses #488:

The user commits at tree training time how samples are drawn from clusters (and implicitly how the estimated tau is weighted depending on differences in cluster sizes) and this propagates to how ATEs/etc are computed in the analysis functions. 

`equalize.cluster.weights = FALSE` (new default) will use all the observations in each cluster. It is the same as setting samples.per.cluster to the maximum number of observations in the clusters. With this option the user is free to provide `sample.weights`.

`equalize.cluster.weights = TRUE` will use the number of observations that are in the smallest cluster. It is the same as setting samples.per.cluster to the minimum number of observations in the clusters. With this option the user can not provide `sample.weights`.

Summary:

| sample.weights | equalize.cluster.weights = FALSE                                                 | equalize.cluster.weights = TRUE                                                                                       |
|----------------|----------------------------------------------------------------------------------|-----------------------------------------------------------------------------------------------------------------------|
|                |                                                                                  |                                                                                                                       |
| No             | Each unit is given the same weight, regardless of cluster size.  **New default** | Each cluster is given equal weight.  If the smallest cluster has K units, each tree is grown with a random K elements |
| Yes            | Additional custom weighting option                                               | X Disallowed X                                                                                                        |

---------------

* breaking change: `samples.per.cluster` is replaced with boolean flag `equalize.cluster.weights`. This is FALSE by default, which is the opposite of the previous behavior.
* observation_weights() is updated
* validate_equalize_cluster_weights() replaces validate_samples_per_cluster()
* Misc modifications to some test tolerances
* average_partial_effect.R fix by @swager


REFERENCE.md
```diff
@@ -249,13 +249,14 @@ Some additional notes about the behavior of boosted regression forests:
...
 
-When clustering is enabled during training, all subsampling procedures operate on entire clusters as opposed to individual examples. Then, to determine the examples used for performing splitting and populating the leaves, `samples_per_cluster` examples are drawn from the selected clusters. By default, `sample_per_cluster` is equal to the size of the smallest cluster. Concretely, the cluster-robust training procedure proceeds as follows:
+When clustering is enabled during training, all subsampling procedures operate on entire clusters as opposed to individual examples. Then, to determine the examples used for performing splitting and populating the leaves, *Nc* examples are drawn from the selected clusters. By default, *Nc* is all the observations in the clusters. If `clusters.subsample = TRUE` *Nc* is equal to the size of the smallest cluster.
```
swager: What is the *Nc* notation style? Is this aligned with well-established R conventions? If not, I'd prefer just writing out `samples_per_cluster` everywhere to avoid the possibility of confusion.

r-package/grf/R/input_utilities.R
```diff
@@ -73,17 +73,22 @@ validate_clusters <- function(clusters, X) {
...
 
-validate_samples_per_cluster <- function(samples.per.cluster, clusters) {
+validate_clusters_subsample <- function(clusters.subsample, clusters) {
   if (is.null(clusters) || length(clusters) == 0) {
     return(0)
   }
   cluster_size_counts <- table(clusters)
-  min_size <- unname(cluster_size_counts[order(cluster_size_counts)][1])
-  if (is.null(samples.per.cluster)) {
+  cluster_sizes <- unname(cluster_size_counts[order(cluster_size_counts)])
```
swager: Why not just do min_size = min(cluster_size_counts), etc.?

erikcs:
@swager the plan https://github.com/grf-labs/grf/issues/488#issuecomment-543883355 was to update `observation_weights` in a second PR since it will touch many files (tests will likely need updating as well) for PR readability, or I could do it here if you are fine with that?

r-package/grf/R/input_utilities.R
```diff
@@ -197,12 +199,19 @@ observation_weights <- function(forest) {
...
+      inverse.counts <- 1 / as.numeric(Matrix::colSums(Matrix::sparse.model.matrix(~ clust.factor + 0)))
+      observation.weight <- sample.weights * inverse.counts[as.numeric(clust.factor)]
+    }
   } else {
-    clust.factor <- factor(forest$clusters)
-    inverse.counts <- 1 / as.numeric(Matrix::colSums(Matrix::sparse.model.matrix(~ clust.factor + 0)))
-    observation.weight <- sample.weights * inverse.counts[as.numeric(clust.factor)]
+    observation.weight <- sample.weights
   }
+
```
swager: For clarity, can we write this function something like this? Like that, we clearly disambiguate between all possible cases of clustering vs sample weighting

if no sample weights provided
&nbsp;  if no clusters provided or we don't do unit per-cluster  weighting
&nbsp;&nbsp;     raw weights are all ones
 &nbsp; if we do unit per-cluster weighting
 &nbsp;&nbsp;    raw weights is what's on 207-209 above
if no sample weights provided
&nbsp;  if no clusters provided or we don't do unit per-cluster  weighting
  &nbsp;&nbsp;   raw weights are sample weights
&nbsp;  if we do unit per-cluster weighting
 &nbsp;&nbsp;    ERROR

return raw weights / sum raw weights

r-package/grf/R/instrumental_forest.R
```diff
@@ -25,14 +25,8 @@
...
-#' @param samples.per.cluster If sampling by cluster, the number of observations to be sampled from
-#'                            each cluster when training a tree. If NULL, we set samples.per.cluster to the size
-#'                            of the smallest cluster. If some clusters are smaller than samples.per.cluster,
-#'                            the whole cluster is used every time the cluster is drawn. Note that
-#'                            clusters with less than samples.per.cluster observations get relatively
-#'                            smaller weight than others in training the forest, i.e., the contribution
-#'                            of a given cluster to the final forest scales with the minimum of
-#'                            the number of observations in the cluster and samples.per.cluster. Default is NULL.
+#' @param clusters.subsample Whether to subsample from the clusters according to the minimum cluster size (TRUE) or
+#'   sample the full clusters (FALSE). Default is FALSE.
```
swager: This comment may not be super clear to users. Can we do something like this instead?

@param equalize.cluster.weights   If FALSE, each unit is given the same weight (so that bigger clusters get more weight). If TRUE, each cluster is given equal weight in the forest. In this case, during training, each tree uses the same number of observations from each drawn cluster: If the smallest cluster has K units, then when we sample a cluster during training, we only give a random K elements of the cluster to the tree-growing procedure. When estimating average treatment effects, each observation is given weight 1/cluster size, so that the total weight of each cluster is the same. Default is FALSE. Note that, if this argument is FALSE, sample weights may also be directly adjusted via the sample.weights argument. If this argument is TRUE, sample.weights must be set to null.

swager: Can you confirm that we error is equalize.cluster.weights = TRUE and sample.weights is not null?

r-package/grf/R/input_utilities.R
erikcs: Do you mean return just `raw weights`? If `raw weights / sum raw weights` some existing tests with no clustering or sample weights will fail?

r-package/grf/R/input_utilities.R
```diff
@@ -191,19 +196,30 @@ create_data_matrices <- function(X, outcome = NULL, treatment = NULL,
...
-  } else {
-    forest$sample.weights * length(forest$Y.orig) / sum(forest$sample.weights)
+  # Case 1: No sample.weights
+  if (is.null(forest$sample.weights)) {
+    if (length(forest$clusters) == 0) {
+      raw.weights <- rep(1, length(forest$Y.orig))
+    } else {
+      # If clustering with no sample.weights provided, give each observation weight 1/cluster size, so that
+      # the total weight of each cluster is the same.
+      sample.weights <- rep(1, length(forest$Y.orig))
```
swager: We don't need to define sample weights here

r-package/grf/R/input_utilities.R
```diff
@@ -191,19 +196,30 @@ create_data_matrices <- function(X, outcome = NULL, treatment = NULL,
...
+      clust.factor <- factor(forest$clusters)
+      inverse.counts <- 1 / as.numeric(Matrix::colSums(Matrix::sparse.model.matrix(~ clust.factor + 0)))
+      raw.weights <- sample.weights * inverse.counts[as.numeric(clust.factor)]
```
swager: why not just `raw.weights <- inverse.counts[as.numeric(clust.factor)]`

r-package/grf/R/input_utilities.R
```diff
@@ -191,19 +196,30 @@ create_data_matrices <- function(X, outcome = NULL, treatment = NULL,
 }
 
 observation_weights <- function(forest) {
-  sample.weights <- if (is.null(forest$sample.weights)) {
-    rep(1, length(forest$Y.orig))
-  } else {
-    forest$sample.weights * length(forest$Y.orig) / sum(forest$sample.weights)
+  # Case 1: No sample.weights
+  if (is.null(forest$sample.weights)) {
+    if (length(forest$clusters) == 0) {
```
swager: The condition here should be: `length(forest$clusters) == 0 || !forest$equalize.cluster.weights`

r-package/grf/R/input_utilities.R
```diff
@@ -191,19 +196,30 @@ create_data_matrices <- function(X, outcome = NULL, treatment = NULL,
...
-    clust.factor <- factor(forest$clusters)
-    inverse.counts <- 1 / as.numeric(Matrix::colSums(Matrix::sparse.model.matrix(~ clust.factor + 0)))
-    observation.weight <- sample.weights * inverse.counts[as.numeric(clust.factor)]
+
+  # Case 2: sample.weights provided
+  if (!is.null(forest$sample.weights)) {
+    if (length(forest$clusters) == 0 || !forest$equalize.cluster.weights) {
+      raw.weights <- forest$sample.weights
+    } else {
+      stop("Invalid forest: sample.weights is not allowed when equalize.cluster.weights = TRUE")
```
swager: Should be `stop("Specifying non-null sample.weights is not allowed when equalize.cluster.weights = TRUE")`

r-package/grf/R/input_utilities.R
swager: Which tests fail? I didn't thing normalizing or not normalizing the weights mattered, but thought it'd be safer to return normalized weights just in case (always good to be strict on outputs).

r-package/grf/R/instrumental_forest.R
```diff
@@ -25,14 +25,15 @@
...
-#' @param samples.per.cluster If sampling by cluster, the number of observations to be sampled from
-#'                            each cluster when training a tree. If NULL, we set samples.per.cluster to the size
-#'                            of the smallest cluster. If some clusters are smaller than samples.per.cluster,
-#'                            the whole cluster is used every time the cluster is drawn. Note that
-#'                            clusters with less than samples.per.cluster observations get relatively
-#'                            smaller weight than others in training the forest, i.e., the contribution
-#'                            of a given cluster to the final forest scales with the minimum of
-#'                            the number of observations in the cluster and samples.per.cluster. Default is NULL.
+#' @param equalize.cluster.weights If FALSE, each unit is given the same weight (so that bigger
+#'  clusters get more weight). If TRUE, each cluster is given equal weight in the forest. In this case,
```
swager: The indentation here (and in other similar spots) is not consistent.

r-package/grf/R/input_utilities.R
```diff
@@ -73,17 +73,22 @@ validate_clusters <- function(clusters, X) {
   clusters
 }
 
-validate_samples_per_cluster <- function(samples.per.cluster, clusters) {
+validate_equalize_cluster_weights <- function(equalize.cluster.weights, clusters, sample.weights) {
```
swager: This function should have a comment, and maybe a different name. Now it's a little confusing: It does two things, validation & returns samples.per.cluster.

@jtibshirani what do you think?

r-package/grf/R/input_utilities.R
erikcs: 10 tests fails: https://travis-ci.com/grf-labs/grf/jobs/259829947#L1564

r-package/grf/R/instrumental_forest.R
erikcs: I thought it was easier to read when the text was not indented all the way to the right (like in for example honesty.prune.leaves, they also do it [here](https://github.com/tidyverse/dplyr/blob/master/R/group-by.r#L30)). If you agree reading from the left is easier we can harmonize the docstring to this format in another PR?

jtibshirani: I agree that we should stick with our current indentation scheme for now, so we don't have inconsistencies within the same method comment. I also think that the indentation scheme you propose is more readable, we could use it for new methods and start to switch existing ones over.

r-package/grf/R/input_utilities.R
jtibshirani: Our `validate_*` methods already do two things at once, maybe we should have used a term like 'prepare' instead of 'validate'. Since this is fairly consistent with our current approach, I think that it's okay to merge.

REFERENCE.md
```diff
@@ -262,15 +262,16 @@ Some additional notes about the behavior of boosted regression forests:
...
 
-When clustering is enabled during training, all subsampling procedures operate on entire clusters as opposed to individual examples. Then, to determine the examples used for performing splitting and populating the leaves, `samples_per_cluster` examples are drawn from the selected clusters. By default, `sample_per_cluster` is equal to the size of the smallest cluster. Concretely, the cluster-robust training procedure proceeds as follows:
+When clustering is enabled during training, all subsampling procedures operate on entire clusters as opposed to individual examples. Then, to determine the examples used for performing splitting and populating the leaves, `samples_per_cluster` examples are drawn from the selected clusters. By default, `samples_per_cluster` is all the observations in the clusters. If `equalize.cluster.weights = TRUE` `samples_per_cluster` is equal to the size of the smallest cluster.
```
jtibshirani: Should we remove the references to `samples_per_cluster` here?

erikcs: As I understood Stefan's comment it should be there since it refers to the number of samples drawn per cluster and not an argument with that name..?

jtibshirani: Ah sorry about that, I didn't see that you discussed it in a previous review. I found it confusing because I thought it referred to the removed option `samples.per.cluster`.

PR #552, opened 2019-10-27
Bump minimum R version to 3.5.0 (#552)

### LINKED ISSUE #549: isFALSE incompatible with R 3.3.2
**Description of the bug**
It looks like that [the isFALSE statement introduced here](https://github.com/grf-labs/grf/blob/96f496b28667025a04d99aae56982e5ee3d4c099/r-package/grf/R/input_utilities.R#L169) is not available in older versions of R (for example, v3.3.2): 

**GRF version**
development

ginward:
closes #549 

Bump minimum R version requirement from 3.3.0 (first grf release) to 3.5.0


erikcs:
Thanks @ginward

PR #561, opened 2019-10-31
Remove reinitialization of responses_by_sample in split_node_internal

### LINKED ISSUE #556: Remove redundant reinitialization of responses_by_sample?
[responses_by_sample](https://github.com/grf-labs/grf/blob/master/core/src/tree/TreeTrainer.cpp#L64) is initialized here but reinitialized [here](https://github.com/grf-labs/grf/blob/master/core/src/tree/TreeTrainer.cpp#L206) at a new split before being refilled by `relabeling_strategy`. Is that unnecessary since `find_best_split` only accesses elements in responses_by_sample by the samples in `node`?

There can be a marginal speed difference with large training data as with honesty off and high sample.fraction:

```R
n=2000
p=10
X=matrix(rnorm(n*p),n,p)
Y=abs(X[,1])+rnorm(n)

system.time(grf <- regression_forest(X, Y, num.trees = 2000, honesty = FALSE,
                                     ci.group.size = 1,
                                     sample.fraction = 0.7))
# user  system elapsed MASTER
# 44.746   0.214   6.956 

# user  system elapsed NO STD::FILL 
# 43.946   0.174   6.807 
```

erikcs:
Performance: avoid clearing this vector after each split. Add a note to RelabelingStrategy.h that after this some entries of responses_by_sample outside the queried sample IDs may be garbage.

`responses_by_sample` is reinitialized in good style between splits because some entries will be garbage, for correctness it is unnecessary since elements are always accessed at the correct index, and profiling shows there is a performance benefit to skipping this.

Screen Shot 2019-10-31 at 15 15 31

Closes #556


erikcs:
Microbenchmark

```R
library(microbenchmark)
set.seed(123)
n <- 10000
p <- 10
Y <- abs(X[, 1]) + rnorm(n)

microbenchmark(regression_forest(X, Y, num.trees = 2000, honesty = FALSE,
                                 ci.group.size = 1,
                                 sample.fraction = 0.5,
                                 mtry = 3,
                                 min.node.size = 5,
                                 num.threads = NULL, seed = 123,
                                 compute.oob.predictions = FALSE),
               times = 10)


# master
Unit: seconds
    min       lq     mean   median       uq      max neval
11.62483 12.56359 13.02556 13.10956 13.58051 14.37497    10

# this branch
Unit: seconds
    min       lq     mean   median       uq      max neval
10.4259 10.66622 10.99918 10.87566 11.20719 12.53335    10
```

ranger

```R
microbenchmark(ranger::ranger(Y ~ ., data = data.frame(X),
                              keep.inbag = TRUE,
                              num.trees = 2000,
                              sample.fraction = 0.5,
                              mtry = 3,
                              min.node.size = 5,
                              num.threads = NULL, seed = 123,
                              oob.error = FALSE),
               times = 10)
Unit: seconds
    min       lq    mean   median       uq      max neval
8.195739 8.821414 9.27015 9.237639 9.609975 10.33471    10
```

core/src/tree/TreeTrainer.cpp
```diff
@@ -203,7 +203,6 @@ bool TreeTrainer::split_node_internal(size_t node,
     return true;
   }
 
-  std::fill(responses_by_sample.begin(), responses_by_sample.end(), 0.0);
```
jtibshirani: I think it'd be good to add a note to `RelabelingStrategy.h` noting that `responses_by_sample` can contain garbage values for indices outside of the given set of sample IDs. This helps define the method contract (which might be surprising to a newcomer).

Issue #553, opened 2019-10-27
average_late with poor overlap or compliance


ginward:
**Is it possible for #490 to calculate Cov[Y, Z | X = x] / Cov[W, Z | X = x] | W=1 , or  Cov[Y, Z | X = x] / Cov[W, Z | X = x] | Z=1? From the covariance term in the denominator, it looks like the denominator will explode.** 

I am estimating average treatment effects with continuous treatments. However, my dataset is imbalanced (only a small set of population is assigned to treatment), meaning that many compliance scores are close to 0. 

I got the following warnings from the `average_late `function: 

```
Warning messages:
1: In average_late(entry.forest) :
  Estimated treatment propensities take values between -1.534 and 3.004 and in particular get very close to 0 or 1. Poor overlap may hurt perfmance for average conditional local average treatment effect estimation.
2: In average_late(entry.forest) :
  The instrument appears to be weak, with some compliance scores as low as -0.1285

```

Because there is no `overlap `or `treated `option available to deal with  the low propensity score issue. Is there a reason why these two options were not implemented?

ginward:
maybe it is less confusing if we could add an overlap option to average partial effect? I understand that the average treatment effect's overlap option also works for average_partial_effect, so it might be relatively easy to do so ...

ginward:
But still, looks like there is still no way to do overlap option for average_late. Any ideas on that?

jtibshirani:
Hello @ginward, I've noticed that you have opened and closed a few issues without any discussion. Could you please hold off on closing an issue until it has been resolved?

If you ended up finding the answer to your question, you can add it in a comment for future reference. We would like to avoid issues that contain open questions with no answer -- many people read over GitHub issues when debugging/ looking for information, and it can be confusing to see closed issues with no resolution.

PR #490, opened 2019-08-21
Estimate average effects using instrumental forests

swager:
Uses an instrumental forest to implement a doubly robust estimator of tau = E[tau(X)], where tau(x) = Cov[Y, Z | X = x] / Cov[W, Z | X = x]. Extending nomenclature from Imbens and Angrist (1994), we call this the average (conditional) local average treatment effect.

The doubly robust estimator is built following the approach in the paper "<a href="https://arxiv.org/pdf/1608.00033.pdf">locally robust semiparametric estimation</a>" by Chernzohukov et al. (2016).

Addresses #358 and #364.

PR #562, opened 2019-11-01
Instrument forest Warning Correction

ginward:
I think originally the warnings given in average_late is incorrect. 

The `if` condition was to compare `z.hat`:

`if (min(subset.Z.hat) <= 0.01 || max(subset.Z.hat) >= 0.99)`

While the warning messages gives the range of W.hat:

`rng <- range(subset.W.hat)`

And the warning on treatment propensities: 

`Estimated treatment propensities take values between
`

However, z.hat is the instrument propensity, not the treatment propensity. I don't understand why we would compare the instrument propensity and gives warnings on treatment propensity. 

Shouldn't we change the if statement to compare `w.hat`, if we gives warnings on `w.hat`? 

I am not sure if this warning message is trying to warn `Z.hat` or `W.hat`.  So I added warning messages for both. 

Note that, I think the warnings on W.hat only makes sense if the variable is binary. So I have also added a condition to check that. 

Referring to #490 and #553

### DIFF
--- a/r-package/grf/DESCRIPTION
+++ b/r-package/grf/DESCRIPTION
@@ -18,7 +18,7 @@ Description: A pluggable package for forest-based statistical estimation and inf
     quantile regression, and treatment effect estimation (optionally using instrumental
     variables).
 Depends:
-    R (>= 3.3.0)
+    R (>= 3.5.0)
 License: GPL-3
 LinkingTo: Rcpp, RcppEigen
 Imports:
--- a/r-package/grf/R/average_late.R
+++ b/r-package/grf/R/average_late.R
@@ -62,6 +62,13 @@ average_late <- function(forest,
           "only implemented for binary instruments."
       ))
   }
+  
+  if (!all(forest$subset.W.orig %in% c(0, 1))) {
+    stop(paste(
+      "Average conditional local average treatment effect estimation",
+      "only implemented for binary treatments"
+    ))
+  }
 
   if (is.null(subset)) {
     subset <- 1:length(forest$Y.hat)
@@ -132,7 +139,7 @@ average_late <- function(forest,
     ))
   }
   
-  if (abs(min(subset.compliance.score)) <= 0.01 * sd(subset.W.orig)) {
+  if (abs(min(subset.compliance.score) <= 0.01 * sd(subset.W.orig)) {
       warning(paste0(
           "The instrument appears to be weak, with some compliance scores as ",
           "low as ", round(min(subset.compliance.score), 4)
--- a/r-package/grf/R/average_late.R
+++ b/r-package/grf/R/average_late.R
@@ -139,7 +139,7 @@ average_late <- function(forest,
     ))
   }
   
-  if (abs(min(subset.compliance.score) <= 0.01 * sd(subset.W.orig)) {
+  if (abs(min(subset.compliance.score)) <= 0.01 * sd(subset.W.orig)) {
       warning(paste0(
           "The instrument appears to be weak, with some compliance scores as ",
           "low as ", round(min(subset.compliance.score), 4)
--- a/r-package/grf/R/average_late.R
+++ b/r-package/grf/R/average_late.R
@@ -111,9 +111,9 @@ average_late <- function(forest,
   subset.weights <- observation.weight[subset]
 
   if (min(subset.Z.hat) <= 0.01 || max(subset.Z.hat) >= 0.99) {
-    rng <- range(subset.W.hat)
+    rng <- range(subset.Z.hat)
     warning(paste0(
-      "Estimated treatment propensities take values between ",
+      "Estimated instrument propensities take values between ",
       round(rng[1], 3), " and ", round(rng[2], 3),
       " and in particular get very close to 0 or 1. ",
       "Poor overlap may hurt perfmance for average conditional local average ",
--- a/r-package/grf/R/average_late.R
+++ b/r-package/grf/R/average_late.R
@@ -63,13 +63,6 @@ average_late <- function(forest,
       ))
   }
   
-  if (!all(forest$subset.W.orig %in% c(0, 1))) {
-    stop(paste(
-      "Average conditional local average treatment effect estimation",
-      "only implemented for binary treatments"
-    ))
-  }
-
   if (is.null(subset)) {
     subset <- 1:length(forest$Y.hat)
   }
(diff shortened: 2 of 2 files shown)


ginward:
Furthermore, I added the absolute value operator for the compliance score warnings: 

`if (abs(min(subset.compliance.score)) <= 0.01 * sd(subset.W.orig))
`
The reason is that, if W is continuous, it is possible that Z will results in a decrease in W, which results in a negative compliance score. Therefore, I think it only makes sense to include an absolute value in front of the compliance score.

swager:
Thanks @ginward. For the first point, the code flow was correct; there's only a typo in the error message, i.e., W.hat being close to 0 or 1 is not a problem.

ginward:
Thanks @swager 

How about the absolute value point?

ginward:
> Thanks @ginward. For the first point, the code flow was correct; there's only a typo in the error message, i.e., W.hat being close to 0 or 1 is not a problem.

If w.hat is binary, there might still be a problem if it is close to 0 and 1?

ginward:
> Thanks @swager
> 
> How about the absolute value point?

Or rather, if Z causes W to decrease, it violates the Monotonicity assumption in the   (Imbens and Angrist 1994) paper: 

Screenshot 2019-11-01 at 20 41 04

Do you see any way to circumvent that? @swager 

Because my instrument could cause the treatment variable to decrease ... Does flipping the sign help?

swager:
That's what the check on the compliance score is for. If the instrument is weak, the compliance score is close to 0. Yes, good catch, we should have an absolute value for the compliance score check.

ginward:
> That's what the check on the compliance score is for. If the instrument is weak, the compliance score is close to 0. Yes, good catch, we should have an absolute value for the compliance score check.

Thanks @swager 

My worry is that, if the monotonicity assumption is violated (i.e. if my instrument causes the continuous treatment to decrease), I cannot use average_late to calculate the average treatment effect anymore.  If I am correct, the average_late function is based on Imbens and Angrist 1994 paper, isn't it? The monotonicity assumption is one of the crucial assumptions in the paper. 

If that is the case, we probably don't need the absolute sign, as the negative value cause will automatically be invalid ... 

Let me know what you think. @swager

ginward:
> That's what the check on the compliance score is for. If the instrument is weak, the compliance score is close to 0. Yes, good catch, we should have an absolute value for the compliance score check.

Thanks @swager

My worry is that, if the monotonicity assumption is violated (i.e. if my instrument causes the continuous treatment to decrease), I cannot use average_late to calculate the average treatment effect anymore.  If I am correct, the average_late function is based on Imbens and Angrist 1994 paper, isn't it? The monotonicity assumption is one of the crucial assumptions in the paper. 

If that is the case, we probably don't need the absolute sign, as the negative value cause will automatically be invalid ... 

Let me know what you think. @swager

ginward:
> That's what the check on the compliance score is for. If the instrument is weak, the compliance score is close to 0. Yes, good catch, we should have an absolute value for the compliance score check.

Thanks @swager 

My worry is that, if the monotonicity assumption is violated (i.e. if my instrument causes the continuous treatment to decrease thus have a negative subset.compliance.score), I cannot use average_late to calculate the average treatment effect anymore.  If I am correct, the average_late function is based on Imbens and Angrist 1994 paper, isn't it? The monotonicity assumption is one of the crucial assumptions in the paper. 

If that is the case, we probably don't need the absolute sign, as the negative value cause will automatically be invalid ... 

Let me know what you think. @swager

ginward:
Thanks @swager 

I suspect that the current implementation of average_late only supports binary treatment only. 

Because the compliance score is estimated as: 

Delta(X) = E[W | X, Z = 1] - E[W | X, Z = 0]     (1)

The definition of compliance score is: 

Delta(X) = P[ [W | X, Z = 1] - E[W | X, Z = 0] >0 ]    (2)

I think (1) only equals (2) when W is binary. However, currently the package only restricted Z to be binary. 

We either need to restrict W as well to be binary as well, or we will need to re-define a compliance score that works for non-binary treatment W.

ginward:
It is also not immediately clear to me how Inverse Compliance Score Weighting applies to non-binary treatment. So probably it is a good idea to restrict treatment W is binary? @swager

ginward:
I also think that the [documentation](https://cran.r-project.org/web/packages/grf/grf.pdf) is not correct: 

"When W is binary and there are no "defiers", Imbens and Angrist
(1994) show that tau(x) can be interpreted as an average treatment effect on compliers."

I think, in the function, you implemented the Inverse Compliance Score Weighted ATE? If that is the case according to Aronow and Carnegie 2013, it is effectively the average treatment effect for all, not only just for the compliers, no?

ginward:
\

> Thanks @swager
> 
> I suspect that the current implementation of `average_late` only supports **binary** treatment only.
> 
> Because the compliance score is estimated as (Athey and Wager 2017):
> 
> `Delta(X) = E[W | X, Z = 1] - E[W | X, Z = 0] (1)`
> 
> The definition of compliance score is (Aronow and Carnegie 2013):
> 
> `Delta(X) = P[ W(Z=1) - W(Z=0) > 0 | X ] (2)`
> 
> I think (1) only equals (2) when `W` is binary. However, currently the package only restricted `Z` to be binary.
> 
> We either need to restrict `W` as well to be binary as well, or we will need to re-define a compliance score that works for non-binary treatment `W`.

I am thinking if it is possible to estimate Compliance Score  when W is continuous with the following approach: 

1) Calculate `Delta(X) = E[W | X, Z = 1] - E[W | X, Z = 0] `

2) Set a dummy `k = 1` when `Delta(X) > 0` and `k = 0` when `Delta(X) < 0`

3) Estimate `P(k=1|X) `with a regression forest, and use `P(k=1|X)` as the compliance score. 

Do you think this will work? @swager

ginward:
I have required W to be binary at this point.

swager:
We should keep this PR discussion on-topic, focused on the first two fixes in the early commit (the typo in the comment, plus the absolute value for the compliance score check).

I'm happy to have a broader conversation about interpretation of local average treatment effects in the context of continuous exposures, but here is not the right place for it.

Also, just to be clear, the function average_late supports continuous treatments. We do not need to (and should not) insert a check that W is binary. (Yes in fact Abadie & Aronow/Carnegie work with a binary treatment, but one can check that the definitions & results generalize seamlessly.)

ginward:
Thanks @swager 

I have removed the check for binary `w` from the commit. Please see if the commit is now ready to be merged.

ginward:
Thanks. 

Can @erikcs merge the request?

Issue #230, opened 2018-04-14
Fully parallelize prediction collection.


jtibshirani:
Currently, only one component of prediction is parallelized across multiple threads: determining which tree leaves each test sample falls in.

We should also parallelize the step in which predictions are collected given this leaf membership information. A non-trivial amount of work is performed during this step, especially with the more complex strategies like `LocalLinearPredictionStrategy`, and since we may additionally compute variance estimates and debiased errors.

Note that in this form of parallelization, test samples would be distributed across threads. This differs from the current instances of parallelization, where trees are split across threads.

erikcs:
- [x] OptimizedPredictionCollector::collect_predictions is handled in #566 
- [ ] DefaultPredictionCollector::collect_predictions: to be done

PR #566, opened 2019-11-10
Parallelize OptimizedPredictionCollector::collect_predictions

erikcs:
First step towards #230, do the following in parallel:

* OptimizedPredictionCollector::collect_predictions

The pattern is the same as in [ForestTrainer::train](https://github.com/grf-labs/grf/blob/master/core/src/forest/ForestTrainer.cpp#L64):

`collect_predictions` keeps signature and the thread prediction collection is done in `collect_predictions_batch` (the PredictionCollector constructor takes `num_threads` as an argument).

Final step of #230: modify DefaultPredictionCollector::collect_predictions in follow up PR


erikcs:
A causal_forest test

```R
n <- 15000
p <- 10
W <- rbinom(n, 1, 0.5)
system.time(c.forest <- causal_forest(X, Y, W, num.threads = NULL))

# Predict using the forest.
nn = 10000
X.test <- matrix(0, nn, p)
X.test[, 1] <- seq(-2, 2, length.out = nn)
```

Timing:

```R
system.time(c.pred <- predict(c.forest, X.test, num.threads = 1))
# user  system elapsed 
# 4.799   0.348   5.155
system.time(c.pred <- predict(c.forest, X.test, num.threads = NULL))
# user  system elapsed 
# 2.929   0.450   1.668 
```

core/src/prediction/collector/OptimizedPredictionCollector.cpp
```diff
@@ -30,15 +32,62 @@ std::vector<Prediction> OptimizedPredictionCollector::collect_predictions(const
...
+                                 estimate_error,
+                                 start_index,
+                                 num_samples_batch));
+  }
+
+  for (auto& future : futures) {
+    std::vector<Prediction> thread_predictions = future.get();
+    predictions.insert(predictions.end(),
+                       std::make_move_iterator(thread_predictions.begin()),
+                       std::make_move_iterator(thread_predictions.end()));
```
erikcs: A note though: the speed is the same without `std::make_move_iterator`, suggesting `Prediction`'s (autogenerated??) move operator is not called?

erikcs: There are some unnecessary copies going on here. I think collect_predictions_batch should return a pointer to the vector of predictions.

erikcs: 3d0c763 has the thread-wise function return a pointer to the possibly "large" object returned. It made no speed difference, but feels more "right"?

This is the diagram of what is happening inside this loop, are there more efficient ways of getting the guys on the right over to the Prediction object on the left?

```
             +-----------------+                                                 +---------------+
             |                 |                                                 |               |
             |                 |                                                 |  Prediction   |
             |                 |                                                 |               |
             |    Prediction   |                                                 |               |
             |    (empty)      |                                                 +---------------+
             |                 |                                 ptr1 +------>   |               |
             |                 |                                                 |               |
             |                 |                                                 |  Prediction   |
             +-----------------+                                                 |               |
             |                 |                <-----------+                    |               |
             |                 |                                                 +---------------+
             |                 |                                                 |               |
             |      ...        |                                                 |               |
             |                 |                                                 |  Prediction   |
             |                 |                                                 |               |
             |                 |                                                 |               |
Predictions  +-----------------+                                                 +---------------+
             |                 |
             |                 |
             |                 |                                                 +---------------+
             |                 |                                                 |               |
             |                 |                                                 |  Prediction   |
             |                 |                                                 |               |
             |                 |                                                 |               |
             |                 |                                                 +---------------+
             +-----------------+                                 ptr2 +------>   |               |
             |                 |                                                 |               |
             |                 |                                                 |  Prediction   |
             |                 |                <-----------+                    |               |
             |                 |                                                 |               |
             |                 |                                                 +---------------+
             |                 |                                                 |               |
             |                 |                                                 |               |
             |                 |                                                 |  Prediction   |
             |                 |                                                 |               |
             |                 |                                                 |               |
             |                 |                                                 +---------------+
             |                 |
             +-----------------+

```

core/src/forest/ForestPredictor.h
```diff
@@ -53,6 +53,8 @@ class ForestPredictor {
   std::vector<Prediction> predict_oob(const Forest& forest,
                                       const Data& data,
                                       bool estimate_variance) const;
+
+  uint num_threads;
```
jtibshirani: Instead of storing this variable here, could we pass `num_threads` through to the `PredictionCollector` constructor as we do for `TreeTraverser`?

core/src/prediction/collector/OptimizedPredictionCollector.cpp
jtibshirani: I'm not sure I totally follow -- maybe we can keep this as-is for now (since it's consistent with the strategy we take elsewhere) and we can plan to discuss the use of move/ copy in person.

PR #573, opened 2019-11-18
Add a comment explaining the small q / large q splitting rule

erikcs:
This regression forest implementation (ranger) balances N vs nlog n depending on where in the tree the split is sought (in large nodes, further up the tree, a linear scan over the N sorted points is quicker than sorting nL values (nL lognL), while further down the tree sorting  nS values is quicker).(more details in chapter 5 in https://arxiv.org/abs/1407.7502)

 For first time readers of the code this comment might be helpful?

### DIFF
--- a/core/src/splitting/RegressionSplittingRule.cpp
+++ b/core/src/splitting/RegressionSplittingRule.cpp
@@ -64,9 +64,6 @@ bool RegressionSplittingRule::find_best_split(const Data& data,
   // For all possible split variables
   for (auto& var : possible_split_vars) {
     // Use faster method for both cases
-    // (tradeoff between nlog(n) and N)
-    // Small q: sorts the data in the node, O(nlog n)
-    // Large q: uses data pre-sorted at the forest initialization, O(N)
     double q = (double) size_node / (double) data.get_num_unique_data_values(var);
     if (q < Q_THRESHOLD) {
       find_best_split_value_small_q(data, node, var, sum_node, size_node, min_child_size,
@@ -197,7 +194,7 @@ void RegressionSplittingRule::find_best_split_value_large_q(const Data& data,
     if (counter[i] == 0) {
       continue;
     }
-
+    
     n_left += counter[i];
     sum_left += sums[i];
 
--- a/core/src/splitting/RegressionSplittingRule.cpp
+++ b/core/src/splitting/RegressionSplittingRule.cpp
@@ -64,6 +64,9 @@ bool RegressionSplittingRule::find_best_split(const Data& data,
   // For all possible split variables
   for (auto& var : possible_split_vars) {
     // Use faster method for both cases
+    // (tradeoff between nlog(n) and N)
+    // Small q: sorts the data in the node, O(nlog n)
+    // Large q: uses data pre-sorted at the forest initialization, O(N)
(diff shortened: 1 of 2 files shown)


core/src/splitting/RegressionSplittingRule.cpp
```diff
@@ -64,6 +64,9 @@ bool RegressionSplittingRule::find_best_split(const Data& data,
   // For all possible split variables
   for (auto& var : possible_split_vars) {
     // Use faster method for both cases
+    // (tradeoff between nlog(n) and N)
```
jtibshirani: Maybe we could move this to the method comment on `find_best_split` in `RegressionSplittingRule.h` and expand it? In particular we could define `n` and `N` and give some background about how this heuristic came about. Right now I think it might be too concise to help a newcomer.

core/src/splitting/RegressionSplittingRule.h
```diff
@@ -32,6 +32,42 @@ class RegressionSplittingRule final: public SplittingRule {
...
+   * @param possible_split_vars: a vector of valid sample IDs.
+   * @param responses_by_sample: a vector of transformed outcomes.
+   * @param samples: a vector of samples at the given node.
+   * @param split_vars: the output of the method, the best split variable, stored at node.
+   * @param split_values: the output of the method, the best split value, stored at node.
+   * @return a boolean that will be true if no best split was found.
+   *
+   * Details:
+   *
+   * At each split variable j, this method calls into `find_best_split_value_small_q` or
```
jtibshirani: This is great!

core/src/splitting/RegressionSplittingRule.h
```diff
@@ -32,6 +32,42 @@ class RegressionSplittingRule final: public SplittingRule {
 
   ~RegressionSplittingRule();
 
+  /**
+   * Finds the best split at a given node in the tree.
+   *
+   * Is called repeatedly to build a tree in a DFS or BFS fashion.
```
jtibshirani: I think we only do breadth-first -- I don't see where `find_best_split` is called in a depth-first manner? Also saying breadth-first instead of BFS would be more accurate (there is no 'search' when you're creating a tree).

core/src/splitting/RegressionSplittingRule.h
```diff
@@ -32,6 +32,42 @@ class RegressionSplittingRule final: public SplittingRule {
...
+   *
+   * @param data: the data matrix containing all test samples.
+   * @param node: the node id in the tree.
+   * @param possible_split_vars: a vector of valid sample IDs.
```
jtibshirani: I don't think this is right -- `possible_split_vars` contains covariate IDs.

core/src/splitting/RegressionSplittingRule.h
```diff
@@ -32,6 +32,42 @@ class RegressionSplittingRule final: public SplittingRule {
...
+   * @param responses_by_sample: a vector of transformed outcomes.
```
jtibshirani: Suggested edit: `responses_by_sample: a map from sample ID to response.` I don't think we need to mention that they were transformed, the splitting rule doesn't care about the relabeling step.

PR #574, opened 2019-11-20
Implement best linear projection of CATE.

swager:
Implements a doubly robust estimate of the best linear projection of the conditional average treatment effect onto a set of covariates. Example:

```R
n <- 2000
p <- 5
W <- rbinom(n, 1, 0.25 + 0.5 * (X[, 1] > 0))
TAU <- pmax(X[, 1], 0)
Y <- TAU * W + X[, 2] + pmin(X[, 3], 0) + rnorm(n)
forest <- causal_forest(X, Y, W, num.trees = 300)

A = X[,1:2]

lm(TAU ~ A)
# Coefficients:
# (Intercept)        A1           A2  
# 0.400541     0.497620    -0.004921  

best_linear_projection(forest, A)
# Best linear projection of the conditional average treatment effect.
# Confidence intervals are cluster- and heteroskedasticity-robust (HC3):
#  
#              Estimate   Std. Error  t value    Pr(>|t|)    
# (Intercept)  0.359338   0.055070    6.5251    8.587e-11 ***
# A1           0.446652   0.056273    7.9372    3.416e-15 ***
# A2          -0.022161   0.055448   -0.3997       0.6894    
```


r-package/grf/tests/testthat/test_forest_summaries.R
```diff
@@ -138,3 +138,33 @@ test_that("causal forest calibration works with clusters", {
...
+  W <- rbinom(n, 1, 0.25 + 0.5 * (X[, 1] > 0))
+  S <- rbinom(n, 1, 0.5)
+  Y <- S * pmax(X[, 1], 0) * W + X[, 2] + pmin(X[, 3], 0) + rnorm(n)
+  forest <- causal_forest(X, Y, W, num.trees = 300)
+  
+  blp.all <- best_linear_projection(forest, X[,1:2])
+  expect_equal(blp.all[1,1], 0.4/2, tol = 0.1)
+  expect_equal(blp.all[2,1], 0.5/2, tol = 0.1)
+  expect_equal(blp.all[3,1], 0.0/2, tol = 0.1)
+
```
davidahirshberg: It's a little opaque using the literal constants .4 and .5 for the regression coefficients of Z * 1(Z > 0) on 1 and Z for standard normal Z.  A comment, or just approximating those coefficients here, would improve readability.

swager:
Thanks @davidahirshberg!

Regarding the normalization -- I thought about it a little, but it wasn't clear to me exactly how we want to do it. For the ATE, we normalize propensity scores for both treated and controls separately, so that sum Wi / hat{e}(Xi) = sum (1 - Wi) / (1 - hat{e}(Xi)) = n. This seems pretty harmless.

But, in the current case, it seems like there's an incentive to perhaps normalize even more aggressively, so that the weighted interactions of Wi and Ai all match up to their full sample analogues, e.g., sum sum Ai Wi / hat{e}(Xi) = sum Ai? However, this starts to be a lot of renormalization... And it might be worth actually attempting to do the second order theory formally to understand when / how this might be expected to help.

My inclination is to go with the simple thing for now, and then potentially update this function if at some point we find that normalization makes a big difference in practice.

susanathey:
I keep getting surprised at how much these normalizations can matter in specific applications, so it may be worth including example code to compare the results across normalizations just so that users can know if it makes a difference in their setting.

davidahirshberg:
Stefan, I'm a little confused by your comment. You can find weights that solve the exact balance conditions sum Ai Wi / hat{e}(Xi) = sum Ai for all A_i, but you can only solve one balance such balance condition by rescaling (normalizing) arbitrarily estimated inverse propensity weights 1/hat{e}(X_i). I suppose we could use different weights for each regression coefficient, but I worry that if you did that, the normalization would cause things like constant shift m(x,w) -> m(x,w) + c to bleed into estimates of the effect of A_i on the treatment effect.

While this is worth thinking about in the future, I don't see a need to wait on that to include the feature. I'm fine with this as-is. Or with an option to use no normalization or the normalization sum Wi / hat{e}(Xi) = sum (1 - Wi) / (1 - hat{e}(Xi)) = n and an example as Susan suggested.

If we were doing balancing weights, I think it'd probably a good idea to use weights that exactly balance each A_i. That should keep baseline main effects of A_i from bleeding into our estimate of treatment effect heterogeneity. I have sketched out how exact balance fits into our AMLE arguments, and it looks like we can get away with o( sqrt(n) ) exact balance constraints without running into trouble. I'm working on a refined characterization for `Minimax Linear Estimation of the Retargeted Mean', so I hope to know more soon. It might be interesting to test this claim in an AMLE implementation of the average derivative analog of this estimand, E[d^2/dwdx_j m(x,w) |_{w=W, x=X}] for m(x,w)=E[Y | X=x, W=w]. That said, what we get by including an exact balance condition can be pretty different from what we get if we don't and then normalize the weights to get exact balance, so this theory wouldn't apply.

davidahirshberg:
Stefan, I'm a little confused by your comment. You can find weights that solve the exact balance conditions sum Ai Wi / hat{e}(Xi) = sum Ai for all A_i, but you can only solve one balance such balance condition by rescaling (normalizing) arbitrarily estimated inverse propensity weights 1/hat{e}(X_i). I suppose we could use different weights for each regression coefficient, but I worry that if you did that, the normalization would cause things like constant shift m(x,w) -> m(x,w) + c to bleed into estimates of the effect of A_i on the treatment effect.

While this is worth thinking about in the future, I don't see a need to wait on that to include the feature. I'm fine with this as-is. Or with an option to use no normalization or the normalization sum Wi / hat{e}(Xi) = sum (1 - Wi) / (1 - hat{e}(Xi)) = n and an example as Susan suggested.

If we were doing balancing weights, I think it'd probably a good idea to use weights that exactly balance each A_i. That should keep baseline main effects of A_i from bleeding into our estimate of treatment effect heterogeneity. I have sketched out how exact balance fits into our AMLE arguments, and it looks like we can get away with o( sqrt(n) ) exact balance constraints without running into trouble. I'm working on a refined characterization for `Minimax Linear Estimation of the Retargeted Mean', so I hope to know more soon. It might be interesting to test this claim in an AMLE implementation of the average derivative analog of this estimand, E[d^2/dwdx_j m(x,w) |_{w=W, x=X}] for m(x,w)=E[Y | X=x, W=w]. That said, what we get by including an exact balance condition can be pretty different from what we get if we don't and then normalize the weights to get exact balance, so this theory wouldn't apply.

swager:
I merged as-is for now. Then, the way it's written, it should be straight-forward to follow up with another commit that uses normalizing weights if we find it to perform better empirically. (Note: If anyone wants to work on this, please start by opening an issue with empirical comparisons of normalized and un-normalized weights for a range of simulation settings, both in terms of MSE and coverage. Then, we can have a conversation around those results.)

PR #576, opened 2019-11-22
Parallelize DefaultPredictionCollector::collect_predictions

### LINKED ISSUE #230: Fully parallelize prediction collection.

erikcs:
Final step of #230, do the following in parallel:

* DefaultPredictionCollector::collect_predictions.

The pattern is the same as in #566.

Local linear forest prediction timing:

```R
n <- 50
p <- 10
Y <- X[, 1] * rnorm(n)
forest <- ll_regression_forest(X, Y)

nn = 10000
Xn = matrix(rnorm(nn * p), nn, p)

system.time(p1 <- predict(forest, Xn)) # master
  user  system elapsed 
  3.574   0.169   3.216 
system.time(p2 <- predict(forest, Xn)) # this
 user  system elapsed 
  4.035   0.221   0.781 
``` 

(Closes #230)


core/src/prediction/collector/DefaultPredictionCollector.cpp
```diff
@@ -15,14 +15,17 @@
...
 
 namespace grf {
 
-DefaultPredictionCollector::DefaultPredictionCollector(std::unique_ptr<DefaultPredictionStrategy> strategy):
-    strategy(std::move(strategy)) {}
+DefaultPredictionCollector::DefaultPredictionCollector(std::unique_ptr<DefaultPredictionStrategy> strategy,
+                                                       uint num_threads):
+    strategy(std::move(strategy)), num_threads(num_threads) {}
 
 std::vector<Prediction> DefaultPredictionCollector::collect_predictions(
```
jtibshirani: This looks like the same logic as in `OptimizedPredictionCollector` -- we could consider moving this method into a shared superclass. To me it's not so much logic to be worth it yet, but is something to keep in mind.

Issue #577, opened 2019-11-23
Look into improving the performance of `get_sample_weights`.


jtibshirani:
As noted in https://github.com/grf-labs/grf/issues/295#issuecomment-435184200, the R method `get_sample_weights` can be fairly slow on a large dataset. We could look into improving the performance through parallelization, and also consider reviving @halflearned's work in https://github.com/grf-labs/grf/pull/360.

erikcs:
The first thing that makes `get_sample_weights` slow is inserting the result into the sparse matrix , which is being resized at every iteration:

<img width="1136" alt="Screen Shot 2019-11-23 at 21 43 53" src="https://user-images.githubusercontent.com/7185264/69490298-62668780-0e3a-11ea-8136-9b2e06daa1f0.png">

halflearned:
As I have written in #359, `unordered_map` does not seem to be a bottleneck for `get_sample_weights` or `DefaultPredictionStrategy`. 

@erikcs' PR #578 already provided a pretty great improvement for `get_sample_weights`. Hoping that's okay with everyone, I'll move to close this now. Please re-open as needed.

Issue #295, opened 2018-09-05
Return mean Y given W, X as well as difference


gannawag:
Currently the predict.causal_forest function returns the predicted treatment effect given the X values for each observation in the new data. 
Is there a way to also return the mean (over all trees) of the leaf-specific mean of Y for both treatment and control groups? 
A previous issue mentions knowing which leaves each observation lands in, it would be easy to back out the leaf-specific means of Y from there.

susanathey:
I agree this would be useful, would be something that researchers might want to plot.  We could also put in the standard errors.  Perhaps this could be turned on as an option; need to think if we are already saving enough information to do this easily.

jtibshirani:
It would also be nice if these quantities were available on a per-tree basisl, so they would be available in the tree visualizations that @Buyannemekh is working on in #286.

gannawag:
Getting closer with this. Is there an easy way to reference the non-centered outcome and treatment variables in the instrumental forest? `observations.get(Observations::OUTCOME, sample)` in `InstrumentalPredictionStrategy.cpp` seems to only reference the centered versions.

swager:
Yup -- like you mention, @gannawag, all the quantities stored by `causal_forest` reference only the centered outcomes and treatments. Moreover, we always assume a continuous treatment (because even if `W` is binary, `W - W.hat` is not binary). The upshot is that getting anything that can be written in terms of these quantities is easy, e.g., the leaf-wise average of `(Y - Y.hat) * (W - W.hat)`, but anything that depends on the uncentered data isn't directly accessible.

If the underlying goal is to get estimates of `mu_w(x) = E[Y(w) | X=x]`, then we recommend using
```
hat{mu}_w(x) = hat{m}(x) + (w - hat{e}(x)) * hat{tau}(x)
```
where `hat{tau}(x)` is the CATE estimate and `hat{m}(x)` and `hat{e}(x)` are the first stages estimates for `E[Y|X=x]` and `E[W|X=x]` respectively (these are the quantities we use for centering). This is what we do in the `average_treatment_effect` function.

More broadly, we expose the forest weights `alpha_i(x)` via the function `get_sample_weights`, so you may be able to use that for richer analyses.

gannawag:
What about this approach:

1. Edit `causal_forest.R` to define the data used in training as `data <- create_data_matrices(X, Y.centered, W.centered, Y)`

[...]

PR #578, opened 2019-11-24
compute_sample_weights: update result insertion based on Eigen docs

erikcs:
Towards #577 point 1 (Rcpp bindings):

Use the Eigen suggestion on effectively filling a sparse matrix https://eigen.tuxfamily.org/dox/group__TutorialSparse.html

Did `reserve(num_neighbors)` as the obvious lower bound on the number of non-zero entries, there may be a better one? (the output weight matrix often looks more dense than sparse)

Time:

```R
n <- 2000
p <- 10
W <- rbinom(n, 1, 0.5)
cf <- causal_forest(X, Y, W)

system.time(w <- get_sample_weights(cf, num.threads = 1))
#master:
# user  system elapsed 
# 41.688  35.603  77.315

#this
#   user  system elapsed
# 1.954   0.051   2.005
```

This kills the Eigen insert bottleneck. Moderate time chunk as before: the slow get operator of the STL hash table.

### DIFF
--- a/r-package/grf/bindings/AnalysisToolsBindings.cpp
+++ b/r-package/grf/bindings/AnalysisToolsBindings.cpp
@@ -70,6 +70,12 @@ Eigen::SparseMatrix<double> compute_sample_weights(Rcpp::List forest_object,
   size_t num_samples = data->get_num_rows();
   size_t num_neighbors = train_data->get_num_rows();
 
+  // From http://eigen.tuxfamily.org/dox/group__TutorialSparse.html:
+  // Filling a sparse matrix effectively
+  typedef Eigen::Triplet<double> T;
+  std::vector<T> tripletList;
+  tripletList.reserve(num_neighbors);
+  // tripletList.reserve(num_samples);
   Eigen::SparseMatrix<double> result(num_samples, num_neighbors);
 
   for (size_t sample = 0; sample < num_samples; sample++) {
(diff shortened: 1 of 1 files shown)


r-package/grf/bindings/AnalysisToolsBindings.cpp
```diff
@@ -70,6 +70,11 @@ Eigen::SparseMatrix<double> compute_sample_weights(Rcpp::List forest_object,
   size_t num_samples = data->get_num_rows();
   size_t num_neighbors = train_data->get_num_rows();
 
+  // From http://eigen.tuxfamily.org/dox/group__TutorialSparse.html:
+  // Filling a sparse matrix effectively
+  typedef Eigen::Triplet<double> T;
+  std::vector<T> tripletList;
```
halflearned: [nit] Stick with snake_cased `triplet_list`?

r-package/grf/bindings/AnalysisToolsBindings.cpp
```diff
@@ -70,6 +70,11 @@ Eigen::SparseMatrix<double> compute_sample_weights(Rcpp::List forest_object,
   size_t num_samples = data->get_num_rows();
   size_t num_neighbors = train_data->get_num_rows();
 
+  // From http://eigen.tuxfamily.org/dox/group__TutorialSparse.html:
+  // Filling a sparse matrix effectively
+  typedef Eigen::Triplet<double> T;
```
jtibshirani: Sorry for the late review, I think we should avoid `typedef` as much as possible -- we don't really use it elsewhere, and it can make things harder to read. I think the `auto` keyword covers a lot of the cases where a `typedef` could help (although it doesn't apply to this example).

jtibshirani:
@gannawag -- you might be interested in this PR, it seems to really help improve the performance of `get_sample_weights`. The improvement will be available in the upcoming 1.0 release. We're tracking further improvements here: https://github.com/grf-labs/grf/issues/295.

r-package/grf/bindings/AnalysisToolsBindings.cpp
erikcs: This is corrected in #580

PR #584, opened 2019-11-30
Correct the 0.10.4 tarball.

jtibshirani:
When preparing the release, I accidentally uploaded a tarball that did not
contain the important bugfix in #519. This commit adds the tarball that was
released to CRAN.


jtibshirani:
Thanks @erikcs for catching this and for double-checking the tarball. Although it's great that it's so simple, it could be nice to introduce a less manual release process to avoid mistakes like these.

Issue #473, opened 2019-08-05
Consider sample weights in tree splitting.


swager:
The current implementation of sample-weighted forests considers sample weights only during the prediction phase, but not during splitting (see #128, #376, #418). For example, in the case of regression forests, given sample weights gamma_i our prediction at a test point x is:

\hat{mu}(x) = sum_{i = 1}^n \gamma_i alpha_i(x) Y_i / sum_{i = 1}^n \gamma_i alpha_i(x),

where the alpha_i(x) are forest weights obtained "as usual" (i.e., without considering the gammas).

In principle, it should be possible to improve power by also considering the gammas in the splitting phase. Of course, doing so requires caution (e.g., we don't want to overfit to a handful of observations with very large weights), but it would be interesting to investigate.

The following script shows the current state of affairs (running 0.10.3). In the case of censored data, using sample weights helps; however, we ought to be able to get closer to the complete data forests with splitting rules that are aware of censoring. (Note, also, that the tree-splits obtained with and without splits are exactly the same; the gain of the weighted forest comes entirely from the prediction phase.)

```R
set.seed(1)
n = 2000; p = 6
X = matrix(rnorm(n*p), n, p)
prob.missing = 0.75 * as.numeric(X[,1] > 0 & X[,2] > 0)
mu = 4 * as.numeric(X[,1] > 0 & X[,2] > 0) +
    as.numeric(X[,3] > 0) + as.numeric(X[,4] > 0)
Y = mu + rnorm(n)
cc = rbinom(n, 1, 1 - prob.missing) == 1

rf.complete = regression_forest(X, Y, seed = 1)
rf.obs = regression_forest(X[cc,], Y[cc], sample.weights = 1/(1 - prob.missing[cc]), seed = 1)
rf.obs.unweight = regression_forest(X[cc,], Y[cc], seed = 1)

X.test = matrix(rnorm(n*p), n, p)
mu.test = 4 * as.numeric(X.test[,1] > 0 & X.test[,2] > 0) +
    as.numeric(X.test[,3] > 0) + as.numeric(X.test[,4] > 0)

mu.hat.complete = predict(rf.complete, X.test)$predictions
mu.hat.obs = predict(rf.obs, X.test)$predictions
mu.hat.obs.unweight = predict(rf.obs.unweight, X.test)$predictions
```

Output...

```R
> sqrt(c(mean(mu.hat.complete - mu.test)^2,
+        mean(mu.hat.obs - mu.test)^2,
[...]

PR #590, opened 2019-12-06
Sample weighted splitting

davidahirshberg:
Use sample weights when splitting for regression and instrumental forests. In a few simulations (see new files in experiments/sample_weighting_examples), this improves MSE by 5-10%.
This is the default when sample weights are passed. It's currently possible to get the old behavior, in which we use sample weights in prediction but not in splitting, by passing negative sample weights. This isn't intended to be left in,  but it's been convenient for switching between the two behaviors for comparison. But we should settle on a what happens to the old behavior. We could remove it completely or make it an option (possibly tunable) in regression_forest, causal_forest, and instrumental_forest.

Tests that behavior is near-exactly as it was in the past fail, as do tests that predictions are invariant to scaling of the sample weights. I think what is happening is that floating point rounding errors in the splitting code are different, and these are amplified by the splitting process, as it is discontinuous. Qualitative behavior, like MSE in tests that behavior is correct, seems to be the same.

Issue: #473


erikcs:
Tests that behavior is exactly as before passes after d7ee499 (removing a constant term from the decrease calculation that might arbitrarily affect precision).

Tests that sample weight scaling does not matter passes after 9cf840d, see if you agree.

This disabled test:

```R
test_that("predictions are invariant to scaling of the sample weights.", {
  n <- 100
  p <- 6
  e <- 1 / (1 + exp(-2 * X[, 1] + 3 * X[, 2]))
  W <- rbinom(n, 1, e)
  tau <- 2 * X[, 1] + X[, 2]
  Y <- W * tau + rnorm(n)
  e.cc <- 1 / (1 + exp(-2 * X[, 1]))
  sample.weights <- 1 / e.cc

  forest.1 <- causal_forest(X, Y, W, sample.weights = sample.weights, seed = 1)
  forest.2 <- causal_forest(X, Y, W, sample.weights = 40 * sample.weights, seed = 1)
  expect_true(all(abs(forest.2$predictions - forest.1$predictions) < 1e-10))
})
```

only fails because of double precision, if you replace the `40` rescaling by something that is a power of two (only bit shifting), for example `512`, it passes.

erikcs:
1. There is some performance degrading, but not striking: [causal_benchmark+large regression_forest](https://gist.github.com/erikcs/d8c62970efc57bdc1c79b7acd4b625f8)
2. Skip added a comparison under experiments for now:

    [test_splits_causal.R](https://github.com/grf-labs/grf/blob/8542af93aa4264e2936aa5029667cdf589cd7eca/experiments/sample_weighting_examples/test_splits_causal.R)

    [test_splits_regression.R](https://github.com/grf-labs/grf/blob/8542af93aa4264e2936aa5029667cdf589cd7eca/experiments/sample_weighting_examples/test_splits_regression.R)

erikcs:
There are extant tests that using inverse probability of censoring weights improves full-data MSE when we fit  to censored data, and that using sample weighting roughly reproduces the effect of weighted subsampling:

https://github.com/grf-labs/grf/blob/master/r-package/grf/tests/testthat/test_regression_forest.R#L202

https://github.com/grf-labs/grf/blob/master/r-package/grf/tests/testthat/test_causal_forest.R#L157

https://github.com/grf-labs/grf/blob/master/r-package/grf/tests/testthat/test_causal_forest.R#L192

We added tests that sample weighted splitting improves over sample weighting with unweighted splitting in earlier commits (7319428 and df1e8fe). However, the hacky way we informed the C code that we wanted sample weighting with unweighted splitting was to pass negative sample weights, and to restore the check that sample weights are nonnegative we had to eliminate this hack and delete the tests that required it. The older tests now show that sample weighting as implemented now, with sample weighted splitting, improves weighted/full-data MSE and roughly reproduces the effect of weighted subsampling.

erikcs:
There is no explicit option in R to use only one splitting rule, but in the existing sample weight tests for regression_forest and causal_forest both big and small Q are called.

jtibshirani:
> Finally, if I understand the perf results, it looks like adding weights everywhere makes us 5% slower.

That seems like an okay trade-off to me.

One thing that jumps out at me from the PR is that we don't have any C++ tests. Since this is a deep and subtle change to core C++ functionality, I think we should have dedicated test coverage. We could test invariants like 'duplicating all observations gives the same results as setting a constant weight of 2.0'. It would also be good to add a characterization test that covers sample weighting, so we don't inadvertently change the behavior in future refactors.

erikcs:
We can't test that the invariant "duplicating all observations gives the same results as setting a constant weight of 2" holds exactly because sample weighting does not take into account the number of sample weight adjusted observations when checking for `min_child_size` or `imbalance_penalty`, so even if all the options are the same to make them comparable (sample.fraction = 1, etc) the splits will not be identical, even though aggregates are very close. 

An R test for an aggregate (like mean) being very close can be added? (Tests for being invariant to sample weight scaling are in place.)

Characterizations tests for regression forest and causal forests with sample weighting on shall be added!

erikcs:
@jtibshirani characterization tests and an exact sample replication invariance check for regression forest has been added. 

The invariance check for causal forest is a bit more lax (it checks that average predictions are close) since I do not think we can get the exact same behavior due to how counts are handled differently. Causal forest also has a another sample replication test in place from before: https://github.com/davidahirshberg/grf/blob/sample-weighted-splitting/r-package/grf/tests/testthat/test_causal_forest.R#L192

There is one issue with sample weighted causal forest characterization test output not being identical if only bigQ/smallQ is active, looking into it.

erikcs:
All characterization tests now pass regardless of which splitting rule (small or Big) is active.

The very rare issue with q/Q sometimes not giving identical splits is as before due to floating point imprecision.

Here is node 8 in one tree out of 2000:

Screen Shot 2020-02-10 at 12 25 12

The largest impurity decrease for q is for for variable 0 at -0.02, big Q also gets the practically identical decrease, but when evaluating the next split candidate (variable 2) the next possible split which should give the identical decrease instead evaluates to slightly larger, and another split is chosen.

core/test/forest/ForestCharacterizationTest.cpp
```diff
@@ -156,6 +156,48 @@ TEST_CASE("causal forest predictions with stable splitting have not changed", "[
   REQUIRE(equal_predictions(predictions, expected_predictions));
 }
 
+TEST_CASE("causal forest predictions with sample weights and stable splitting have not changed", "[causal], [characterization]") {
```
jtibshirani: It's nice that we're exercising `stabilize_splits` here as well.

core/src/splitting/RegressionSplittingRule.cpp
```diff
@@ -37,6 +38,9 @@ RegressionSplittingRule::~RegressionSplittingRule() {
   if (sums != nullptr) {
     delete[] sums;
   }
+  if(weight_sums != nullptr) {
```
jtibshirani: Huge nitpick: missing space here.

jtibshirani:
Thanks for the additional tests! I didn't do a formal review so won't add an 'approval', but I don't have any more questions/ concerns before you merge.

r-package/grf/tests/testthat/test_causal_forest.R
```diff
@@ -250,6 +250,42 @@ test_that("Weighting is roughly equivalent to replication of samples", {
   expect_true(mean(predict(causal.forest.rep, X[test, ]) > 100 + predict(causal.forest.biased, X[test, ])) >= .5)
 })
 
+test_that("sample weighting is practically identical to replicating samples", {
```
jtibshirani: Does this test give additional confidence over the one directly above? If not, we could drop it.

erikcs: It does not, I removed it

erikcs:
For future reference here is the valgrind check for this commit

```

    MB
217.2^                                                                       :
     |                                                                    @#::
     |                                                          :::@:::@:@@#::
     |                                        :@:::::::::::::::::::@:::@:@@#::
     |                           :@@:::@@:@::::@::: : :::: : ::::::@:::@:@@#::
     |                   ::@@:::::@ :::@ :@:: :@::: : :::: : ::::::@:::@:@@#::
     |             @@::::: @ : :::@ :::@ :@:: :@::: : :::: : ::::::@:::@:@@#::
     |             @ : ::: @ : :::@ :::@ :@:: :@::: : :::: : ::::::@:::@:@@#::
     |             @ : ::: @ : :::@ :::@ :@:: :@::: : :::: : ::::::@:::@:@@#::
     |            @@ : ::: @ : :::@ :::@ :@:: :@::: : :::: : ::::::@:::@:@@#::
     |            @@ : ::: @ : :::@ :::@ :@:: :@::: : :::: : ::::::@:::@:@@#::
     |       @@@::@@ : ::: @ : :::@ :::@ :@:: :@::: : :::: : ::::::@:::@:@@#::
     |       @@ : @@ : ::: @ : :::@ :::@ :@:: :@::: : :::: : ::::::@:::@:@@#::
     |       @@ : @@ : ::: @ : :::@ :::@ :@:: :@::: : :::: : ::::::@:::@:@@#::
     |   :: @@@ : @@ : ::: @ : :::@ :::@ :@:: :@::: : :::: : ::::::@:::@:@@#::
     |   : :@@@ : @@ : ::: @ : :::@ :::@ :@:: :@::: : :::: : ::::::@:::@:@@#::
     |   : :@@@ : @@ : ::: @ : :::@ :::@ :@:: :@::: : :::: : ::::::@:::@:@@#::
     |   : :@@@ : @@ : ::: @ : :::@ :::@ :@:: :@::: : :::: : ::::::@:::@:@@#::
     | ::: :@@@ : @@ : ::: @ : :::@ :::@ :@:: :@::: : :::: : ::::::@:::@:@@#::
     | : : :@@@ : @@ : ::: @ : :::@ :::@ :@:: :@::: : :::: : ::::::@:::@:@@#::
   0 +----------------------------------------------------------------------->Gi
     0                                                                   18.22
```

PR #593, opened 2019-12-11
Add validate_newdata to get_sample_weights

erikcs:
Earlier this all went through and produced output (which was wrong)

```R
n <- 500
p <- 10
W <- rbinom(n, 1, 0.5)
c.forest <- causal_forest(X, Y, W)

get_sample_weights(c.forest, matrix(0, 1, 3))
get_sample_weights(c.forest, 42) 
get_sample_weights(c.forest, c(42, 12))
```

The added input check is the same as the one performed in predict.

validate_newdata update: The change to `NCOL` from `ncol` is for case 2) and 3) above: `NCOL(c(42, 12)` is 1. `ncol(c(42,12)` is `NULL`.


jtibshirani:
@erikcs I didn't even know we supported cases (2) and (3), could we instead reject those inputs? It seems cleaner to only accept matrices and data frames.

PR #596, opened 2019-12-12
Update tests/benchmarks/

erikcs:
Update `tests/benchmarks/causal_benchmark.R` with the latest Quip version, and deletes an old unused file.


halflearned:
[As seen on TV](https://en.wikipedia.org/wiki/As_seen_on_TV#/media/File:As_seen_on_TV.svg).

PR #598, opened 2019-12-13
XPtr serialization draft [On hold]

erikcs:
A test hybrid approach between #386 and #408 

Add two new functions `save_grf/load_grf` for storing big forests (similar to what lightGBM does). Serialization is only done in these two functions. The rest of grf use fast forest access with XPtrs.

```R
n <- 100
p <- 10
Y <- X[, 1] * rnorm(n)
rf <- regression_forest(X, Y) # no serialization, holds an XPtr
rf
GRF forest object of type regression_forest
Number of trees: 2000
Number of training samples: 100
Variable importance:
    1     2     3     4     5     6     7     8     9    10
0.240 0.071 0.110 0.093 0.080 0.072 0.069 0.120 0.075 0.070

save_grf(rf, file = "forest.Rds") # serialized now
RF <- load_grf("forest.Rds")  # creates a new XPtr
RF
GRF forest object of type regression_forest
Number of trees: 2000
Number of training samples: 100
Variable importance:
    1     2     3     4     5     6     7     8     9    10
0.240 0.071 0.110 0.093 0.080 0.072 0.069 0.120 0.075 0.070
```

-[R script valgrind leak check](https://gist.github.com/erikcs/674277dc71a152dcf7c1f22a8d8d46d5)

Note:
-Merge forest functionality: previous forests are now invalidated...

Genereal note:
-Existing packages that uses forking to do something, like `parallell`, will not work if it tries to serialize a forest to copy from one R process to another.

Misc.
-Rstudio devtools load_all can be finicky


r-package/grf/R/merge_forests.R
```diff
@@ -32,7 +35,9 @@ merge_forests <- function(forest_list, compute.oob.predictions = TRUE) {
   validate_forest_list(forest_list)
   first_forest <- forest_list[[1]]
 
-  big_forest <- merge(forest_list)
+  big_forest <- merge(lapply(forest_list, get_xptr))
+  big_forest[["_forest"]] <- ForestPtr$new(xptr = big_forest[["_forest"]])
+  lapply(forest_list, clear_xptr) # Invalidate previous forests.
```
jtibshirani: Why did we start to invalidate the old forests?

r-package/grf/bindings/AnalysisToolsBindings.cpp
```diff
@@ -91,34 +91,124 @@ Eigen::SparseMatrix<double> compute_sample_weights(Rcpp::List forest_object,
...
+  Rcpp::List result;
+  result.push_back(drawn_samples.size(), "num_samples");
+  result.push_back(drawn_samples, "drawn_samples");
+  result.push_back(node_objects, "nodes");
+  return result;
+}
+
+// [[Rcpp::export]]
+Rcpp::List merge(const Rcpp::List forest_xptr_list) {
+  // copy-pastes from rather than calls Forest::Merge because
```
jtibshirani: I think we can just delete `Forest::merge`, since it is no longer used anywhere. It'd be good to double-check that the test cases around merging in `ForestSmokeTest` are still covered tests.

r-package/grf/bindings/RcppUtilities.cpp
```diff
@@ -50,13 +55,14 @@ Forest RcppUtilities::deserialize_forest(const Rcpp::List& forest_object) {
...
 }
 
-Rcpp::List RcppUtilities::serialize_forest(Forest& forest) {
+Rcpp::List RcppUtilities::serialize_forest(SEXP forest_xptr) {
+  Rcpp::XPtr<Forest> forest(forest_xptr);
   Rcpp::List result;
 
-  result.push_back(forest.get_ci_group_size(), "_ci_group_size");
-  result.push_back(forest.get_num_variables(), "_num_variables");
+  result.push_back((*forest).get_ci_group_size(), "_ci_group_size");
```
jtibshirani: Generally instead of doing `(*object).method()`, you can write `object->method()`.

r-package/grf/R/merge_forests.R
erikcs: Because merging is a destructive operation: https://github.com/grf-labs/grf/blob/master/core/src/forest/Forest.h#L52 (which was not an issue before when a copy of the forest was merged)

jtibshirani: We caught up offline and decided that we're okay to change merging to a destructive operation. We'll rename the method to something like `merge_and_destroy` to be very clear about the behavior.

erikcs:
Closing this for now as a good design compromise on this has not been reached, and this is perhaps not a very big performance bottle neck in the grand scheme (forest memory, e.g. tree compression seem more useful).

The original issue #187 is still open.

PR #606, opened 2020-01-09
fix bug when using only one variable in A with best_linear_projection

AlexiaJM:
When A has more than one variable with best_linear_projection(), it result in an error. This fixes the bug.

### DIFF
--- a/r-package/grf/R/forest_summary.R
+++ b/r-package/grf/R/forest_summary.R
@@ -177,7 +177,7 @@ best_linear_projection <- function(forest, A = NULL, subset = NULL) {
   
   if (!is.null(A)) {
     if (nrow(A) == length(forest$Y.orig)) {
-      A.subset <- A[subset,]
+      A.subset <- A[subset,,drop=FALSE]
     } else if (nrow(A) == length(subset)) {
       A.subset <- A
     } else {


erikcs:
Thanks @AlexiaJM but if I run this 

```R
n <- 800
p <- 5
W <- rbinom(n, 1, 0.25 + 0.5 * (X[, 1] > 0))
forest <- causal_forest(X, Y, W)
best_linear_projection(forest, X[, 1])
```

there is still an error
```
Error in if (nrow(A) == length(forest$Y.orig)) { : 
  argument is of length zero
```

Running the original grf code with

```R
best_linear_projection(forest, X[, 1, drop = FALSE])
```

works fine?

swager:
Thanks a lot @AlexiaJM and good catch, @erikcs! While we're fixing this, would it still be possible to add a code path to make sure this runs if `A` is a vector?

erikcs:
@AlexiaJM how about adding 

```R
A <- as.matrix(A)
```

right after 

```R
  if (!is.null(A)) {
```

r-package/grf/R/forest_summary.R
```diff
@@ -177,7 +177,7 @@ best_linear_projection <- function(forest, A = NULL, subset = NULL) {
   
   if (!is.null(A)) {
     if (nrow(A) == length(forest$Y.orig)) {
-      A.subset <- A[subset,]
+      A.subset <- A[subset,,drop=FALSE]
```
erikcs: Maybe make consistent with the rest: space between = and ,?

erikcs:
@AlexiaJM are you still available to complete this PR? Thanks

erikcs:
Closing as resolved in #615

PR #614, opened 2020-02-07
Update local linear split implementation

rinafriedberg:
This PR updates naming conventions and adds const references to a few variables in the local linear relabeling strategy. These are a few improvements to the larger PR adding local linear splits, #603.


rinafriedberg:
@jtibshirani I added your updates from the prior branch to a new branch/PR here! Still to discuss before pushing- (a) more offiical ways to mark this as an experimental feature, and (b) any other variables, maybe in the other LL implementation, that should be const? Thank you so much!

r-package/grf/R/local_linear_forest.R
```diff
@@ -5,8 +5,8 @@
 #'
 #' @param X The covariates used in the regression.
 #' @param Y The outcome.
-#' @param ll.splits Optional choice to make forest splits based on ridge residuals as opposed to standard
-#'                  CART splits. Defaults to FALSE.
+#' @param enable.ll.split Optional choice to make forest splits based on ridge residuals as opposed to standard
```
jtibshirani: We've just been putting the experimental notice right after the variable name, for example 
```
#' @param sample.weights (experimental)
```

PR #620, opened 2020-02-20
Add a causal forest diagnostics vignette

erikcs:
This vignette adds two quick diagnostics checks one can do after fitting a model with grf:

* check the histogram of estimated propensity scores
* check the histogram of each Xi in the control/treated group

A note at the end of the main grf vignette refers to the collection of "Tutorials" (renamed from "Examples")

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ In addition, GRF supports 'honest' estimation (where one subset of the data is u
 Some helpful links for getting started:
 
 - The [R package documentation](https://grf-labs.github.io/grf) contains usage examples and method reference.
-- The [GRF reference](REFERENCE.md) gives a detailed description of the GRF algorithm and includes troubleshooting suggestions.
+- The [GRF reference](https://grf-labs.github.io/grf/REFERENCE.html) gives a detailed description of the GRF algorithm and includes troubleshooting suggestions.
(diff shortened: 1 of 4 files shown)


swager:
Thanks @erikcs for preparing this! However, I still have a few comments. Maybe it's easiest to open a new PR to address these?

- I like the idea of having a tutorial for assessing causal forest fit. However, analyses based on overlap (as given here) are only part of the picture. One should also combine them with calibration checks, e.g., via the `test_calibration` function and/or by comparing average effects across estimated CATE quantiles (like we did in the causal forest application paper).

- To evaluate balance in the X-distribution, we need to inverse-weight samples using the propensity score (i.e., weight by 1/W.hat for treated units and by 1/(1 - W.hat) for controls) in order to make the distributions comparable. Then, if the propensity model fits, the weight covariate distributions should match pretty much exactly.

- We should have a more prominent link to these tutorials from the main github landing page. Currently, it took me some time to navigate to the tutorials.

susanathey:
+1 on Stefan's suggestion.  Generally applied researchers want to see covariate balance with and without adjustment.  One other thing which is lower priority, but while I'm thinking about it, would be to implement the suggestions in this paper https://arxiv.org/abs/1702.01250, where we suggest calculating a bias measure  that helps evaluate how much work the propensity and outcome models have to do to get an unbiased estimate, relative to looking at a simple difference-in-means.

PR #627, opened 2020-03-04
Add grf logo

erikcs:
Add grf logo to README and browser icons for pkgdown.

Thank you to [Egil Paulsen](https://www.egilpaulsen.com/) and Gjermund Bohne for making this!!


jtibshirani:
So great we have a logo now! I sent you some ideas offline about how we could adjust the alignment. One other comment -- could we rename the 'diagrams' folder to 'images' or 'resources', since it now contains more than just diagrams?

PR #629, opened 2020-03-04
More informative error message when subset equals one unit in summary functions

### LINKED ISSUE #628: Average partial effect with a continuous treatment and clustered data
I build a causal forest with the option clusters and I would like to compute the average partial effect for a given cluster.
I run the command: 
average_partial_effect( cf
                                         , calibrate.weights = TRUE
                                         , subset = (X$cluster_id_1 == 1)
                                         , num.trees.for.variance = 500
), where cluster_id_1 is a dummy variable indicating if the observations belong to cluster 1.
I get the following error message:
Error in `contrasts<-`(`*tmp*`, value = contr.funs[1 + isOF[nn]]) : 
  contrasts can be applied only to factors with 2 or more levels

erikcs:
When calculating a treatment summary statistic, if a user specifies `subset` to be equal only one of the clusters, there is an error.

```R
n <- 200
p <- 10
W <- rbinom(n, 1, 1 / (1 + exp(-X[, 2]))) + rnorm(n)
cf <- causal_forest(X, Y, W, num.trees = 50, clusters = c(rep(1, 100), rep(2, 100)))
average_partial_effect(cf, subset = cf$clusters == 1)
 Error in `contrasts<-`(`*tmp*`, value = contr.funs[1 + isOF[nn]]) : 
  contrasts can be applied only to factors with 2 or more levels 
```

Add a more informative error message by checking the number of distinct observations after subsetting. Affected functions:

best_linear_projection (fixes previous mistake)
average_partial_effect
average_late
average_treatment_effect

Note: @swager [this line](https://github.com/grf-labs/grf/blob/master/r-package/grf/R/average_partial_effect.R#L123) line will always be false...? (but the output will be the same)

Closes #628 (Thanks @gcasamat)

### DIFF
--- a/r-package/grf/R/average_late.R
+++ b/r-package/grf/R/average_late.R
@@ -111,7 +111,7 @@ average_late <- function(forest,
   subset.weights <- observation.weight[subset]
 
   if (length(unique(subset.clusters)) <= 1) {
-    stop("The passed subset specifies too few distinct units.")
+    stop("The specified subset must contain units from more than one cluster.")
   }
 
   if (min(subset.Z.hat) <= 0.01 || max(subset.Z.hat) >= 0.99) {
--- a/r-package/grf/R/average_late.R
+++ b/r-package/grf/R/average_late.R
@@ -110,8 +110,8 @@ average_late <- function(forest,
   subset.clusters <- clusters[subset]
   subset.weights <- observation.weight[subset]
 
-  if (length(unique(subset.clusters)) <=1) {
-    stop("subset specifies too few distinct units")
+  if (length(unique(subset.clusters)) == 1) {
+    stop("The passed subset specifies too few distinct units.")
   }
 
   if (min(subset.Z.hat) <= 0.01 || max(subset.Z.hat) >= 0.99) {
--- a/r-package/grf/R/average_late.R
+++ b/r-package/grf/R/average_late.R
@@ -110,7 +110,7 @@ average_late <- function(forest,
   subset.clusters <- clusters[subset]
   subset.weights <- observation.weight[subset]
 
-  if (length(unique(subset.clusters)) == 1) {
+  if (length(unique(subset.clusters)) <= 1) {
     stop("The passed subset specifies too few distinct units.")
   }
 
--- a/r-package/grf/R/average_late.R
+++ b/r-package/grf/R/average_late.R
@@ -62,7 +62,7 @@ average_late <- function(forest,
           "only implemented for binary instruments."
       ))
   }
-  
+
   if (is.null(subset)) {
     subset <- 1:length(forest$Y.hat)
   }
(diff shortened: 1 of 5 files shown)


r-package/grf/R/average_late.R
```diff
@@ -62,7 +62,7 @@ average_late <- function(forest,
           "only implemented for binary instruments."
       ))
   }
-
```
erikcs: All team grf @davidahirshberg @swager @halflearned @rinafriedberg : Can you make sure your text editor removes white space when you save? That will keep this from popping up in diffs.

r-package/grf/R/forest_summary.R
```diff
@@ -180,6 +180,10 @@ best_linear_projection <- function(forest,
   subset.weights.raw <- observation.weight[subset]
   subset.weights <- subset.weights.raw / mean(subset.weights.raw)
 
+  if (length(unique(subset.clusters)) == 1) {
```
rinafriedberg: Tiny change, but maybe it's better to move this earlier? So we don't go through the if(cluster.se) loop at line 165 if we only have 1 cluster.

r-package/grf/tests/testthat/test_forest_summaries.R
```diff
@@ -199,6 +199,6 @@ test_that("best linear projection works with edge case input types", {
   # a single covariate
   blp.single.covariate <- best_linear_projection(forest, X[, 1])
   # a forest trained with clusters, and subset equal to only one of the clusters
-  blp.single.cluster <- best_linear_projection(forest, X[, 1], subset = which(forest.clustered$clusters == 1))
-  expect_equal(1, 1)
+  expect_error(best_linear_projection(forest.clustered, X[, 1], subset = which(forest.clustered$clusters == 1)),
```
rinafriedberg: Good call on this test! :)

r-package/grf/R/forest_summary.R
erikcs: I think I am misunderstanding, but doesn't the check have to be done after `clusters[subset]`, that will catch the case if subset = only one observation, but no clusters too? it doesn't matter if there is only one cluster? 

Also I noticed this condition should be <= 1 to catch the case if you accidentally pass a subset that would be all false, like `average_treatment_effect(cf, subset = X[, 1] > 9999999)`

rinafriedberg: Ok I had misunderstood! There is a bug here, just not what I thought before. We get the "too few distinct units" error with:

forest <- causal_forest(X, Y, W, clusters = c(rep(1, 200), rep(2, 200)))
average_partial_effect(forest, subset = (forest$clusters == 1))

This is because each point belongs to cluster 1, so length(unique(subset.clusters)) == 1, but we actually have 200 points; they're just all in the same cluster.

erikcs: Sorry if I have a brainfreeze, but that is the behavior we want to disallow with this PR? I checked with Stefan and in #628 wrote: "And sorry for the confusion, but this should raise an error (but more informative): it is not possible to calculate this estimate for the desired subset as with clustering enabled you treat each cluster as a distinct unit, which here would be the same as asking for the average partial effect for a single observation."

rinafriedberg: Brainfreeze all around tonight! I had thought you said above that it didn't matter if we only have one cluster but must have misunderstood. This PR checks for that correctly

erikcs: Thank you @rinafriedberg!

Issue #633, opened 2020-03-10
Average partial effect per cluster


gcasamat:
I have clustered data with a continuous treatment and I would like to compute the average partial effect at the cluster level. 
This is not possible with the current implementation of the grf software because "with clustering enabled you treat each cluster as a distinct unit, which here would be the same as asking for the average partial effect for a single observation" (see #628). 
I suggested in this previous issue modifying the computation of V.hat in the script average_partial_effect.R. Having received no reply to this suggestion, I presume this is not a valid way to proceed. However I would be very interested to understand why.
More generally, do you have any suggestion on the way to calculate the per cluster average partial effect ?

erikcs:
Sorry, issues are closed automatically when the PR is merged. [average_partial_effect.R](https://github.com/grf-labs/grf/blob/master/r-package/grf/R/average_partial_effect.R) (continuous treatment) is a generalization of the [average_treatment_effect.R](https://github.com/grf-labs/grf/blob/master/r-package/grf/R/average_treatment_effect.R) (binary treatment), which as you see does not require that same variance estimation, but still requires more than one cluster, @swager may have a better explanation?

gcasamat:
Thanks for your reply. What you said is clear to me: with the current implementation of grf, it is not possible to compute the average partial effect in a single cluster. However, it seems that such a computation (in the binary treatment case) is done in the accompanying code of the article [https://arxiv.org/abs/1902.07409](url): school scores are computed by using formula (8) in the paper.

erikcs:
You mean an ATE with only one cluster is computed [here](https://github.com/grf-labs/grf/tree/master/experiments/acic18)? (Sorry, I am not able to find it)

gcasamat:
The histogram drawn below (taken from script.R) could not be interpreted as the distribution of ATEs per school?
pdf("school_hist.pdf")
pardef = par(mar = c(5, 4, 4, 2) + 0.5, cex.lab=1.5, cex.axis=1.5, cex.main=1.5, cex.sub=1.5)
hist(school.score, xlab = "School Treatment Effect Estimate", main = "")
dev.off()

gcasamat:
To summarize, my concern is about the way the variance of W given X is estimated in average_partial_effect.R. 
Currently it is:
variance_forest <- regression_forest(subset.X.orig,
    (subset.W.orig - subset.W.hat)^2,
    clusters = subset.clusters,
    num.trees = num.trees.for.variance
  )
Why couldn't we have:
[...]

PR #615, opened 2020-02-11
Add correction to best_linear_projection for non-binary treatment

erikcs:
Adds the double robust weight correction for continuous treatments to `best_linear_projection` as described in https://github.com/grf-labs/grf/issues/608#issuecomment-582747602.

The user is also allowed to bypass weights calculation by passing the debiasing weights explicitly through `debiasing.weights`.

* The added tests fail with the previous version
* It also passes the following check: hardcode `discrete.W=FALSE` inside the function. Then `best_linear_projection(forest)` ~= `average_treatment_effect(forest)` when `forest` has binary treatment. 
* bugfix: 1) A is a vector 2) clustered forest with subset equal to only one of the clusters

Closes #608

PR #637, opened 2020-03-18
Add optional debiasing.weights to average_partial_effect

erikcs:
As mentioned in #633: add an optional parameter `debiasing.weights` to `average_partial_effect` which allows a user to bypass the forest based variance estimate Var[Wi | Xi = x]  and input custom weights (this is already present in `best_linear_projection`).

@swager as in `best_liner_prediction`, debiasing.weights is here required to be equal to length n https://github.com/grf-labs/grf/pull/615#discussion_r380324198. Is that ok?

### DIFF
--- a/r-package/grf/R/average_partial_effect.R
+++ b/r-package/grf/R/average_partial_effect.R
@@ -113,7 +113,8 @@ average_partial_effect <- function(forest,
     variance_forest <- regression_forest(subset.X.orig,
       (subset.W.orig - subset.W.hat)^2,
       clusters = subset.clusters,
-      num.trees = num.trees.for.variance
+      num.trees = num.trees.for.variance,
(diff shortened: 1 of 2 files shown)


erikcs:
The latest commit allows `debiasing.weights` to be the same length as `subset`, though to not break grf v.1.1.0 behavior, they can still be have length n when `subset` is passed (only the subset elements are used).

The addition "accentuates" one of several small code divergences among the different summary functions (mostly variable naming): here `best_linear_projection` needs an [extra](https://github.com/erikcs/grf/blob/add-weights/r-package/grf/R/forest_summary.R#L202) branch because `weights` is bound to `debiasing.weights`. We can open a "house cleaning" issue to tidy some of this up?

PR #643, opened 2020-03-24
added a draft of llf user guide

rinafriedberg:
This PR adds a guide for users showing basic LLF usage, explaining features, and addressing simple modifications that help this function in large data settings.


erikcs:
If this is a user guide and not code for replicating a paper you may want to add it to the tutorial vignettes instead? (unless you think Travis will take like an hour to build it) Example: https://grf-labs.github.io/grf/articles/diagnostics.html

(steps: 1) move the .Rmd to `grf/r-package/grf/vignettes/` (look at existing files for metadata) 2) add a vignette entry in the [pkgdown-toc](https://github.com/grf-labs/grf/blob/master/r-package/grf/pkgdown/_pkgdown.yml#L9) 3) run `pkgdown::build_articles()` to see that it runs 4) make sure [travis](https://github.com/grf-labs/grf/blob/master/.travis.yml#L14) has the necessary R packages installed for building the vignette).

rinafriedberg:
@erikcs that's a great idea! I think there's two purposes here: (1) guide through general usage and (2) quick guidelines for speeding this up with large data. The former won't take long to compile but the latter still will. Maybe I should split this into two additions: one to the vingettes and one in the experiments folder?

erikcs:
That sounds reasonable, but why can't everything be in the same vignette? Is it not possible to explain the speedup tips without running a big forest?

rinafriedberg:
That would be totally reasonable too! Let me try to write something up :)

r-package/grf/vignettes/llf.Rmd
```diff
@@ -0,0 +1,147 @@
...
+
+mu <- function(x){
+  2 * x[1] + x[2] + x[3] * x[4]
+}
+
+X <- matrix(rnorm(n*p, 0, 1), nrow = n)
+Y <- apply(X, FUN = mu, MARGIN = 1) + sigma * rnorm(n)
+
+X.test <- matrix(rnorm(n*p, 0, 1), nrow = n)
+truth = apply(X.test, FUN = mu, MARGIN = 1)
```
erikcs: You have some`=` mixed in with `<-`.
Also, if you make section headings with `##` then a table of content will be displayed in the [documentation](https://grf-labs.github.io/grf/articles/diagnostics.html)

r-package/grf/vignettes/llf.Rmd
```diff
@@ -0,0 +1,147 @@
...
+set.seed(123)
+```
+
+```{r setup, warning = FALSE, message = FALSE}
+library(grf)
+library(glmnet)
+library(ggplot2)
+```
+
+This document aims to show how to implement local linear forests (LLF). We begin with the standard use case, walking through parameter choices and method details, and then discuss how to use local linear corrections with larger datasets.
```
halflearned: [nitpick] The tutorial will show you how to _use_ local linear forests. The implementation was done by you.

r-package/grf/vignettes/llf.Rmd
```diff
@@ -0,0 +1,147 @@
...
+When we perform LLF predictions, we can either do a standard ridge regression (ll.weight.penalty set to FALSE), or scale by the covariance matrix (ll.weight.penalty set to TRUE). This defaults to FALSE.
+
+```{r}
+results.llf.weighted <- predict(forest, X.test, 
+                               ll.weight.penalty = TRUE) 
+preds.llf.weighted <- results.llf.weighted$predictions
+mean((preds.llf.weighted - truth)**2)
+```
+
+The user can choose to specify a ridge regression parameter ll.lambda. This defaults to NULL, and if not modified will be selected by automatic parameter tuning. In general, we recommend letting the forest tune this parameter, or performing your own cross-validation loop. The exception to this would be for very large datasets.
```
halflearned:
```suggestion
The user can choose to specify a ridge regression parameter ll.lambda. When this variable is not set by the user, it will be selected by automatic parameter tuning. In general, we recommend letting the forest tune this parameter, or performing your own cross-validation loop. The exception to this would be for very large datasets. 
```

r-package/grf/vignettes/llf.Rmd
```diff
@@ -0,0 +1,147 @@
...
+The user can choose to specify a ridge regression parameter ll.lambda. This defaults to NULL, and if not modified will be selected by automatic parameter tuning. In general, we recommend letting the forest tune this parameter, or performing your own cross-validation loop. The exception to this would be for very large datasets. 
+
+```{r}
+results.llf.lambda <- predict(forest, X.test, 
+                             ll.lambda = 0.1) 
+preds.llf.lambda <- results.llf.lambda$predictions
+mean((preds.llf.lambda - truth)**2)
+```
+
+We also consider the role of tree training in local linear forests. A standard CART split minimizes prediction error from predicting leaf-wide averages. Instead, we can use residual splits, which minimize the corresponding prediction errors on ridge regression residuals. This is currently an experimental feature.
```
halflearned: Would it be possible to include a little more about what residual splits do and when we expect them to be beneficial?

r-package/grf/vignettes/llf.Rmd
```diff
@@ -0,0 +1,147 @@
...
+
+Especially with many covariates, it is reasonable to restrict the local regression to only include a few features of interest. We recommend using the lasso. 
+
+```{r}
+# Select covariates 
+lasso.mod <- cv.glmnet(X, Y, alpha = 1)
+lasso.coef <- predict(lasso.mod, type = "nonzero")
+selected <- ifelse(is.null(dim(lasso.coef)), 1:ncol(X), lasso.coef[,1])
+
+# Train LLF
```
halflearned: This says "train" but it looks like we're only predicting.

r-package/grf/vignettes/llf.Rmd
```diff
@@ -0,0 +1,147 @@
...
+llf.lasso.preds <- predict(forest, X.test, 
+                           linear.correction.variables = selected,
+                           ll.weight.penalty = TRUE)
+results.llf.lasso <- llf.lasso.preds$predictions
+mean((results.llf.lasso - truth)**2)
+```
+
+Last, consider variance estimates and confidence intervals. A user must specify estimate.variance to be TRUE. A parameter under the hood is ci.group.size, which must be at least 2 in order to grow variance estimates (defaults to 2); the forest will grow ci.group.size trees on every subsample in order to provide confidence intervals. We use a different data-generating process for easier visualization.
```
halflearned: I think this is not different from the usual GRF setting, so you can probably cut this in the interest of making this about LLFs only.

r-package/grf/vignettes/llf.Rmd
```diff
@@ -0,0 +1,147 @@
...
+ggplot(df, aes(ticks)) + 
+  geom_point(aes(y = preds.llf.var, color = "Local Linear Forest"), show.legend = F, size = 0.6) +
+  geom_line(aes(y = truth)) + 
+  geom_line(aes(y = lower.llf), color = "gray", lty = 2) + 
+  geom_line(aes(y = upper.llf), color = "gray", lty = 2) + 
+  xlab("x") + ylab("y") + theme_bw()
+```
+
+
+A note on larger datasets: Local linear forests work well in low dimensions; when we get more data, however, it starts to take a very long time to run. This is because with n_train and n_test train and test points, we are running n_test regressions with n_train data points each. However, sometimes we still want to use random forests and correct for linear trends. In this case (datasets with roughly 100,000 or more observations, although always context-dependent), selecting a small number linear correction variables is especially important. The current ridge parameter tuning will also take prohibitively long, and so we recommend either setting the value to 0.01 consistently, tuning this on a subset of the data, restricting the range of values considered, or cross-validating using a small number of shallow trees.
```
halflearned: Consider moving this closer to the section where you talk about variable selection and the lasso?

r-package/grf/vignettes/llf.Rmd
```diff
@@ -0,0 +1,147 @@
...
+preds.ll.splits <- predict(ll.forest, X.test)
+
+mse.cart.splits <- mean((preds.cart.splits$predictions - truth)^2)
+mse.ll.splits <- mean((preds.ll.splits$predictions - truth)^2)
+
+mse.cart.splits
+mse.ll.splits
+```
+
+Especially with many covariates, it is reasonable to restrict the local regression to only include a few features of interest. We recommend using the lasso.
```
halflearned: This is pretty neat, I didn't know this was something we were advising people to do.

r-package/grf/vignettes/llf.Rmd
rinafriedberg: Thanks for the suggestion! Let me think about that and try to put some nice visualizations in there. And I'll add section headers back in!

r-package/grf/vignettes/llf.Rmd
rinafriedberg: :)

rinafriedberg:
@halflearned That's a very good question and an interesting discussion point! If you use llf with lambda = infinity, you recover grf regression forests; regression forests can be seen as a special case OF llf.  

Now, llf is also using a forest to fit a weights estimating equation, which fits under the grf framework; but we can't exactly use the grf framework, because that estimating equation contains a function of the weights used to make it local. So, this isn't purely a special case of grf in a theoretical sense, although in a practical sense (and certainly from the perspective of the software) it is a special case of grf. 

It would be good to think about a clear way of presenting this! What makes more sense from your perspective?

rinafriedberg:
@erikcs @halflearned thank you again for so much thoughtful feedback! I would love to hear any feedback you have about the updates here.

r-package/grf/vignettes/llf.Rmd
```diff
@@ -0,0 +1,397 @@
...
+
+cat(paste("Training took", round(time.train, 3), "seconds. \n",
+          "GRF predictions took", round(time.grf, 3), "seconds. \n",
+          "Lasso selection took", round(time.lasso, 3), "seconds. \n",
+          "LLF prediction took", round(time.llf, 3), "seconds. \n",
+          "LLF and lasso all in all took", round(time.lasso + time.llf, 3), "seconds."))
+
+cat(paste("GRF predictions had MSE", round(mse.grf, 3), "\n",
+          "LLF predictions had MSE", round(mse.llf, 3)))
+```
```
halflearned, line 396: Consider adding a commented output of this `cat` here?

.idea/codeStyles/Project.xml
```diff
@@ -0,0 +1,29 @@
+<component name="ProjectCodeStyleConfiguration">
```
erikcs, line 1: There are some files added by accident here.

r-package/grf/vignettes/llf.Rmd
```diff
@@ -0,0 +1,414 @@
...
+p <- 5
+XX <- matrix(runif(n * p, 0 ,1), nrow = n)
+YY <- apply(XX, MARGIN = 1, FUN = mu) + sigma * rnorm(n)
+
+forest2 <- regression_forest(XX, YY)
+
+max.depth <- 4
+freqs <- split_frequencies(forest2, max.depth = max.depth)
+d <- data.frame(freqs)
+dm <- melt(d)
```
erikcs, line 232: Instead of installing reshape2 on travis, could you reshape it in base R instead?
```
dm <- data.frame(variable = sort(rep(names(d), nrow(d))),
                 value = as.vector(as.matrix(d)))
```

.idea/codeStyles/Project.xml
rinafriedberg: Thank you for catching this!!

rinafriedberg:
@swager If you have a second, can you let me know if this looks good to you? (Or ok Erik and Vitor's approvals?) Thanks so much! :)

PR #647, opened 2020-04-13
Add survival forests

erikcs:
This PR adds `survival_forest(X, Y, D)` functionality to grf (`D` is the censoring indicator).

#### Splitting

* `SurvivalSplittingRule.cpp`
-Implements splitting with the [logrank](https://en.wikipedia.org/wiki/Logrank_test) statistic, as [Iswaran et. al (2008)](https://kogalur.github.io/randomForestSRC/theory.html)
-Split constraints: requires `min_child_size` failures in each node (where `min_child_size = alpha * node_size`, the default grf `alpha` is 0.05). Note the parameter `min.node.size` controls something different: it is the terminal observation count considered before entering the splitting function.
-`imbalance_penalty` is left out
-Missing values are handled with MIA splitting (verified with the same correctness exercise as in #612)
-Short description added to DEVELOPING.md

#### Prediction

* `SurvivalPredictionStrategy.cpp`
-Kaplan-Meier terminal node estimator of the survivor function (with honesty on, there may be leaf nodes with either zero or all failures)
-No Variance estimates (for now).
-Counts are adjusted by `sample.weights` (the same way the `survival` package estimates Kaplan-Meier with weights)

#### Misc.

* Behavior: if `survival_forest(X, Y, C)` is passed a data set with zero failures, then the returned predictions will be empty. 
* The default number of trees is 1000 and not 2000 like other grf forests (since there are no variance estimates, and survival splitting is slower than regression splitting)
* Default `min.node.size` is 15 (reasonable default from randomForestSRC)

_Checks_:
* C++ test: 
a) Characterization tests are added for survival forests with and without NaNs 
b) SplittingRuleInvarianceTest.cpp
c) SurvivalSplittingRuleTest.cpp
d) SurvivalPredictionStrategyTest.cpp

* Performance: [survival_benchmark.R](https://gist.github.com/erikcs/a59432730e6bd774dc1d2c9e8dfb1ff0)
* Comparison with `randomForestSRC`: [survival_comparison.R](https://gist.github.com/erikcs/89bdf47509fbeddd2f62f986af90e4f7)
* MSE check: coming

------------------

#### Future follow ups
- Prediction functionality: Nelson-Aalen conditional hazard function?
- Tuning functionality: needs error estimates
- Sample weighted splitting? (Is not implemented here because of some awkward properties of the logrank statistic, duplicating all samples does not maintain the split rank ordering, not invariant to rescaling of the weights)
- Performance: can improve logrank updating with segment trees (future follow-up project with @kanodiaayush). Surival forests are slow on large data because for a candidate variable splitting scales with O(nm) where n is the number of split points and m is the number of failure points. With large continuous data, in nodes with many failures this is close to quadratic. The logrank statistic can be updated in log(m) time using segment trees, drastically reducing this term to O(nlog m).


core/src/splitting/SplittingRule.h
```diff
@@ -51,6 +51,14 @@ class SplittingRule {
...
                                std::vector<double>& split_values,
                                std::vector<bool>& send_missing_left) = 0;
+
+ /**
+  * This is an optional public member used for C++ unit testing purposes.
+  *
+  * For example, Survival Splitting sets this to the best logrank statistic
+  * and verifies this is correct in a unit test.
+  */
+  double test_statistic;
```
jtibshirani, line 61: Can we move this to SurvivalSplittingRule ?

core/test/prediction/SurvivalPredictionStrategyTest.cpp
```diff
@@ -0,0 +1,156 @@
...
+#include "commons/Data.h"
+#include "commons/utility.h"
+#include "prediction/SurvivalPredictionStrategy.h"
+
+#include "catch.hpp"
+
+using namespace grf;
+
+TEST_CASE("Kaplan-Meier survival estimates are correct", "[survival], [prediction]") {
+  /*
```
jtibshirani, line 27: We could move this into a README too, like we do for splitting and characterization tests?

r-package/grf/R/input_utilities.R
```diff
@@ -30,27 +30,34 @@ validate_X <- function(X, allow.na = FALSE) {
   has.missing.values
 }
 
-validate_observations <- function(V, X) {
+validate_observations <- function(V, X, binary = FALSE) {
   if (is.matrix(V) && ncol(V) == 1) {
     V <- as.vector(V)
   } else if (!is.vector(V)) {
-    stop(paste("Observations (W, Y, or Z) must be vectors."))
+    stop(paste("Observations (W, Y, Z or C) must be vectors."))
```
jtibshirani, line 37: Instead of `C`, I think we should have `delta`. One aspect I liked about the old naming scheme is that observation vectors are always single uppercase characters -- would calling this `D` instead of `delta` make sense at all?

r-package/grf/R/input_utilities.R
```diff
@@ -30,27 +30,34 @@ validate_X <- function(X, allow.na = FALSE) {
   has.missing.values
 }
 
-validate_observations <- function(V, X) {
+validate_observations <- function(V, X, binary = FALSE) {
```
jtibshirani, line 33: If this parameter is only used for the censor currently, I think it's simplest to just move the binary check into `survival_forest` itself.

r-package/grf/R/input_utilities.R
```diff
@@ -204,13 +216,18 @@ create_data_matrices <- function(X, outcome = NULL, treatment = NULL,
...
 
   if (inherits(X, "dgCMatrix") && ncol(X) > 1) {
-    sparse.data <- cbind(X, outcome, treatment, instrument, sample.weights)
+    sparse.data <- cbind(X, outcome, treatment, instrument, censor, sample.weights)
   } else {
     X <- as.matrix(X)
-    default.data <- as.matrix(cbind(X, outcome, treatment, instrument, sample.weights))
+    default.data <- as.matrix(cbind(X, outcome, treatment, instrument, censor, sample.weights))
+  }
+  if (train.data) {
```
jtibshirani, line 224: Could you explain this change? (I'm not up-to-speed on the survival forest algorithm and may be missing some context).

core/test/prediction/SurvivalPredictionStrategyTest.cpp
```diff
@@ -0,0 +1,156 @@
...
+  SurvivalPredictionStrategy prediction_strategy(num_failures);
+  PredictionValues prediction_values = prediction_strategy.precompute_prediction_values(leaf_samples, data);
+
+  std::vector<double> predictions = prediction_values.get_all_values()[0];
+  for (size_t i = 0; i < predictions.size(); i++) {
+    REQUIRE(equal_doubles(predictions[i], expected_predictions[i], 1e-10));
+  }
+}
+
+TEST_CASE("Kaplan-Meier estimates on duplicated data is the same as with sample weights equal to two", "[survival], [prediction]") {
```
jtibshirani, line 91: I'm a bit confused as to why this test refers to sample weights when they aren't implemented yet for survival forests ?

r-package/grf/R/survival_forest.R
```diff
@@ -0,0 +1,203 @@
+#' Survival forest
+#'
+#' Trains a forest for right-censored surival data that can be used to estimate the
+#' conditional surival function S(t, x) = P[Y > t | X = x]
+#'
+#' @param X The covariates.
+#' @param Y The response time (may be negative).
+#' @param delta The censoring indicator (1: failure, 0: censored)
+#' @param num.trees Number of trees grown in the forest. Default is 1000.
+#' @param sample.weights (experimental) Weights given to an observation in prediction.
```
jtibshirani, line 10: I think we can omit the `sample.weights` parameter for now.

core/src/splitting/SplittingRule.h
jtibshirani, line 61: Instead of introducing a mutable variable for testing (which is a bit tricky), it's common to factor out a method that computes the intermediate result. In particular, I wonder if we could add a method like `SurvivalSplittingRule#find_best_split_internal` that returns `best_log_rank`, and check its output in unit tests. Then `SurvivalSplittingRule#find_best_split` would delegate to that method.

Another option is to add `test_statistic` as an output variable to `SplittingRule#find_best_split`. We could just add a comment noting that it's used in unit tests.

r-package/grf/R/input_utilities.R
erikcs: I renamed it to `D`, it was `C` before, but Yifan wanted to use `delta`.

r-package/grf/R/input_utilities.R
erikcs: This is only a convince for the `do.call.rcpp` command: 

https://github.com/erikcs/grf/blob/add-survival-forest/r-package/grf/R/survival_forest.R#L202

A follow-up can simplify the other wrappers as well:

https://github.com/grf-labs/grf/blob/master/r-package/grf/R/causal_forest.R#L399

core/test/prediction/SurvivalPredictionStrategyTest.cpp
erikcs: No sample weights are implemented for prediction, but not splitting.

r-package/grf/R/survival_forest.R
erikcs: It says it is used for predictions only, should this be more emphasized?

r-package/grf/R/input_utilities.R
erikcs: And yes, Yifan has tried `surival_forest` out and OK'd it.

core/src/splitting/SplittingRule.h
erikcs: It sounds like adding it as an output to `SplittingRule#find_best_split` would be the simplest option?

r-package/grf/R/input_utilities.R
erikcs: (agree that using a single letter is better for consistency)

core/test/prediction/SurvivalPredictionStrategyTest.cpp
jtibshirani: Thanks for the explanation, I read too fast.

core/src/splitting/SplittingRule.h
jtibshirani: Either one works for me, whatever you think turns out cleanest.

r-package/grf/R/input_utilities.R
jtibshirani: Ah, I see now -- that's in fact simpler. To me it seems clearer if we created a new method specifically for test data like `create_test_matrices` (and renamed this to `create_train_matrices`). The new method could skip the logic related to observations. It looks like very little code would be duplicated.

erikcs: Thanks, that was cleaner, I can add `do.call.rcpp` in the other predicts in another PR!

core/src/splitting/SplittingRule.h
erikcs: I added `SurvivalSplittingRule::find_best_split_internal`, that seemed better!

core/src/splitting/SurvivalSplittingRule.cpp
```diff
@@ -0,0 +1,316 @@
...
+  }
+
+  // The number of unique failure values in this node
+  size_t num_failures = node_failures.size();
+  // If there are no failures or only one failure time there is nothing to do.
+  if (num_failures <= 1) {
+    return true;
+  }
+
+  std::vector<double> failure_values;
```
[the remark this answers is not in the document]
erikcs: I had this originally but inserting into a set was slightly faster (not by very much though). Do you prefer to get rid of the set?

core/src/prediction/SurvivalPredictionStrategy.cpp
```diff
@@ -0,0 +1,99 @@
...
+    // the failure times will always range from 0, ..., num_failures
+    // where num_failures is the count of failures in the training data.
+    std::vector<double> count_failure(num_failures + 1);
+    std::vector<double> count_censor(num_failures + 1);
+    double sum = 0;
+
+    for (auto& sample : leaf_node) {
+      size_t failure_time = data.get_outcome(sample);
+      double sample_weight = data.get_weight(sample);
+      if (data.get_censor(sample)) {
```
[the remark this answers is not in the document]
erikcs: I changed `get_censor` to explicitly return a boolean, is this what you meant?

```
bool Data::get_censor(size_t row) const {
  if (get(row, censor_index.value()) == 1.0) {
    return true;
  } else {
    return false;
  }
}
```

core/src/splitting/SurvivalSplittingRule.cpp
```diff
@@ -0,0 +1,316 @@
...
+                                            size_t node,
+                                            const std::vector<size_t>& possible_split_vars,
+                                            const std::vector<double>& responses_by_sample,
+                                            const std::vector<std::vector<size_t>>& samples,
+                                            std::vector<size_t>& split_vars,
+                                            std::vector<double>& split_values,
+                                            std::vector<bool>& send_missing_left) {
+
+  double unused_test_value; // This output is only used in interal unit tests.
+  return find_best_split_internal(data, node, possible_split_vars, responses_by_sample,
```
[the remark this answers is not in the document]
erikcs: I have modified it to return the best logrank now, is it as you were thinking now?

core/src/prediction/SurvivalPredictionStrategy.cpp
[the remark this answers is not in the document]
jtibshirani: Yes that looks good. Small comment, but we could make it briefer:

bool Data::get_censor(size_t row) const {
  return get(row, censor_index.value()) > 0.0;
}

core/src/splitting/SurvivalSplittingRule.cpp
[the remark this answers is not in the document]
jtibshirani: Oh interesting, part of the reason I mentioned this is that I've seen microbenchmarks suggesting it's faster. I'm fine to keep it the way it is.

core/src/splitting/SurvivalSplittingRule.cpp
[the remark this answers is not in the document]
jtibshirani: This works for me. I was thinking `find_best_split_internal` would still accept `split_vars`, `split_values`, and `send_missing_left`. Now it's a bit inconsistent, some `best_*` values are returned as parameters, while `best_logrank` is not. But I'm really getting into style minutiae here!

core/src/prediction/SurvivalPredictionStrategy.cpp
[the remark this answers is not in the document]
erikcs: Thanks, that makes sense, I also renamed `get_censor` to `is_censored` as that is even clearer?

core/src/splitting/SurvivalSplittingRule.cpp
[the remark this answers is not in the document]
erikcs: Sorry, I read the that sentence as literally it should return something with `return`, I changed it to `void` and have it return everything as passed parameters as before. Now it is more consistent, but it still does not take the `split_vars` vectors, etc, since then it would only `best_logrank=0` outside, and initialize `best_value=0` etc inside, which would be a bit confusing?

core/src/splitting/SurvivalSplittingRule.cpp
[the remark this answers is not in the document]
erikcs: I re-ran the benchmarks and I changed it back to just inserting directly into the vector. I Think my moderate performance difference came from a not very realistic survival data set with very very many duplicated failure times.

erikcs:
Thanks for the review, I am not sure if it is necessary to mark it as experimental, since this is a quite standard survival forest?

I added a section to DEVELOPING for now, but maybe it could be moved to REFERENCE? The reference is currently focused on causal and regression forests, with no section dedicated to a single forest type (for example the Quantile Forest), and I did not find a natural place to mention the Survival Forest.

r-package/grf/R/survival_forest.R
```diff
@@ -0,0 +1,206 @@
...
+#' @param num.threads Number of threads used in training. By default, the number of threads is set
+#'                    to the maximum hardware concurrency.
+#' @param seed The seed of the C++ random number generator.
+#'
+#' @return A trained survival_forest forest object. The attribute `failure.times` contains the unique failure
+#'  times in the data set.
+#'
+#' @examples
+#' \dontrun{
+#' # Train a standard survival forest.
```
swager, line 55: It would be nice to give the user an example where the user can actually see what the forest did. For example, one thing we could do is along these lines:

```R
n <- 2000
p <- 5
latent.time <- exp(0.5 * X[, 1]) * rexp(n)
censor.time <- 2 * rexp(n)
Y <- pmin(latent.time, censor.time)
D <- as.integer(latent.time <= censor.time)
s.forest <- survival_forest(X, Y, D)

# Predict using the forest.
X.test <- matrix(0, 3, p)
X.test[, 1] <- seq(-2, 2, length.out = 3)
s.pred <- predict(s.forest, X.test)

cols = RColorBrewer::brewer.pal(3, "Set1")
plot(NA, NA, xlab = "failure time", ylab = "survival function",
     xlim = range(s.pred$failure.times),
     ylim = c(0, max(s.pred$predictions)))

for(iter in 1:3) {
  lines(s.pred$failure.times, s.pred$predictions[iter,], col = cols[iter], lwd = 2)
  s.true = exp(-s.pred$failure.times / exp(0.5 * X.test[iter,1]))
  lines(s.pred$failure.times, s.true, col = cols[iter], lty = 2, lwd = 2)
}
```

r-package/grf/R/survival_forest.R
```diff
@@ -0,0 +1,206 @@
...
+#' X.test <- matrix(0, 101, p)
+#' X.test[, 1] <- seq(-2, 2, length.out = 101)
+#' s.pred <- predict(s.forest, X.test)
+#'
+#' # Predict on out-of-bag training samples.
+#' s.pred <- predict(s.forest)
+#' }
+#'
+#' @export
+survival_forest <- function(X, Y, D,
```
swager, line 75: For performance, it'd be helpful if the user could provide a vector of event times at which they want to fit the survival curve; then we could round down the actual even times to the nearest event time in this vector. Otherwise, when we're running this with a sample size of, say, `n = 10000`, it seems like we're forcing the forest to learn the survival function at a needlessly fine granularity.

erikcs: I thought the user could just round the `Y` vector, but this makes much more sense. It is added in commit 70a0e5d

erikcs: An example (from the docstring):

```R
# Train a standard survival forest.
n <- 2000
p <- 5
failure.time <- exp(0.5 * X[, 1]) * rexp(n)
censor.time <- 2 * rexp(n)
Y <- pmin(failure.time, censor.time)
D <- as.integer(failure.time <= censor.time)
s.forest <- survival_forest(X, Y, D)

# Predict on out-of-bag training samples.
s.pred <- predict(s.forest)

# Plot the survival curve for the first five individuals.
matplot(s.pred$failure.times, t(s.pred$predictions[1:5, ]),
        xlab = "failure time", ylab = "survival function (OOB)",
        type = "l", lty = 1)

# Train the forest on a less granular grid.
failure.summary <- summary(Y[D == 1])
events <- seq(failure.summary["Min."], failure.summary["Max."], by = 0.1)
s.forest.grid <- survival_forest(X, Y, D, failure.times = events)
s.pred.grid <- predict(s.forest.grid)
matpoints(s.pred.grid$failure.times, t(s.pred.grid$predictions[1:5, ]),
          type = "l", lty = 2)
```

Screen Shot 2020-04-29 at 10 26 03

r-package/grf/R/survival_forest.R
```diff
@@ -0,0 +1,206 @@
...
+#' X.test[, 1] <- seq(-2, 2, length.out = 101)
+#' s.pred <- predict(s.forest, X.test)
+#'
+#' # Predict on out-of-bag training samples.
+#' s.pred <- predict(s.forest)
+#' }
+#'
+#' @method predict survival_forest
+#' @export
+predict.survival_forest <- function(object, newdata = NULL, num.threads = NULL, ...) {
```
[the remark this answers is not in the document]
erikcs: This would only be for the case `newdata=X.new`? (like `predict(sf, X.test, failure.times = c(1, 5, 20))`) Yes it is possible, I just have to think a bit more.

erikcs:
I am sorry to bring this up now, but I realized that a Survival Forest uses an incredible amount of memory compared to a standard regression forest. It is because I implemented `OptimizedPredictionStrategy`, which saves prediction values, now the forests stores a vector of size `num.failures` in each tree's leaf node, which is a huge memory hog for even moderate numbers of failures.

If I change it to  `DefaultPredictionStrategy` which does not store prediction values, memory footprint is reduced by orders of magnitude. But, OOB prediction can take 5x the forest training time, because of [this loop](https://github.com/grf-labs/grf/blob/master/core/src/prediction/collector/DefaultPredictionCollector.cpp#L98) - I am wondering if this can be improved?

(in-bag prediction with `DefaultPredictionStrategy` is still very fast)

erikcs:
The survival prediction is now done with the DefaultPredictionStrategy, this gives a weighted Kaplan-Meier estimate of the survival curve, using the random forest weights (alpha in the grf paper).

(a change to the prediction collector merged in from master makes this prediction strategy more performant)

r-package/grf/R/survival_forest.R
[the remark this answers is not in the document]
erikcs: I'll leave this for now then, we can add it later if needed!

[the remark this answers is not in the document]
swager: Sounds good

PR #650, opened 2020-05-01
Remove estimate_error from DefaultPredictionCollector

erikcs:
The [DefaultPredictionStrategy](https://github.com/grf-labs/grf/blob/master/core/src/prediction/DefaultPredictionStrategy.h) does not implement [compute_error](https://github.com/grf-labs/grf/blob/master/core/src/prediction/OptimizedPredictionStrategy.h#L104) so we can avoid [recording](https://github.com/grf-labs/grf/blob/master/core/src/prediction/collector/DefaultPredictionCollector.cpp#L98) leaf samples.

When recording leaf samples, also avoid an accidental copy (this will give a speedup for predictions with the DefaultPredictionStrategy).

Future follow up: this step may delegated to the `SampleWeightComputer`

### DIFF
--- a/core/src/prediction/collector/DefaultPredictionCollector.cpp
+++ b/core/src/prediction/collector/DefaultPredictionCollector.cpp
@@ -35,6 +35,7 @@ std::vector<Prediction> DefaultPredictionCollector::collect_predictions(
     const std::vector<std::vector<bool>>& valid_trees_by_sample,
     bool estimate_variance,
     bool estimate_error) const {
+  // Note: estimate_error is not necessary for the DefaultPredictionStrategy and is unused here.
 
   size_t num_samples = data.get_num_rows();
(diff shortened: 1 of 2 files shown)


core/src/prediction/collector/DefaultPredictionCollector.cpp
```diff
@@ -35,6 +35,7 @@ std::vector<Prediction> DefaultPredictionCollector::collect_predictions(
     const std::vector<std::vector<bool>>& valid_trees_by_sample,
     bool estimate_variance,
     bool estimate_error) const {
+  // Note: estimate_error is not necessary for the DefaultPredictionStrategy and is unused here.
```
jtibshirani, line 38: Small comment, I think it'd be cleaner + more visible to add this as a method comment on the interface file `DefaultPredictionCollector.h`

Issue #642, opened 2020-03-23
Consider splitting rule designs


erikcs:
The logic behind the tree splitting can be summarized succinctly in [pseudocode](https://grf-labs.github.io/grf/DEVELOPING.html#tree-splitting-algorithm)

The actual implementation may contain some extra overhead. [RegressionSplittingRule](https://github.com/grf-labs/grf/blob/master/core/src/splitting/RegressionSplittingRule.cpp) and [InstrumentalSplittingRule](https://github.com/grf-labs/grf/blob/master/core/src/splitting/InstrumentalSplittingRule.cpp) calculate the same impurity objective, but incorporate different constraints on valid splits. There may be other designs to consider that factor in this commonality.

erikcs:
This could simplify #660 where we reuse InstrumentalSplittingRule and only add one more split constraint.

PR #660, opened 2020-05-10
Add causal survival forest

erikcs:
#### Overview

* Add `causal_survival_forest(X, Y, W, D)` consisting of
   - `CausalSurvivalRelabelingStrategy` 
  - `CausalSurvivalSplittingRule`
    This is exactly the same as `InstrumentalPredictionRule` but with the additional requirement that there are at least `min_child_size` number of failures in each child node. Details: we use `InstrumentalPredictionStrategy`'s "`W`-constraint" on `W - W.hat`, with the minimum child criterion being `alpha` * "effective size parent". The added `D`-constraint is that there are at least `alpha` * "number of samples in parent" number of failures in each child. (a future simplification is pending #642)
   - `CausalSurvivalPredictionStrategy`

#### Details

- More computations in R wrapper code than earlier forests - more work for potential Python wrappers. 

Since this feature will initially be experimental and subject to change, some details can be changed later, such as:
- All notation consistent with paper or rest of package? (for example, use a different word than `tau` for maximum follow-up time?)
- The way optional nuisance parameters are passed may be a bit finicky: we require `S.hat` and `C.hat` to both correspond to the same grid `U`, since they enter computations where both are indexed by the same time value, but `S0.hat` and `S1.hat` can be fitted on an arbitrary grid since that issue does not arise with these objects. In the future we may just fit everything on the same grid `U` for simplicity.

#### Follow up PR:
- Integration with summary functions
- Best linear projection
- Tuning? (Prediction: `debiased.error` and `excess.error`)
- Update causal_benchmark
- Add paper code to `experiments/`


r-package/grf/tests/testthat/data/test_causal_survival_forest_1.csv
```diff
@@ -0,0 +1,500 @@
+0.444564019146147
```
jtibshirani, line 1: It would be good to make a note somewhere about how this data was generated. And as a small comment, the file names could be more descriptive file names instead of using 1, 12, ...

r-package/grf/tests/testthat/test_causal_survival_forest.R
```diff
@@ -0,0 +1,189 @@
...
+  csf.mia <- causal_survival_forest(X.mia, data$Y, data$W, data$D, num.trees = 500)
+
+  mse.oob.diff <- mean((predict(csf)$predictions - predict(csf.mia)$predictions)^2)
+  mse.diff <- mean((predict(csf, data$X)$predictions - predict(csf.mia, X.mia)$predictions)^2)
+
+  expect_equal(mse.oob.diff, 0, tol = 0.001)
+  expect_equal(mse.diff, 0, tol = 0.001)
+})
+
+# This characterization test locks in behavior for the default causal survival forest.
```
jtibshirani, line 149: Good idea to add this!

core/test/forest/CausalSurvivalForestTest.cpp
```diff
@@ -0,0 +1,50 @@
...
+#include "forest/ForestPredictors.h"
+#include "forest/ForestTrainer.h"
+#include "forest/ForestTrainers.h"
+#include "utilities/ForestTestUtilities.h"
+
+#include "catch.hpp"
+
+using namespace grf;
+
+TEST_CASE("causal survival forests give positive variance estimates", "[causal survival]") {
```
jtibshirani, line 29: When implementing survival forests, we added `SurvivalSplittingRuleTest` and `SurvivalPredictionStrategyTest`. How feasible would it be to add this sort of unit test for causal survival forests?

core/src/splitting/CausalSurvivalSplittingRule.cpp
```diff
@@ -0,0 +1,322 @@
...
+}
+
+bool CausalSurvivalSplittingRule::find_best_split(const Data& data,
+                                                  size_t node,
+                                                  const std::vector<size_t>& possible_split_vars,
+                                                  const std::vector<double>& responses_by_sample,
+                                                  const std::vector<std::vector<size_t>>& samples,
+                                                  std::vector<size_t>& split_vars,
+                                                  std::vector<double>& split_values,
+                                                  std::vector<bool>& send_missing_left) {
```
jtibshirani, line 72: The other splitting rules have tests in `SplitRuleInvarianceTest` for missing values -- we could add a case for this new rule.

r-package/grf/R/causal_survival_forest.R
```diff
@@ -0,0 +1,758 @@
...
+#' T(1) and T(0) are potental outcomes corresponding to the two possible
+#' treatment states.
+#'
+#' @param X The covariates.
+#' @param Y The event time (may be negative).
+#' @param W The treatment assignment (must be a binary vector with no NAs).
+#' @param D The event type (0: censoring, 1: failure).
+#' @param W.hat Estimates of the treatment propensities E[W | Xi]. If W.hat = NULL,
+#'              these are estimated using a separate regression forest. Default is NULL.
+#' @param S1.hat Estimates of the conditional survival function S(t, x, 1) = P[Y > t | X = x, W = 1].
```
jtibshirani, line 16: Would we plausibly expect users to pass in values for `S1.hat`, `S0.hat`, `C.hat`, `lambda.C.hat`? If not, we could avoid making them configurable to reduce the API surface.

core/src/splitting/CausalSurvivalSplittingRule.cpp
erikcs: Yes, I've added one now!

erikcs:
> Could you expand on what you mean here?

Yes, it is only that `S1.hat` and `S0.hat` are lists (first entry is the survival matrix and second the corresponding grid), while the rest are just the survival matrices (but constrained to correspond to the full `Y`-grid, which `S1.hat` and `S0.hat` do not necessarily need)

r-package/grf/R/causal_survival_forest.R
erikcs: It is for consistency with other forests that has nuisance parameters, like `Y.hat` and `W.hat` the user may want to fine tune (change min.node.size etc), or for example, a user can threshold low inverse probability weighting of censoring weights by doing
```
sf.censor = survival_forest(X, Y, 1 - D, prediction.type = "Nelson-Aalen")
C.hat = predict(sf.censor, failure.times = sort(unique(Y)))$predictions
C.hat[C.hat < 0.01] = 0.01
cs.forest = causal_survival_forest(X, Y, W, D,
                                   C.hat = C.hat)
```

and for ` lambda.C.hat`, a user may fine-tune this by calculating derivates from a spline fit instead of the default forward difference.

core/test/forest/CausalSurvivalForestTest.cpp
erikcs: I think it may be trickier with direct tests here because `SurvivalSplitting` had ground truth to compare with (log-rank test statistic for splitting, and Kaplan-Meier estimate for prediction) and this is practically the same as Instrumental/Causal forest where there are indirect test that behavior is reasonable

r-package/grf/tests/testthat/data/README.md
```diff
@@ -0,0 +1,36 @@
+# Datasets
+
+The datasets is this directory were generated through the following R scripts:
```
jtibshirani, line 3: I'm sorry about my incorrect suggestion earlier -- I read too fast and didn't see that all of these CSV files are simply used in characterization tests. This should have been obvious since they all end in `_predictions`!

We could omit the README and go with what you had before then, since it nicely matches the C++ characterization tests.

core/test/forest/CausalSurvivalForestTest.cpp
jtibshirani: That makes sense. I'm okay with the level of test coverage, especially if we have someone test it manually and 'lock in' the behavior with characterization tests. But just to brainstorm, if you're looking for more testing ideas: we could perhaps unit test some of the helper functions like `compute_eta` and the nuisance component calculations?

r-package/grf/R/causal_survival_forest.R
jtibshirani: Okay, this makes sense. I also caught up with @swager and he confirms that it's helpful to be able to pass these in (even just for performance reasons -- you can calculate nuisance parameters once, then try several different forests).

core/test/forest/CausalSurvivalForestTest.cpp
erikcs: Oki. Computing eta is practically just a bunch of tedious algebra: Screen Shot 2020-08-19 at 22 49 58

it involved taking care of a handful of edge cases which is locked in with the current characterization tests. I thought about tests for it, but did not come up with much, like plugging in oracle components for a simple DGP and seeing that the output is reasonable - it is already implicitly verified in the final simulation results in the paper.

core/src/prediction/CausalSurvivalPredictionStrategy.cpp
```diff
@@ -0,0 +1,159 @@
...
+
+size_t CausalSurvivalPredictionStrategy::prediction_length() const {
+    return 1;
+}
+
+std::vector<double> CausalSurvivalPredictionStrategy::predict(const std::vector<double>& average) const {
+  return { average.at(NUMERATOR) / average.at(DENOMINATOR) };
+}
+
+std::vector<double> CausalSurvivalPredictionStrategy::compute_variance(
```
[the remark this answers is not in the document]
erikcs: I am sorry, did you mean 0cf5083 which I corrected now, or something else? I followed the indent from [InstrumentalPredictionStrategy](https://github.com/erikcs/grf/blob/add-causal-sf/core/src/prediction/InstrumentalPredictionStrategy.cpp#L48) where the function parameters where indented 4 spaces?

r-package/grf/tests/testthat/data/README.md
erikcs: Oki, I've deleted this file and added the details to the R test file instead

core/src/prediction/CausalSurvivalPredictionStrategy.cpp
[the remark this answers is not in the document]
erikcs: Sorry about that, it should be fixed in 469805e

core/test/forest/CausalSurvivalForestTest.cpp
erikcs: Also: @yifan-cui and I have done sanity checks by verifying reasonable output with DGPs over the course of this PR.

@jmzeng has also kindly beta tested the package, and as we are marking this first release "Experimental" we can incorporate improvements and suggestions as more people try it out.

jtibshirani: 👍  it sounds like we really have confidence in correctness.

erikcs:
Would anyone object to amend the `causal_survival_forest` optional parameter API:

Instead of

```
S0.hat : a list of ...
S1.hat:  a list of ...
```

Simplify it to:

```
E0.hat: An n-vector of estimates of E[Y | X, W = 1], the expected survival time conditioning on being treated. If NULL (default), then estimated with an S learner using a survival forest. (...)
E1.hat: " "
```

yifan-cui:
@erikcs Looks good to me.

PR #665, opened 2020-05-12
Add  `compute.oob.predictions` option to Quantile Forest

### LINKED ISSUE #663: Add a compute.oob.predictions option to Quantile Forest
Most forests have this option, but Quantile Forest is missing it.

erikcs:
Closes #663 

~2b4c2cd + 9b123e3 is a suggested small breaking change for conformity with the rest of grf: return the point predictions in the `$predictions` attribute (and the quantiles in $quantiles). Can revert if you disagree!~

### DIFF
--- a/r-package/grf/R/quantile_forest.R
+++ b/r-package/grf/R/quantile_forest.R
@@ -188,7 +188,7 @@ predict.quantile_forest <- function(object,
   # If possible, use pre-computed predictions.
   quantiles.orig <- object[["quantiles.orig"]]
   if (is.null(newdata) & identical(quantiles, quantiles.orig) & !is.null(object$predictions)) {
-    return(object$predictions)
+    return(list(predictions = object[["predictions"]]))
   }
 
   num.threads <- validate_num_threads(num.threads)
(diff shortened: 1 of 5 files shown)


r-package/grf/R/quantile_forest.R
```diff
@@ -44,6 +44,7 @@
 #'  Only applies if honesty is enabled. Default is TRUE.
 #' @param alpha A tuning parameter that controls the maximum imbalance of a split. Default is 0.05.
 #' @param imbalance.penalty A tuning parameter that controls how harshly imbalanced splits are penalized. Default is 0.
+#' @param compute.oob.predictions Whether OOB predictions on training set should be precomputed. Default is TRUE.
```
jtibshirani, line 47: I think we should default this to `FALSE`. My understanding is that quantile prediction is substantially more expensive than prediction with other forests, so it would be better precomputing OOB predictions were opt-in for users.

erikcs: I see, I just thought there was some room for breaking interface for a good cause, I have reverted these two changes and added a FALSE by default now.

r-package/grf/R/quantile_forest.R
```diff
@@ -171,6 +185,12 @@ predict.quantile_forest <- function(object,
     stop("Error: Quantiles must be in (0, 1)")
   }
 
+  # If possible, use pre-computed predictions.
+  quantiles.orig <- object[["quantiles.orig"]]
+  if (is.null(newdata) & identical(quantiles, quantiles.orig) & !is.null(object$predictions)) {
```
jtibshirani, line 190: Would `&&` be better than `&` here?

erikcs: I kept it consistent with [this](https://github.com/grf-labs/grf/blob/master/r-package/grf/R/regression_forest.R#L249), but I agree it is not consistent with the style choice in input_utilities.R. I can bring all of them up to date? (it does not matter in these examples as only single-values are compared)

jtibshirani: In case it's helpful context -- for single values I prefer && and || since they short-circuit.

erikcs: I was planning on sending a follow-up PR making all of these consistent.

PR #666, opened 2020-05-12
Add neighbors check to DefaultPredictionCollector

erikcs:
Add a check to the DefaultPredictionCollector which is already in place for the [OptimizedPredictionCollector](https://github.com/grf-labs/grf/blob/master/core/src/prediction/collector/OptimizedPredictionCollector.cpp#L120-L124).

QuantilePredictionStrategy [segfaults](https://user-images.githubusercontent.com/7185264/81639585-367f9400-93d1-11ea-8eba-ff6e757998a7.png) if `weights_by_sample` is empty as is the case when OOB predicting with sample.fraction = 1 and honesty = TRUE as in for example [this](https://github.com/grf-labs/grf/blob/master/r-package/grf/tests/testthat/test_analysis_tools.R#L36) test case.

### DIFF
--- a/core/src/prediction/QuantilePredictionStrategy.cpp
+++ b/core/src/prediction/QuantilePredictionStrategy.cpp
@@ -38,6 +38,10 @@ std::vector<double> QuantilePredictionStrategy::predict(
     const std::unordered_map<size_t, double>& weights_by_sample,
     const Data& train_data,
     const Data& data) const {
+  if (weights_by_sample.empty()) {
+    return std::vector<double>(quantiles.size(), NAN);
+  }
+
   std::vector<std::pair<size_t, double>> samples_and_values;
   for (const auto& entry : weights_by_sample) {
     size_t sample = entry.first;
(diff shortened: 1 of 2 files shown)


core/src/prediction/QuantilePredictionStrategy.cpp
```diff
@@ -38,6 +38,10 @@ std::vector<double> QuantilePredictionStrategy::predict(
     const std::unordered_map<size_t, double>& weights_by_sample,
     const Data& train_data,
     const Data& data) const {
+  if (weights_by_sample.empty()) {
```
jtibshirani, line 41: We could pull this up into `DefaultPredictionCollector` so that all prediction strategies get the benefit of returning early. This is what we do [in `OptimizedPredictionCollector`]( https://github.com/grf-labs/grf/blob/master/core/src/prediction/collector/OptimizedPredictionCollector.cpp#L120-L124).

erikcs: Sorry, that makes much more sense, corrected now.

PR #668, opened 2020-05-13
Predict with training quantiles by default

### LINKED ISSUE #667: Quantile forest prediction should use training quantiles by default
**Description of the bug**
If no quantiles are specified during prediction, then quantile forests default to using 0.1, 0.5, 0.9. It would be better if we used the quantiles that were used to train the forest. Otherwise, it's easy to predict at quantiles that you didn't intend to.

**Steps to reproduce**
```
X = matrix(rnorm(100*10), 100, 10)
Y = rnorm(100)
qrf = quantile_forest(X, Y, quantiles=c(0.2, 0.8))
p = predict(qrf, X)
```

Running `dim(p)` gives
> [1] 100   3

**GRF version**
1.1.0

erikcs:
Closes #667.

```R
X = matrix(rnorm(100*10), 100, 10)
Y = rnorm(100)
qrf = quantile_forest(X, Y, quantiles=c(0.2, 0.8))
p = predict(qrf, X)

dim(p)
#[1] 100   2
```

### DIFF
--- a/r-package/grf/R/quantile_forest.R
+++ b/r-package/grf/R/quantile_forest.R
@@ -182,16 +182,18 @@ predict.quantile_forest <- function(object,
                                     num.threads = NULL, ...) {
   if (is.null(quantiles)) {
     quantiles <- object[["quantiles.orig"]]
-    # If possible, use pre-computed predictions.
-    if (is.null(newdata) & !is.null(object$predictions)) {
-      return(object$predictions)
-    }
   } else {
     if (!is.numeric(quantiles) | length(quantiles) < 1) {
       stop("Error: Must provide numeric quantiles")
     } else if (min(quantiles) <= 0 | max(quantiles) >= 1) {
       stop("Error: Quantiles must be in (0, 1)")
     }
+  }
+
+  # If possible, use pre-computed predictions.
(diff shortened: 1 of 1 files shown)


r-package/grf/R/quantile_forest.R
```diff
@@ -177,18 +178,20 @@ quantile_forest <- function(X, Y,
...
-    stop("Error: Quantiles must be in (0, 1)")
-  }
-
-  # If possible, use pre-computed predictions.
-  quantiles.orig <- object[["quantiles.orig"]]
-  if (is.null(newdata) & identical(quantiles, quantiles.orig) & !is.null(object$predictions)) {
-    return(object$predictions)
+  if (is.null(quantiles)) {
+    quantiles <- object[["quantiles.orig"]]
+    # If possible, use pre-computed predictions.
```
jtibshirani, line 185: Small comment, for me it'd be clearer to keep this as before, outside of this conditional:

```
if (is.null(newdata) & identical(quantiles, quantiles.orig) & !is.null(object$predictions)) {	      
    return(object$predictions)
}
```

That way we have the invariant 'if the training and prediction quantiles are the same, then we use pre-computed OOB values'.

PR #679, opened 2020-05-29
Prepare the 1.2.0 release

erikcs:


erikcs:
This is ready for review once also @davidahirshberg has taken the Survival Forest addition for a test drive.

erikcs:
Does everything else here look OK to you @jtibshirani ? I can change the wording or remove cd75f50 if you think it is unnecessary?

jtibshirani:
This looks ready to merge after the survival forest 'test drive' is done. I'm okay with keeping cd75f50, the description is getting long but I think users will be really interested in this functionality.

erikcs:
Great, as you mentioned earlier, we can streamline the entire README for the 1.3 release!

PR #690, opened 2020-06-11
Replace `prediction.type = c("Kaplan-Meier", "Nelson-Aalen")`

erikcs:
This makes the `prediction.type` choices obvious from the function signature (the default `predict` behavior is to use the `prediction.type` the forest was trained with).


erikcs:
@swager do you think this looks better now?

PR #694, opened 2020-06-18
Remove `samples.per.cluster` option from the ACIC 2018 script

### LINKED ISSUE #693: Error in causal_forest(): unused argument (samples.per.cluster)
**Description of the bug**
Error in causal_forest(): unused argument (samples.per.cluster)

**Steps to reproduce**
To learn, I am using your code for replicating the exercise and after applying it to my data as in https://github.com/grf-labs/grf/blob/master/experiments/acic18/script.R. When running lines 52-56:

cf = causal_forest(X[,selected.idx], Y, W,
                   Y.hat = Y.hat, W.hat = W.hat,
                   clusters = school.id,
                   samples.per.cluster = 50,   
                   tune.parameters = "all")

I get the following error:
'Error in causal_forest(X[, selected.idx], Y, W, Y.hat = Y.hat, W.hat = W.hat,  : 
  unused argument (samples.per.cluster = 50)'

I found in the documentation ([Cluster Std.Err](https://grf-labs.github.io/grf/REFERENCE.html#cluster-robust-estimation)), a reference to the currently available options, and explains that the draw is the total sample within cluster. However, it still makes the case that the argument, samples.per.cluster, can be used. 

I cannot use the option, and I was wondering if it was a bug or an error in my code.
Thanks :)

**GRF version**
1.1.0

erikcs:
The `samples.per.cluster` option was removed in grf version 1.0 and is now by default (`equalize.cluster.weights = FALSE`) set internally to the size of the largest cluster, which means that larger clusters receive larger weights than small ones.

To be closer to the original behavior, this script is updated by setting `equalize.cluster.weights` to `TRUE`, which means each school receives equal weight in ATE estimation.

(closes #693)

### DIFF
--- a/experiments/acic18/script.R
+++ b/experiments/acic18/script.R
@@ -1,8 +1,3 @@
-# Notes:
-# - The paper refers to an argument called `samples.per.cluster`. This option was removed in grf version 1.0
-# and is now by default (`equalize.cluster.weights = FALSE`) internally set to the size of the largest cluster.
-# For more details on clustering in grf, see the algorithm reference at:
-# https://grf-labs.github.io/grf/REFERENCE.html#cluster-robust-estimation
 set.seed(1)
 
 rm(list = ls())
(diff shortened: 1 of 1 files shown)


erikcs:
Does the added `samples.per.cluster` note added to the ACIC 2018 script look OK to you @swager ?

PR #704, opened 2020-07-16
Reproducible tuning without change to global seed

halflearned:
In order to ensure reproducible results when tuning is enabled, the user currently needs to do the following.
```r
# Now
set.seed(1) # messing with global seed
tuned.forest <- regression_forest(X, Y, tune.parameters = "all", seed = 1) # seeding twice
# also global seed is changed again due to calls to runif inside tune_forest
```

This PR is addresses this issue by ensuring that the seed is respected.
```r
# The future
tuned.forest <- regression_forest(X, Y, tune.parameters = "all", seed = 1) # single seed, global seed unchanged
```

**Notes**

+ The `with_seed` function is a simplified version of R.utils [`withSeed`](https://github.com/HenrikBengtsson/R.utils/blob/18f2db262ef5ce6457546782f1f7a1890ca80ab2/R/withSeed.R#L41).
+ Besides the two runif calls, we also fixed the seed during [this](https://github.com/grf-labs/grf/blob/029a692e9a27ab7f8031871e931db3f0380a3ecc/r-package/grf/R/tuning.R#L40) call to `DiceKriging::km`. I wasn't able to ascertain exactly where DiceKriging needs randomness -- perhaps [here](https://github.com/cran/DiceKriging/blob/cf5f3c02b171cf3253ff29c23834d6ecaa78136d/R/kmNuggets.init.R#L22). 
+ Since all tuning functions call `tune_forest`, I added a single test for [here](https://github.com/grf-labs/grf/blob/029a692e9a27ab7f8031871e931db3f0380a3ecc/r-package/grf/tests/testthat/test_regression_forest_tuning.R#L21) for the regression forest case.

Thank you to @zhuzhudiana04 for noticing this issue.

### DIFF
--- a/r-package/grf/R/tune_regression_forest.R
+++ b/r-package/grf/R/tune_regression_forest.R
@@ -155,7 +155,8 @@ tune_regression_forest <- function(X, Y,
                                num.fit.trees = tune.num.trees,
                                num.fit.reps = tune.num.reps,
                                num.optimize.reps = tune.num.draws,
-                               train = train)
+                               train = train,
+                               seed = seed)
 
   tuning.output
 }
--- a/r-package/grf/R/tune_regression_forest.R
+++ b/r-package/grf/R/tune_regression_forest.R
@@ -155,8 +155,7 @@ tune_regression_forest <- function(X, Y,
                                num.fit.trees = tune.num.trees,
                                num.fit.reps = tune.num.reps,
                                num.optimize.reps = tune.num.draws,
-                               train = train,
-                               seed = seed)
+                               train = train)
 
   tuning.output
 }
--- a/r-package/grf/R/tuning.R
+++ b/r-package/grf/R/tuning.R
@@ -28,7 +28,8 @@ tune_forest <- function(data,
                         num.fit.trees,
                         num.fit.reps,
                         num.optimize.reps,
-                        train) {
+                        train,
+                        seed) {
   fit.parameters <- args[!names(args) %in% tune.parameters]
   fit.parameters[["num.trees"]] <- num.fit.trees
   fit.parameters[["ci.group.size"]] <- 1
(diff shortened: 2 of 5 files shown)


r-package/grf/R/tune_regression_forest.R
```diff
@@ -155,7 +155,8 @@ tune_regression_forest <- function(X, Y,
                                num.fit.trees = tune.num.trees,
                                num.fit.reps = tune.num.reps,
                                num.optimize.reps = tune.num.draws,
-                               train = train)
+                               train = train,
+                               seed = seed)
```
erikcs, line 159: Instead of adding this argument you can access it from the `args` list (args$seed), same for the rest!

r-package/grf/R/tuning.R
```diff
@@ -156,3 +161,21 @@ get_tuning_output <- function(status, params, error = NA, grid = NA) {
...
+
+with_seed <- function(expr, seed) {
+  # Record entry seed
+  env <- globalenv()
+  oseed <- env$.Random.seed
+
+  # Restore entry seed on exit
+  on.exit({
+    if (is.null(oseed)) {
+      rm(list=".Random.seed", envir=env, inherits=FALSE)
```
erikcs, line 173: Can make this function consistent with space between all `=` ?

r-package/grf/R/tuning.R
```diff
@@ -66,11 +68,13 @@ tune_forest <- function(data,
   variance.guess <- rep(var(small.forest.errors) / 2, nrow(fit.draws))
   kriging.model <- tryCatch({
     capture.output(
-      model <- DiceKriging::km(
-        design = data.frame(fit.draws),
-        response = small.forest.errors,
-        noise.var = variance.guess
-      )
+      with_seed(
+        model <- DiceKriging::km(
```
erikcs, line 72: From the link you provided it seems randomness in DiceKringing arises when `parinit` is not provided, would an alternative be to find the equivalent value to provide for that instead? (though I guess either way is fine)

(https://www.rdocumentation.org/packages/DiceKriging/versions/1.5.5/topics/km)

r-package/grf/R/tune_regression_forest.R
halflearned: I should have checked that!

r-package/grf/R/tuning.R
halflearned: Left as is -- let me know if you change your mind about either way being fine.

Issue #708, opened 2020-07-27
Multi-class probability estimation


swager:
It would be useful to have a method for multi-class probability estimation. We could use that, for example, for propensity score estimation in `multi_causal_forest` in `policytree`.

One thing to note here is that we already have a ProbabilitySplittingRule implemented in the package (we use it for quantile forests), and this splitting rule accommodates the multi-class setting. So we wouldn't need to write a new splitting rule in order to allow for multi-class probability estimation forests.

erikcs:
a) For splitting: do we simply do ProbabilitySplitting on the class label Yi : [0, 1, ..., num_classes-1]? or on pseudo outcomes based on some score function?

b) For the predictions: do we simply predict the forest weighted class probabilities for each sample? As in:

```
// Prediction for sample i:
// Input: 
// weights_by_sample<size_t, double> : forest weight for all neighbors of sample i
// num_classes : number of classes in Y
// output: num_classes sized vector of predictions of class probabilities
vector<double> predictions(num_classes)
for (entry : weights_by_sample) {
  sample = entry.first
  forest_weight = entry.second
  sample_class = Y[sample] // relabeled to [0, 1, ..., num_classes-1]
  predictions[sample_class] += forest_weight // (potentially adjusted by a sample_weight as well?)
}
return predictions
```

c) The way ProbabilitySplitting is implemented all num_classes is looped over in every split node, with very many classes there may be a moderate speedup by only considering num_classes_in_current_node

erikcs:
- [ ] `probability_forest(X, Y)` is added in #711
- [ ] Update `ProbabilitySplittingRule` to only consider num_classes_in_current_node:

erikcs:
Closing this as we keep an overview of this + more in #733.

PR #711, opened 2020-07-31
Add probability forest

erikcs:
Addresses #708: add `probability_forest(X, Y)`. `Y` may be a category encoded as integer/character/factor.

The only new C++ addition is `ProbabilityPredictionStrategy` which computes class probabilities (optional `sample.weights` is used in prediction) and variance estimates (same as regression forest). Splitting is based on the existing `ProbabilitySplittingRule`.

Since `ProbabilitySplittingRule` is used by quantile forest we already have missing value test coverage through that forest's characterization tests.

_Follow ups_
- (performance) Update `ProbabilitySplittingRule` to only consider num_classes_in_node when splitting? (the usefulness of this depends on how large the number of classes will be for a "typical" use case. Note also that we use the `OptimizedPredictionStrategy` where each prediction vector is stored for every tree's leaf node - a very large amount of classes will be very memory consuming)
- (performance - minor) Add node purity check to `ProbabilitySplittingRule`? As with regression splitting, there is no purity stopping condition (see #546) - whether we should add this here or not (this should not break anything for Quantile Forest where the relabeling strategy ensures that no nodes are pure).
- Incorporate sample weights into `ProbabilitySplittingRule`?
- Tuning?
- Doc related: clarify how sample weighting works (in estimation or just prediction) for different forests in the [GRF Reference](https://github.com/grf-labs/grf/blob/master/REFERENCE.md#sample-weighting)


erikcs:
Thank you for the review @halflearned, hopefully the notes on potential follow-ups could give a slight speed improvement.

Would @swager and @jtibshirani want to have a pass before this is merged?

erikcs:
It is not directly related to classification, which may want a different splitting rule, this is more for propensity estimation for a future multivariate GRF.

Since the splitting part originates from GRF's fork of ranger, it should not be too much to change to update it with sample weights (class weights here: https://github.com/imbs-hl/ranger/blob/master/src/TreeProbability.cpp#L377)

jtibshirani:
>  it should not be too much to change to update it with sample weights...

That makes sense to me. I think we should either fully implement sample weighting or omit it for now. I'm happy to go with whatever you prefer -- if this PR is already substantial enough, we could just omit sample weighting and file a follow-up issue about it?

erikcs:
Ok, but let us wait for @swager's thoughts on sample weights in prediction and splitting here too!

erikcs:
> Let's just document the behavior clearly in the R method docs. After that I have no objections to merging!

In [survival_forest](https://grf-labs.github.io/grf/reference/survival_forest.html) I had written "...Weights given to an observation in _prediction_", but in the other [forests](https://grf-labs.github.io/grf/reference/regression_forest.html) we have "...Weights given to an observation in _estimation_" - I should change that last line to "in _estimation and prediction_"!

jtibshirani:
> Ok, but let us wait for @swager's thoughts on sample weights in prediction and splitting here too!

I checked with @swager earlier and he's on the same page.

erikcs:
Ok, then I suggest go ahead and merge now and address sample weighted splitting in a follow up.

The R docstring currently says "... Weights given to an observation in _prediction_" like in `survival_forest`, do you think that is clear enough?

jtibshirani:
That sounds good, I didn't see that the method doc was already precise. I think we could do a clearer job explaining how sample weighting works for different forests in the algorithm reference, but that's not really in scope for this PR.

xincheng-zhang:
@erikcs This function is very useful ! Thank you very much! But when I reproduce example in 'probability_forest.R' there exist an Error:

Error in do.call.rcpp(probability_train, c(data,args)): object 'probability_train' not found.

WechatIMG133

I don't know how to solve this problem. Hope to get you help, thank you! ^_^

erikcs:
@xincheng-zhang `probability_forest` is not available in the released GRF yet. To use it now you'd have to install from source:

`devtools::install_github("grf-labs/grf", subdir = "r-package/grf")`

(https://grf-labs.github.io/grf/DEVELOPING.html)

xincheng-zhang:
> @xincheng-zhang `probability_forest` is not available in the released GRF yet. To use it now you'd have to install from source:
> 
> `devtools::install_github("grf-labs/grf", subdir = "r-package/grf")`
> 
> (https://grf-labs.github.io/grf/DEVELOPING.html)

Thank you very much!!  ^_^

PR #723, opened 2020-08-25
Unify implementation of ATE-type estimators

swager:
This PR unifies the constructions of different "averaging" estimators that rely on doubly robust scoring rules, including `average_treatment_effect` and `best_linear_projection`.

The motivation for this refactor is that, until now, these functions had largely been written as one-offs, resulting in some inconsistencies. For example:
- Due to small implementation differences, the average treatment effect function didn't give the same point estimate as the best linear projection of the CATE on the treatment (although both point estimates were reasonable on their own).
- Some functions returned CIs that cover population-type estimands (e.g., E[Y(1) - Y(0)]), while others cover sample-type estimands (e.g., 1/n \sum_{i = 1}^n E[Y(1) - Y(0) | Xi]).

Furthermore, statistically sensitive logic (e.g., around construction of doubly robust scores or sample weighting) was needlessly duplicated.

Here, we refactor these functions with the following goal in mind. Each problem type (including causal forests with a binary treatment, causal forests with a continuous treatment, and instrumental forests) are associated with a specific function for computing doubly robust scores. Then, functions like `average_treatment_effect` and `best_linear_projection` abstract away the construction of these scores. We also anticipate building new estimators based on causal survival forests into this framework.

One limitation of this program is that, currently, we implement much more extensive functionality for causal forests with a binary treatment than for other forest types, namely estimators for specific functionals like the ATT and the ATC, as well as an alternative DR construction via TMLE. The code paths for this extra functionality has largely been untouched, and causal forests with binary treatment are only part of the unified architecture when we use AIPW to estimate the ATE.

Finally, this refactor does result in some functionality changes:
- From now on, all functions target population-type quantities (e.g., E[Y(1) - Y(0)]). This will make CIs from functions that used to target sample-type quantities slightly wider.
- Previous functions attempted various ad-hoc normalization schemes for improved precision (e.g., running AIPW with re-normalized propensity weights). Here, for simplicity, we omit any such attempts for improved normalization (e.g., with AIPW, we just take the inverse-propensity score weights as is). However, users can opt to override the construction of these weights (e.g., inverse-propensity score weights) by manually setting the `debiasing.weights` argument.


erikcs:
Just some initial thoughts (have also spent some time on this for the upcoming BLP for causal_survival_forest)
- The optional `debiasing.weights` are only used if `method == "AIPW" && target.sample == "all"`, I thought these would apply to all estimands 
- The IPW's are [renormalized](https://github.com/grf-labs/grf/blob/ate-refactor/r-package/grf/R/average_treatment_effect.R#L326) only if target.sample is "treated" or "control"? (thought the user could do this through `debiasing.weights` now?)
- Having everything in one function (`average_treatment_effect`) is perhaps not much clearer than separating into `average_partial_effect`/etc, at least currently `average_treatment_effect` is still complex
- Is it worth considering splitting this into more than one PR?

As this involves some big breaking API changes, and we have permission from @jtibshirani (potentially reserving it for the next major release),  is it worthwhile to consider other "drastic" changes as well? 
- Given DR scores from a forest (it could be a generic function too), then computing an ATE/etc could internally be delegated to `best_linear_projection` without covariates since the computations involved are the same? (forgetting about this only being implemented for `causal_forest` for now) and return a `lmtest::coeftest` object:


```R
Confidence intervals are cluster- and heteroskedasticity-robust (HC3):
  
     Estimate Std. Error t value Pr(>|t|)  
ATE  0.46423    0.17900  2.5935  0.01021 *
  ---
  Signif. codes:  0 ‘***’ 0.001 ‘**’ 0.01 ‘*’ 0.05 ‘.’ 0.1 ‘ ’ 1
# also show some more forest specific metadata: like for example 
method: AIPW
target: all
etc...?

# Other, like IV:

Confidence intervals are cluster- and heteroskedasticity-robust (HC3):
  
     Estimate Std. Error t value Pr(>|t|)  
LATE  0.46423    0.17900  2.5935  0.01021 *
  ---
  Signif. codes:  0 ‘***’ 0.001 ‘**’ 0.01 ‘*’ 0.05 ‘.’ 0.1 ‘ ’ 1

# etc...
```

this would make BLP and the other summary functions consistent wrt to output, since `lmtest` is an ubiquitous object in the R world, getting ready latex tables could be easier for the user:

```R
texreg::texreg(blp)
xtable::xtable(matrix(blp))
stargazer::stargazer(blp)
knitr::kable(blp[,])
# etc...
```
this would also make all SE's the same, obviating the need to calculate them manually in `average_treatment_effect`?

- Since this is a big change reducing complexity: is it necessary to keep the `TMLE` option? If keep it: can it be baked into `get_scores_ATE`? (given it's feasible to elegantly express the different target computations all/treated/control in this function, that is where my sketch on this stopped) 

- Yet another design option, instead of `average_treatment_effect`, could be to follow an R idiom for estimators by for example making a `summary` method for each forest, then `summary` for `causal_forest` would take extra arguments `target.sample`, `method`, `subset` and could print:
```R
Confidence intervals are cluster- and heteroskedasticity-robust (HC3):
  
     Estimate Std. Error t value Pr(>|t|)  
ATE  0.46423    0.17900  2.5935  0.01021 *
  ---
  Signif. codes:  0 ‘***’ 0.001 ‘**’ 0.01 ‘*’ 0.05 ‘.’ 0.1 ‘ ’ 1
#etc 
```
if table reporting software is properly generic enough (though most is not), it should call summary on an object and get all coefficient info from `$coefficients` and report in it whatever format without problems

(sorry for the long post)

erikcs:
On the last items I am perhaps digressing a reverse-thinking a bit what a suitable output could be, the idea was something that would "fit in" typical workflows of researchers like:

```R
library(grf)
library(texreg)
set.seed(1)

n = 8000;
p = 5
X = matrix(rnorm(n * p), n, p)
colnames(X) = c("age", "blood pressure", "height", "weight", "body mass index (BMI)")
L = 2 / (1 + exp(5 * X[,1] + 5 * X[,2]))
W = L + rnorm(n)
Y = 0.25 * W - 1.6 * L + rnorm(n)

cf = causal_forest(X, Y, W)
estimate.grf = best_linear_projection(cf)
rownames(estimate.grf) = "W"

estimate.ols = lm(Y ~ ., data.frame(Y, W, X))

texreg::texreg(list(OLS=estimate.ols, GRF=estimate.grf))
```
would give:

Screen Shot 2020-08-24 at 21 43 07

(ignoring I renamed ATE to W here) - I think some thought should go into what metrics should be reported for a non-parametric estimate like GRF, anyhow, the idea was to allow for a workflow where researchers could "easily" bake GRF into their coefficient tables if they wanted too

swager:
Thanks @erikcs for your thoughts on this! First, yes, this is a pretty big (and API-breaking) change. I think we should go for it, though. I wrote many of these global analysis tools as one-offs, without really thinking about how they all fit together. Given that we now have so many of them and they play such a big role in the package, I think it's time to clean all this up.

The upshot, of course, is that if we're open to making API-breaking changes here, then we can also think about bigger changes if we believe they improve user experience. Then, we'd probably be looking at cutting a 2.0 release once this + causal survival forests are ready.

Now, in terms of how to fine-tune the API, the way I've been presenting this material recently is that there's a set of high-level tasks (average effect estimation, best linear projections, policy learning). Then there's a set of statistical settings where one might want to carry them out (unconfoundedness with binary treatment, unconfoundedness with continuous treatment, etc.), and the setting is reflected in how we construct doubly robust scores.

I think it'd be nice for our API to reflect this structure. We have functions (`average_treatment_effect`, `best_linear_projection`, etc.) that depend on the high-level tasks, and then dispatch to appropriate doubly robust score calculators depending on the setting. The more I think about it, the more having different functions for, e.g., `average_treatment_effect` vs `average_partial_effect` is mostly just confusing. Both are just trying to do the same thing; having separate functions doesn't seem to give the user any value.

On whether or not to renormalize the weights, here are some thoughts I was working around:
- For the ATT and ATC, renormalization matters asymptotically to first order, and so we have to do it. For the ATE, renormalizing doesn't asymptotically matter to first order, so maybe it helps but it's more ambiguous.
- In `best_linear_projection`, there's no particularly natural way to do renormalization. Also, renormalizing doesn't asymptotically matter to first order. So the case for doing it is really weak.
- From a user perspective, the ATE function and `best_linear_projection` on null predictors should give the same point estimate.

So, putting all this together, the natural solution seems to be to renormalize for the ATT and ATC, but not for ATE and `best_linear_projection`. Maybe it looks a little complicated, but at least currently I'm OK with it. The ATT and ATC are somewhat subtle and specialized estimands, so I'm OK with having a slightly different construction for them. (At the same time, since estimating the ATT and ATC require more care, I don't think it's worth the trouble of letting the user pass in debiasing weights for them. They are only available for causal forests with a binary treatment, and there's really no practical reason to pass in special case choices for debiasing weights -- or, at the very least, any user who might want to do so would also be sophisticated enough they could it it themselves.)

On dispatching ATE to best linear projection -- I mean, in the current implementation, we almost do it (in that all the tricky code is shared in the get score methods)! My intuition is that actually going through with the dispatch would not save us much trouble, and might also risk confusion (since ATE is a much much more common workflow than BLP, and dispatching ATE to BLP might obscure what's going on to some users).

On TMLE, yes and no :). Sorry. In the current refactor, I tried just not to touch TMLE as much as possible so as not to break it. But writing our inference framework in a way that unifies the TMLE and AIPW calls would beyond the scope of what I think we should aim for here.

On coming up with better tools for ready-to-use tables, etc.: That would be super helpful, and I think our users would like it a lot. Having `average_treatment_effect` and `best_linear_projection` return specific types of objects that know how to print themselves as a regression tables is a really nice idea! I look forward to exploring it further; we can discuss in our 1:1.

r-package/grf/man/probability_forest.Rd
```diff
@@ -111,7 +111,7 @@ X.test <- matrix(0, 10, p)
 X.test[, 1] <- seq(-1.5, 1.5, length.out = 10)
 p.hat <- predict(p.forest, X.test, estimate.variance = TRUE)
 
-# Plot the estimated success probabilities with 95 \% confidence bands.
+# Plot the estimated success probabilities with 95 % confidence bands.
```
erikcs, line 114: Why did these generated files change? (see causal survival too)   It is not necessary to escape the %  here since it will be rendered as an R comment `# ...` https://grf-labs.github.io/grf/reference/causal_survival_forest.html

r-package/grf/R/get_scores.R
```diff
@@ -0,0 +1,234 @@
...
+                          subset = NULL,
+                          debiasing.weights = NULL) {
+
+  if (!("causal_forest" %in% class(forest))) {
+    stop("The forest must be a causal_forest.")
+  }
+  subset <- validate_subset(forest, subset)
+
+  if(!all(forest$W.orig %in% c(0, 1))) {
+    stop("ATE is only implemented for a binary treatment. See APE for continuous case.")
```
erikcs, line 31: See `get_scores_APE`?

r-package/grf/R/average_treatment_effect.R
```diff
@@ -66,61 +122,138 @@
...
+
+  #
+  #
+  # From here on out, we follow a code path that is specific to causal forests.
+  #
+  #
+
+  if (!("causal_forest" %in% class(forest))) {
+    stop(paste("For any forest type other than causal_forest, the only",
+               "implemented options as method=AIPW and target.sample=all"))
```
erikcs, line 235: `options`-> `option`, `as` -> `is`

r-package/grf/R/average_treatment_effect.R
```diff
@@ -1,21 +1,38 @@
...
+#'   \item The average treatment effect on the treated (target.sample = treated): E[Y(1) - Y(0) | Wi = 1]
+#'   \item The average treatment effect on the controls (target.sample = control): : E[Y(1) - Y(0) | Wi = 0]
+#'   \item The overlap-weighted average treatment effect (target.sample = overlap):
+#'   E[e(X) (1 - e(X)) (Y(1) - Y(0))] / E[e(X) (1 - e(X)), where e(x) = P[Wi = 1 | Xi = x].
 #' }
 #' This last estimand is recommended by Li, Morgan, and Zaslavsky (JASA, 2017)
 #' in case of poor overlap (i.e., when the propensities e(x) may be very close
 #' to 0 or 1), as it doesn't involve dividing by estimated propensities.
 #'
+#' In the case of a causal forest continuous treatment, we provides estimates of the
```
erikcs, line 16: causal forest `with` continuous treatment, `provides` -> `provide`

r-package/grf/R/average_treatment_effect.R
```diff
@@ -1,21 +1,38 @@
...
+#' average partial effect, i.e., E[Cov[W, Y | X] / Var[W | X]]. In the case of a binary treatment,
+#' the average partial effect matches the average treatment effect. Computing the average partial
+#' effect is somewhat more involved, as the relevant doubly robust scores require an estimate
+#' estimate of Var[Wi | Xi = x]. By default, we get such estimates by training an auxiliary forest;
```
erikcs, line 20: `estimate` is repeated

r-package/grf/R/average_treatment_effect.R
```diff
@@ -1,21 +1,38 @@
...
+#' In the case of instrumental forests with a binary treatment, we provide an estimate
+#' of the the Average (Conditional) Local Averate Treatment (ACLATE).
+#' Specifically, given an outcome Y, treatment W and instrument Z, the (conditional) local
+#' average treatment effect is tau(x) = Cov[Y, Z | X = x] / Cov[W, Z | X = x].
+#' This is the quantity that is estimated with an instrumental forest.
+#' It can be intepreted causally in various ways. Given a homogeneity
+#' assumption, tau(x) is simply the CATE at x. When W is binary
+#' and there are no "defiers", Imbens and Angrist (1994) show that tau(x) can
+#' be interpreted as an average treatment effect on compliers. This function
+#' provides and estimate od tau = E[tau(X)]. See Chernozhukov
```
erikcs, line 32: `od` -> `of`

r-package/grf/R/average_treatment_effect.R
```diff
@@ -66,61 +122,138 @@
...
   target.sample <- match.arg(target.sample)
   method <- match.arg(method)
   cluster.se <- length(forest$clusters) > 0
 
-  if (!("causal_forest" %in% class(forest))) {
-    stop("Average effect estimation only implemented for causal_forest")
+  if (method == "TMLE") {
+    loaded <- requireNamespace("sandwich", quietly = TRUE)
+    if (!loaded) {
+      stop("To use TMLE, please install the package `sandwich`.")
```
erikcs, line 137: I don't think this message is needed anymore, `sandwich` is in the GRF package requirements (BLP does not check for this)

r-package/grf/R/average_treatment_effect.R
```diff
@@ -25,14 +42,42 @@
...
+#' @references Chernozhukov, Victor, Juan Carlos Escanciano, Hidehiko Ichimura,
+#'             Whitney K. Newey, and James M. Robins. "Locally robust semiparametric
+#'             estimation." arXiv preprint arXiv:1608.00033, 2016.
+#' @references Imbens, Guido W., and Joshua D. Angrist. "Identification and Estimation of
+#'             Local Average Treatment Effects." Econometrica 62(2), 1994.
+#' @references Robins, James M., and Andrea Rotnitzky. "Semiparametric efficiency in
+#'             multivariate regression models with missing data." Journal of the
+#'             American Statistical Association 90(429), 1995.
+#' 
+#' @importFrom stats coef lm predict var weighted.mean
```
erikcs, line 80: since `lm` is imported here it can be omitted from `best_linear_projection`'s import

r-package/grf/R/get_scores.R
```diff
@@ -0,0 +1,234 @@
...
+#'               If NULL (default) they are obtained via inverse-propensity weighting.
+#'               
+#' @references Farrell, Max H. "Robust inference on average treatment effects with
+#'             possibly more covariates than observations." Journal of Econometrics
+#'             189(1), 2015.
+#' @references Robins, James M., and Andrea Rotnitzky. "Semiparametric efficiency in
+#'             multivariate regression models with missing data." Journal of the
+#'             American Statistical Association 90(429), 1995.
+#'
+#' @export
```
erikcs, line 20: Is it necessary to export these or can they be internal? (`#' @keywords internal`)

r-package/grf/R/get_scores.R
```diff
@@ -0,0 +1,234 @@
...
+#'               are obtained via inverse-propensity weighting.
+#' @param compliance.score An estimate of the causal effect of Z on W, i.e., Delta(X) = E[W | X, Z = 1]
+#'               - E[W | X, Z = 0], which can then be used to produce debiasing.weights. If not provided,
+#'               this is estimated via an auxiliary causal forest.
+#' @param num.trees.for.weights In some cases (e.g., with causal forests with a continuous
+#'               treatment), we need to train auxiliary forests to learn debiasing weights.
+#'               This is the number of trees used for this task. Note: this argument is only
+#'               used when debiasing.weights = NULL.
+#'
+#'  @references Aronow, Peter M., and Allison Carnegie. "Beyond LATE: Estimation of the
```
erikcs, line 153: Extra space before `@references`

r-package/grf/R/get_scores.R
```diff
@@ -0,0 +1,234 @@
...
+#' @param subset Specifies subset of the training examples over which we
+#'               estimate the ATE. WARNING: For valid statistical performance,
+#'               the subset should be defined only using features Xi, not using
+#'               the treatment Wi or the outcome Yi.
+#' @param debiasing.weights A vector of length n (or the subset length) of debiasing weights.
+#'               If NULL (default) these are obtained by estimating Var[W | X = x] using a new forest.
+#' @param num.trees.for.weights Number of trees used to estimate Var[W | X = x]. Note: this
+#'               argument is only used when debiasing.weights = NULL.
+#'
+#'  @references Graham, Bryan S., and Cristine Campos de Xavier Pinto. "Semiparametrically
```
erikcs, line 71: Extra space before `@references`

erikcs (commented):
Oki, I've made a pass through now and left some small comments

1. Another minor thing that needs update is the [online reference](https://github.com/grf-labs/grf/blob/ate-refactor/r-package/grf/pkgdown/_pkgdown.yml#L90) ✔️

2. Wrt. a difference between `average_treatment_effect` and `best_linear_projection` SEs here is one edge case (which might not be worthwhile worrying about):

```
set.seed(6)
n <- 2000
p <- 10
W <- rbinom(n, 1, 1 / (1 + exp(-X[, 2]))) + rnorm(n)
tau.forest <- causal_forest(X, Y, W, clusters = sample(0:1, n, T))

> average_treatment_effect(tau.forest)
    estimate      std.err 
3.233108e-01 2.004571e-05 
> best_linear_projection(tau.forest)

Best linear projection of the conditional average treatment effect.
Confidence intervals are cluster- and heteroskedasticity-robust (HC3):

            Estimate Std. Error t value  Pr(>|t|)    
(Intercept) 0.323799   0.013916  23.269 < 2.2e-16 ***
---
Signif. codes:  0 ‘***’ 0.001 ‘**’ 0.01 ‘*’ 0.05 ‘.’ 0.1 ‘ ’ 1
```

3.  (not a big issue) There might be an inconsistency in how `average_treatment_effect` and `best_linear_projection` deals with pathological data, i.e some W.hat=1 leading to division by zero: ATE will warn and return NaN, BLP will warn and stop due to error in `lm`)

4. To avoid future github issues, we could still retain (for a while) `average_late`/`average_partial_effect` and have it return `warning/stop: this function has been deprecated in v 2.0 etc.`? ✔️

swager:
Thank you @erikcs for the detailed code review, and also for fixing everything!! Sorry I was slow to get to it.

Only one small point for discussion, on whether the `get_scores_*` should be exported. These scores may be useful to other users, and exporting them would allow others to use them. For example, the `double_robust_scores` function in `policytree` is essentially the same as `get_scores_ATE` in the case of a binary treatment. (In fact, once we have a multi-treatment causal forest available in `grf`, we may be able to eliminate `double_robust_scores` from `policytree` and just rely on relevant `get_scores_*` functions instead.)

erikcs:
Oki, I'll re-export them! But I think it is still worth considering changing them to be a generic function, like `get_scores` later on (less function names to remember for the user)?

All tests pass locally but fail on Travis due to an external issue, we can merge once that is [cleared](https://github.com/travis-ci/travis-build/commit/51f5f85ad56a6cfea94a3ce71d8ced29c26d4e45)

swager:
Sounds good! And yes, having a unified `get_scores` function that infers what to do based on, e.g., the type of the forest and whether or not the treatment is binary, could be a nice simplification to the API.

erikcs:
Some of the `average_treatment_effect` tests now give a warning (sorry, I did not notice this before now, rectified in #729):

Screen Shot 2020-08-31 at 16 13 52

because you reduced the overlap tolerance (.99 to .95): I think we should then wrap [these calls](https://github.com/grf-labs/grf/blob/ate-refactor/r-package/grf/tests/testthat/test_average_effect.R#L369) with `expect_warning`, are you OK with that?

Also the tolerance check is higher in [best_linear_projection](https://github.com/grf-labs/grf/blob/ate-refactor/r-package/grf/R/forest_summary.R#L169) - is this intentional?

erikcs:
@swager should all the summary functions be consistent and use the same thresholds for the overlap check or is it intentional that average_treatment_effect checks for <= 0.95 while the rest is at 0.99?

swager:
@erikcs yeah I changed the overlap warning threshold to 0.05 / 0.95, because this seems like a setting where overlap is already getting tenuous. Having propensity scores around 0.05 doesn't necessarily mean the analysis is doomed -- but it does suggest that caution is necessary (so a warning doesn't seem so bad).

I don't think clipping the propensity scores in our tests is necessarily a good idea. Maybe a better option would be to add an option `warn_for_poor_overlap` to the function `average_treatment_effect` that defaults to `TRUE`; and then we can set it to `FALSE` in some tests. (And maybe some users will also want to set it to `FALSE`, if they know what they're doing and don't want the function to continuously issue warnings.)

erikcs:
Oki, sounds good, but I suggest we interact with R's warning through the global `options(warn)` instead: set `options(warn = -1)` in the start of the test to ignore warnings, then restore it with `options(warn = 2)`

Are you fine with the threshold being different in `best_linear_projection` and `LATE` (Z.hat)?

erikcs:
Going ahead and merging this, we can make potential amendments later, all the important logic is locked in now!

erikcs:
Just recording this note here, we are aware that

```
n <- 500
p <- 10
W <- rbinom(n, 1, 0.5)
cf <- causal_forest(X, Y, W, clusters = sample(1:4, n, T))
c(average_treatment_effect(cf)["std.err"], best_linear_projection(cf)[2], use.names=F)
[1] 0.4245641 0.5192696
```

on clustered data with very few clusters `average_treatment_effect` and `best_linear_projection` might give slightly different standard errors. This use case is not very realistic though.

PR #726, opened 2020-08-28
Simplify two nuisance arguments in `causal_survival_forest`

erikcs:
Instead of passing:

```
S0.hat: a list of ...
S1.hat: a list of ...
```

simplify it to:

```
E0.hat: A vector of estimates of E[Y | X, W = 1].
E1.hat: "... "
```

Note that `expected_survival` is an internal function, but for power users wishing to pass this argument the computation involved is [minor](https://github.com/grf-labs/grf/blob/master/r-package/grf/R/causal_survival_forest.R#L576).


yifan-cui:
@erikcs It looks fine!

PR #728, opened 2020-08-31
Fix a pairwise comparison in old test

erikcs:
This is only a minor fix, the comparison should be pairwise (`&` not `&&`).


erikcs:
Just for reference, this line is what causes the following obscure message when running R CMD check through RStudio 
Screen Shot 2020-08-31 at 14 39 33

PR #732, opened 2020-09-02
Make `get_scores` a generic function

erikcs:
This is a purely aesthetic follow-up to #723: replace the different `get_scores` functions with a generic `get_scores` (suggest holding on renaming to `double_robust_scores` as it will cause name collision with policytree's methods, which produce slightly different outputs - matrices of DR conditional means).

Also add 
-get_scores.causal_forest
-get_scores.instrumental_forest
-get_scores.causal_survival_forest

to the online function reference (under the "Treatment effect estimation" heading)


swager:
Nice! Given this refactor, though, I think we should be able to get rid of the old `get_scores_ATE`, etc. Then, for example, `get_scores.causal_forest` would just have a big if/else check on whether treatment is binary, and proceed from there.

PR #735, opened 2020-09-14
Move `dgps.R` to R/

erikcs:
This is only a convenience - move the DGP simulation functionality to the R package.

We already have `generate_survival_data()` in `R/` which is convenient to generate test data.

Note the two functions `generate_causal_data` (causal forest) and `generate_survival_data` (causal survival forest) are exported but not added to the [online reference list](https://github.com/grf-labs/grf/blob/master/r-package/grf/pkgdown/_pkgdown.yml).

Having this under R/ gives a convenient render of the docstring, and we get some light test coverage through the `\donttest` section.


halflearned (approved):
Looks harmless enough to me.

erikcs:
Going ahead and merging this, it is easy to make future amendments now that the file is part of the R package.

PR #738, opened 2020-09-21
Add a "Visualize trees" entry to vignettes

erikcs:
This PR adds a short vignette entry `Visualize trees` giving a short overview of the tree plotting functionality in GRF. @susanathey mentioned the documentation could be clarified in this area, hopefully this vignette will be useful for users.

We demonstrate `plot` and `print` with `get_tree`, and emphasize these are only simple visualization tools, and refer the user to the `policytree` package (with an example) for tree-based treatment assignment rules.

A printout of the rendered vignette is attached [here](https://github.com/grf-labs/grf/files/5253388/out.pdf) for convenience.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -82,7 +82,7 @@ matrix:
         - cp ../../DEVELOPING.md .
         - cp ../../releases/CHANGELOG.md .
         - Rscript -e 'install.packages("pkgdown")'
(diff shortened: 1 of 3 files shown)


halflearned (approved):
Thank you for the extremely fast work on this, @erikcs! I think this vignette will be very useful to users.

PR #774, opened 2020-12-15
Update sample weights for regression/causal/instrumental forest

erikcs:
Issue: #796. This PR updates

- Regression / MultiRegression
- Causal / Instrumental forest

The components to update are

1. Point prediction
2. Variance estimates
3. Debiased error estimates

---
### Details
Attached example sketch on how some point predictions and sufficient statistics are computed, for reference:

[grf_predictions.pdf](https://github.com/grf-labs/grf/files/6255533/grf_predictions.pdf)


core/src/prediction/InstrumentalPredictionStrategy.cpp
```diff
@@ -191,13 +189,16 @@ PredictionValues InstrumentalPredictionStrategy::precompute_prediction_values(
     if (std::abs(sum_weight) <= 1e-16) {
       continue;
     }
+    std::vector<double>& value = values[i];
```
erikcs, line 192: This line is also moved to after 
```
 if (std::abs(sum_weight) <= 1e-16) {
      continue;
    }
```
which is a bug in case you supply samples weights where some of them are zero (like would be the case where using sample weights for bandwith estimation), predictions would then be zero instead of empty

core/src/prediction/InstrumentalPredictionStrategy.cpp
```diff
@@ -191,13 +189,16 @@ PredictionValues InstrumentalPredictionStrategy::precompute_prediction_values(
...
+    value.resize(NUM_TYPES);
```
erikcs, line 193: This line is also moved to after
```
 if (std::abs(sum_weight) <= 1e-16) {
      continue;
    }
```
which is a bug in case you supply samples weights where some of them are zero (like would be the case where using sample weights for bandwith estimation), predictions would then be zero instead of empty

swager (approved):
Looks great!

PR #778, opened 2020-12-23
Add Azure Pipelines

erikcs:
Replace Travis with Azure Pipelines, as in https://github.com/grf-labs/policytree/pull/68.

This pipeline does not use `build_package.R` but  instead runs all the tests through the native "R CMD check" (means that devtools etc does not need to be installed on the image). It may be convenient to retain `build_package.R` in the repo for local use - it has all the setup to produce generated files (rcpp/roxygen) if one does not use RStudio.

It is unwieldy to have `PKG_CPPFLAGS="-UNDEBUG"` activated here as before to catch Eigen mistakes, so added a note to DEVELOPING instead (R CMD check runs all the package tests, but also checks the package for violations like debug symbols in the code, and would complain here. One could enable this flag in the C++ tests, but valgrind will catch it there anyways...)

The GCC and Clang version is the default from the image setup, details are [here](https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops&tabs=yaml#software).

The default R version is release (4.0).


jtibshirani (approved):
Thanks for doing this migration! I didn't review the new build script carefully, but understand the general approach and the benefits of using Azure Pipelines.

PR #789, opened 2021-01-06
Add `vcov.type` to `test_calibration/best_linear_projection`

### LINKED ISSUE #786: test_calibration/best_linear_projection: slow with clustered SEs and large data
I am running causal_forest with 2000 trees, using data with about 2.8 million rows, 16 covariates, binary outcome and treatment. I am using a Windows 10 machine with 384 GB ram and 42 cores of i7 Intel, and loaded version grf_1.2.0. That causal_forest model takes about 10.5 h to run. To test for the presence of heterogeneity in an omnibus test, I am using the test_calibration model. However, this model has now run for over four days (96h<)—at the time of writing—without converging. 

Is this long estimation time of test_calibration considered normal compared to the causal_forest estimation time? 

If not, what advice can you give to speed up this long estimation time?

The causal_forest code block is the following 

```
tau.forest.2000.u5 <- causal_forest(df4[covariates],
                                    df4[,outcome], 
                                    df4[,treatment],  
                                    clusters = df4[,clusters],
                                    num.trees = 2000,
                                    honesty.fraction = 0.5,
                                    tune.parameters = "all")


m_cal <- test_calibration(tau.forest.2000.u5)
```

erikcs:
Closes #786. Thanks to @adeldaoud for reporting and @zeileis for the help with `sandwich`.

### DIFF
--- a/r-package/grf/R/forest_summary.R
+++ b/r-package/grf/R/forest_summary.R
@@ -15,9 +15,13 @@
 #' @param vcov.type Optional covariance type for standard errors. The possible
 #'  options are HC0, ..., HC3. The default is "HC3", which is recommended in small
 #'  samples and corresponds to the "shortcut formula" for the jackknife
-#'  (see MacKinnon & White for more discussion). For large data sets with clusters,
-#'  "HC0" or "HC1" are significantly faster to compute.
+#'  (see MacKinnon & White for more discussion, and Colin & Miller for a review).
+#'  For large data sets with clusters, "HC0" or "HC1" are significantly faster to compute.
 #' @return A heteroskedasticity-consistent test of calibration.
+#'
+#' @references Cameron, A. Colin, and Douglas L. Miller. "A practitioner's guide to
+#'  cluster-robust inference." Journal of human resources 50, no. 2 (2015): 317-372.
+#'
 #' @references Chernozhukov, Victor, Mert Demirer, Esther Duflo, and Ivan Fernandez-Val.
 #'             "Generic Machine Learning Inference on Heterogenous Treatment Effects in
 #'             Randomized Experiments." arXiv preprint arXiv:1712.04802 (2017).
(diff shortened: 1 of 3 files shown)


swager (approved):
Thanks! Just a few comments on the documentation. We should add a reference defining the HC choices:

MacKinnon, James G., and Halbert White. "Some heteroskedasticity-consistent covariance matrix estimators with improved finite sample properties." Journal of Econometrics 29.3 (1985): 305-325.

We should also mention that:
- The possible options are HC0, ..., HC4,
- The choice HC3 is recommended in small samples, and is used as default, and
- HC3 corresponds to the "shortcut formula" for the jackknife (see MacKinnon & White for more discussion).

zeileis:
Adding documentation will surely be helpful. MacKinnon & White (1985) only pertains to the cross-section case, though (implemented in `sandwich::vcovHC`). For the clustered case (implemented in `sandwich::vcovCL`), Cameron & Miller (2015) is a standard survey reference. In that case the recommendation for HC3 is not so clear-cut, hence `vcovCL` defaults to HC1 for linear models while `vcovHC` defaults to HC3. Also, there is no HC4 counterpart in the clustered case. When implementing `vcovCL` we considered this but didn't see a straightforward generalization. (And for the cross-section case there is even HC4m and HC5, both of which are less popular, though.)

Cameron AC, Miller DL (2015). “A Practitioner’s Guide to Cluster-Robust Inference.” _Journal of Human Resources_, **50**(2), 317–372. [doi:10.3368/jhr.50.2.317](https://doi.org/10.3368/jhr.50.2.317).

swager:
Good point, thanks! Regarding the default choice of HC3 -- I believe most users will be using this in the non-clustered case (in which case HC3 seems good), and my thought was that I wanted the behavior with clustering to be consistent with the behavior without clustering.

One question this brings up, though, is: When calling into `coeftest`, should we add a check on whether there is clustering or not (i.e., whether there exist duplicate elements in the clusters vector), and then use `vcov = sandwich::vcovHC` instead of `vcov = sandwich::vcovCL` in the case there is no clustering? I'd be curious if you have any recommendations about that, @zeileis.

zeileis:
I agree that HC3 is a reasonable choice here, just wanted to point out that the default for `vcovCL` differs.

And when each observation is its own cluster then `vcovCL` yields the same results as `vcovHC`, e.g.,

```
m <- lm(dist ~ speed, data = cars)
library("sandwich")
sqrt(diag(vcovHC(m, type = "HC3")))
## (Intercept)       speed 
##   5.9318033   0.4275372 
sqrt(diag(vcovCL(m, type = "HC3", cluster = 1:50)))
## (Intercept)       speed 
##   5.9318033   0.4275372 
sqrt(diag(vcovCL(m, type = "HC3")))
## (Intercept)       speed 
##   5.9318033   0.4275372 
```

There may be small performance differences but we have special-cased the common `cluster = 1:n` scenario within `vcovCL` so that unnecessary computations are avoided.

swager:
Makes sense, thanks!

erikcs:
Do you two think this is OK to merge now or should we add the Cameron reference too?

swager:
Yes let's add Cameron & Miller; beyond that, looks good to me.

erikcs:
Just tacking on a note for the future @grf-labs: all names should be chosen meticulously.

In hindsight I don't see why we didn't name this option `vcov` instead of `vcov.type`, to be consistent with @zeileis interface.

Another old naming "glitch" in GRF is the argument `clusters`: `cluster` is used in `sandwhich` and almost everywhere else.

zeileis:
I had noticed that you named `vcov.type` differently but thought that this was intentional, but I overlooked `clusters`. In any case, the good thing about both names is that partial matching will work for both `vcov = ...` and `cluster = ...`.

Issue #758, opened 2020-11-19
Remove "tune_regression_forest" and "tune_causal_forest"


erikcs:
The upcoming 2.0 release allows for making breaking changes and to trim fat. This additional tuning interface adds code complexity and maintenance burden, especially when adding new forests.

The main entry point should cover the necessary use cases: `tuned.forest = causal_forest(X, Y, W, tune.parameters = "all")` returns a tuned forest with the tuning result and details accessible through `tuned.forest$tuning.output`.  (note: add to _deprecated.R_)

Discussed and agreed by @halflearned , @jtibshirani and @davidahirshberg !

PR #790, opened 2021-01-07
Deprecate `tune_***_forest`

erikcs:
See #758. 

We remove the optional entry points:

-`tune_regression_forest`
-`tune_causal_forest`
-`tune_instrumental_forest`

As well as making these internal:

-`tune_ll_regression_forest`
-`tune_ll_causal_forest`

@rinafriedberg: Do you feel strongly about keeping `tune_ll_regression_forest` and `tune_ll_causal_forest` `@exported` or can they be made internal? They  kind of fit the same bill as above, you get the LLF parameter tuned by default when doing `predict`.


rinafriedberg:
@erikcs I don't feel strongly about keeping them external, thanks for checking :)

erikcs:
Thanks @rinafriedberg, `@ internal`'d them!

erikcs:
You OK with this @halflearned?

halflearned (approved):
No issues on my end! 

I think we discussed this in a previous meeting and got consensus that it was a good idea, so I'll go ahead and approve this PR. Thanks @erikcs!

PR #797, opened 2021-01-21
Update some references

erikcs:
Adding info for the published version of @swager and @rinafriedberg's papers. The reference for the "best_linear_projection" paper has changed title, so updated that as well. Also replaced "forthcoming" with year in the missing values paper.


swager (approved):
Looks good, thanks!

PR #815, opened 2021-02-12
Add multiple outcomes Y to `multi_arm_causal_forest`

erikcs:
This PR adds support for matrix `Y` input to `multi_arm_causal_forest(X, Y, W)`.

API: 

let there be M outcomes and K treatments (Y is `n*M`).  `predict` will then return a 3-d array of size `n*(K-1)*M` (there are K-1 contrasts). `average_treatment_effect` gets an extra argument `outcome`, an integer specifying the outcome column you want an ATE for.

(a characterization test update is in principal not needed for this addition, but tiny numerical differences required it, it is explained in the commits below)


jtibshirani (approved):
This makes sense to me!

PR #820, opened 2021-02-26
Rename "deprecated->removed" in warning message

erikcs:


r-package/grf/R/deprecated.R
```diff
@@ -9,7 +9,7 @@
 #'
 #' @export
 average_late <- function(forest, ...) {
-  stop("This function has been deprecated after version 1.2.0. See the function `average_treatment_effect` instead.")
+  stop("This function has been removed after version 1.2.0. See the function `average_treatment_effect` instead.")
```
jtibshirani, line 12: Maybe we could update the title in the comment to "Average LATE (removed)"

jtibshirani (approved):

r-package/grf/R/deprecated.R
erikcs: Oki

PR #841, opened 2021-03-18
Disable sample.weights for local linear forests

erikcs:
As agreed with by @swager, disable sample.weights for local linear forests.

Currently, they are supported, but in inconsistent ways: local linear regression takes them into account during splitting but not elsewhere, same for local linear causal predictions.

Instead of implementing all this before GRF 2.0, disable this for now, pending a potential future feature.


rinafriedberg (approved):
Hey - this looks great. Thanks so much Erik!

PR #846, opened 2021-03-24
Add e*tau to NW dgps

erikcs:
The way all these DGPs are parameterized: needs a `e*tau` term in `m`. Missed in #734.


halflearned (approved):
Thanks, @erikcs. Note that there might also some issues with the normalization -- I'll write an issue about it later.

PR #848, opened 2021-03-25
Simplify setting the dimension of `responses_by_sample`

erikcs:
This is just an aesthetic suggestion to simplify how the dimension of `responses_by_sample` is handled.

Recall `responses_by_sample(i, )` is typically length 1 for most forests, but with the introduction of a vector valued splitting rule, it may be longer, like `num_outcomes * num_treatments` for the `multi_arm_causal_forest`. 

Now TreeTrainer initializes this array with the general and more future friendly

```C++
Eigen::ArrayXXd responses_by_sample(data.get_num_rows(), relabeling_strategy->get_response_length());
```

instead of

```C++
Eigen::ArrayXXd responses_by_sample(data.get_num_rows(), data.get_num_outcomes() * data.get_num_treatments());
```

with `get_response_length()` delegated to the `RelabelingStrategy` (default value of 1, overriden by forests which wants to)

### DIFF
--- a/core/src/forest/ForestTrainers.cpp
+++ b/core/src/forest/ForestTrainers.cpp
@@ -54,11 +54,12 @@ ForestTrainer instrumental_trainer(double reduced_form_weight,
 }
 
 ForestTrainer multi_causal_trainer(size_t num_treatments,
-                                   size_t num_outcomes) {
+                                   size_t num_outcomes,
+                                   size_t response_length) {
 
-  std::unique_ptr<RelabelingStrategy> relabeling_strategy(new MultiCausalRelabelingStrategy());
+  std::unique_ptr<RelabelingStrategy> relabeling_strategy(new MultiCausalRelabelingStrategy(response_length));
   std::unique_ptr<SplittingRuleFactory> splitting_rule_factory =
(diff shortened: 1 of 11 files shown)


core/src/forest/ForestTrainers.cpp
```diff
@@ -54,11 +54,12 @@ ForestTrainer instrumental_trainer(double reduced_form_weight,
 }
 
 ForestTrainer multi_causal_trainer(size_t num_treatments,
-                                   size_t num_outcomes) {
+                                   size_t num_outcomes,
+                                   size_t response_length) {
```
jtibshirani, line 58: Instead of adding a new parameter, could this method just compute `response_length = num_treatments * num_outcomes` internally?

jtibshirani (approved):

core/src/forest/ForestTrainers.cpp
erikcs: Yes, but I left it slightly redundant like this for now, as this allows us (Stefan!) to more easily experiment with more funky stuff later, like splitting on an intercept, gives response_length = num_treatments * num_outcomes + 1, stuff like that :P

erikcs: Going ahead with the merge - assuming you are fine with this minor redundancy for the sake of future experimentation ease!

PR #876, opened 2021-04-05
Remove defunct PredictionValues.clear()

erikcs:
Minor aesthetic: After #519 this method is not used anymore as there are no destructive iterations over forest objects.

### DIFF
--- a/r-package/grf/bindings/RcppUtilities.h
+++ b/r-package/grf/bindings/RcppUtilities.h
@@ -14,6 +14,10 @@ public:
    * Converts the provided {@link Forest} object and OOB predictions to an R list
    * to be returned through the Rcpp bindings. The provided predictions vector can
    * be present if OOB predictions were not requested as part of training.
+   *
+   * NOTE: To conserve memory, this method destructively modifies the forest
+   * object by clearing out individual {@link Tree} objects. The forest cannot
+   * be used once it has been passed to the method.
    */
   static Rcpp::List create_forest_object(Forest& forest,
                                          const std::vector<Prediction>& predictions);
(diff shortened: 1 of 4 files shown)


r-package/grf/bindings/RcppUtilities.h
```diff
@@ -14,10 +14,6 @@ class RcppUtilities {
    * Converts the provided {@link Forest} object and OOB predictions to an R list
    * to be returned through the Rcpp bindings. The provided predictions vector can
    * be present if OOB predictions were not requested as part of training.
-   *
-   * NOTE: To conserve memory, this method destructively modifies the forest
```
jtibshirani, line 18: My memory is not totally clear, but I think this note still applies since `RcppUtilities::serialize_forest` can destructively modify the forest?

erikcs: Ah, yes I was taking this note too literally, there is no "explicit" clearing being done,

https://github.com/grf-labs/grf/blob/master/r-package/grf/bindings/RcppUtilities.cpp#L55

BUT it loops over by fetching a unique_ptr to the invividual trees, since the pointer is unique the object dies after it goes out of scope and the original forest is empty.

Yes, that makes sense, that note is technically correct - I was just confused by the semantic.

I can add it back and note that it _implicitly_ clears out the forest since it moves all the unique tree pointers - or do you think that is too obvious?

erikcs: (can restore the original docstring comment and just a note on this in the code moving the tree ptr?)

jtibshirani (approved):

PR #890, opened 2021-04-08
Update stabilized split section in REFERENCE

erikcs:
Just some minor clarifications, as @susanne-207 pointed out this section is not exactly crystal clear.

* The header name was outdated -> "Selecting Balanced Splits"
* alpha <= 0.25 is not enforced anywhere in the code
* clarify the sentence refers to causal splitting 
* clarify this refers to the stabilize.splits=TRUE argument

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -124,13 +124,13 @@ The parameter `min.node.size` relates to the minimum size a leaf node is allowed
 There are several important caveats to this parameter:
 - When honesty is enabled, the leaf nodes are 'repopulated' after splitting with a fresh subsample. This means that the final tree may contain leaf nodes smaller than the `min.node.size` setting.
 - For regression forests, the splitting will only stop once a node has become smaller than `min.node.size`. Because of this, trees can have leaf nodes that violate the `min.node.size` setting. We initially chose this behavior to match that of other random forest packages like `randomForest` and `ranger`, but will likely be changed as it is misleading (see [#143](https://github.com/grf-labs/grf/issues/143)).
-- When training a causal forest, `min.node.size` takes on a slightly different notion related to the number of treatment and control samples. More detail can be found in the 'Split Penalization' section below, under the 'Causal Forests' heading.
+- When training a causal forest, `min.node.size` takes on a slightly different notion related to the number of treatment and control samples. More detail can be found in the 'Selecting Balanced Splits' section below, under the 'Causal Forests' heading.
 
 #### `alpha`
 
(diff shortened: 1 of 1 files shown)


REFERENCE.md
```diff
@@ -180,9 +180,9 @@ Empirically, we've found orthogonalization to be essential in obtaining accurate
 
 In the sections above on `min.node.size`, `alpha`, and `imbalance.penalty`, we described how tree training protects against making splits that result in a large size imbalance between the children, or leaf nodes that are too small. In a causal setting, it is not sufficient to consider the number of examples in each node -- we must also take into account the number treatment vs. control examples. Without a reasonable balance of treated and control examples, there will not be enough information in the node to obtain a good estimate of treatment effect. In the worst case, we could end up with nodes composed entirely of control (or treatment) examples.
 
-For this reason, causal splitting uses modified notions of each split balancing parameter:
+For this reason, causal splitting (enabled with the argument `stabilize.splits = TRUE`) uses modified notions of each split balancing parameter:
```
jtibshirani, line 183: I find it a little confusing to say "enabled with the argument `stabilize.splits = TRUE`" because it is already enabled by default. We actually don't expect users to disable it, as it's proved important to performance, so I'm not sure we need to call out the argument at all.

REFERENCE.md
```diff
@@ -124,13 +124,13 @@ The parameter `min.node.size` relates to the minimum size a leaf node is allowed
...
-- When training a causal forest, `min.node.size` takes on a slightly different notion related to the number of treatment and control samples. More detail can be found in the 'Split Penalization' section below, under the 'Causal Forests' heading.
+- When training a causal forest, `min.node.size` takes on a slightly different notion related to the number of treatment and control samples. More detail can be found in the 'Selecting Balanced Splits' section below, under the 'Causal Forests' heading.
 
 #### `alpha`
 
-The parameter `alpha` controls the maximum imbalance of a split. In particular, when splitting a parent node, the size of each child node is not allowed to be less than `size(parent) * alpha`. Its value must lie between (0, 0.25), and defaults to 0.05.
+The parameter `alpha` controls the maximum imbalance of a split. In particular, when splitting a parent node, the size of each child node is not allowed to be less than `size(parent) * alpha`. It defaults to 0.05.
```
jtibshirani, line 131: Instead of just removing the requirement from the docs, should we make sure it's enforced? It looks like when tuning forests, we make sure alpha lies in [0, 0.25). Some values like alpha=0.7 really don't make sense.

PR #893, opened 2021-04-09
Update tuning call in `ll_regression_forest`

erikcs:
Overlooked in #790: tuning now directly calls into the internal `tune_forest`, update this.

To avoid code bloat/making complicated code even more complicated (for the sake of future maintainability), tuning is disabled when `enable.ll.splits = TRUE`. Let me know if you are OK with this @rinafriedberg.


rinafriedberg (approved):
Looks good to me!

PR #924, opened 2021-04-23
Add optional weight vector to MultiCausalRelabeling

erikcs:
Add optional weight vector  `influence_weight` (default equal to [1, ..., 1], size = num.treatments * num.outcomes) which multiplies each element of the relabeled output by this weight.

Is currently non-functional (defaults to ones) for `multi_arm_causal_forest`, but is implemented as an optional binding/C++ feature here allowing for example a future RDD forest to easily re-use the existing R wrapper and supply an optional weight vector ignoring the running-variable influence (= no need to set up a new forest trainer from the C++ side to use MultiCausalRelabeling with single-variable RegressionSplitting).


erikcs:
Closing for now and re-opening when needed for a concrete application.

Issue #747, opened 2020-10-23
Consider alternatives to virtual function getter in Data


erikcs:
`Data::get` is virtual to allow for different underlying storage (dense and sparse). This gives a modest performance hit since `get` is very hot function that benefits from compile-time optimizations. 

1. An alternative way to bake this in at compile time is with the [curiously recurring template pattern](https://en.wikipedia.org/wiki/Curiously_recurring_template_pattern). However, this adds code complexity as the `Data` passed to all forest components now has to be parameterized according to the underlying storage type `Data<T>`.

2. Another tangential item for this issue is other storage options for the multivariate outcomes/treatments that would allow `get_outcomes` to just return a view of outcome row j. `Data` could hold a separate `Eigen::Array` (row major) for only outcomes for example (is an empirical question whether this is a bottleneck).

jtibshirani:
Adding an idea we discussed: could we just remove support for sparse matrices in order to make `Data` a concrete class?

erikcs:
Another tangential item for this issue is other storage options for the multivariate outcomes/treatments that would allow `get_outcomes` to just return a view of outcome row j. `Data` could hold a separate `Eigen::Array` (row major) for only outcomes for example (is an empirical question whether this is a bottleneck).

erikcs:
> Adding an idea we discussed: could we just remove support for sparse matrices in order to make `Data` a concrete class?

That would be convenient. 

In GRF forest size >>> training data size X, so the savings from a compressed X seems almost negligible, unless X is gigantic, in which case GRF would have speed issues anyways.

erikcs:
Not relevant anymore after #946.

PR #921, opened 2021-04-19
Add stabilized splits for multi_arm_causal_forest

erikcs:
Add a multivariate extension of `InstrumentalSplittingRule` -> `MultiCausalSplittingRule` (using same variable names, same code flow, etc. You can look at them in split screen and see they do exactly the same) that applies the same constraints to each treatment arm. Enabled by default in `multi_arm_causal_forest` with `stabilize.splits = TRUE`.

A check of the regret simulations from policytree:
```
Multi action causal forest - stabilize.splits = TRUE
#      n        x.Min.     x.1st Qu.      x.Median        x.Mean     x.3rd Qu.        x.Max.
# 1 1000  0.0007333333  0.0724000000  0.1579000000  0.1637372500  0.2076083333  0.8226333333
# 2 1500  0.0034333333  0.0248750000  0.0563666667  0.0817366667  0.1464166667  0.3499333333
# 3 2000 -0.0001666667  0.0175750000  0.0304666667  0.0546278333  0.0788000000  0.2746333333
# 4 2500 -0.0005333333  0.0113750000  0.0242666667  0.0417414167  0.0487750000  0.2077333333

Multi action causal forest - stabilize.splits = FALSE
#      n        x.Min.     x.1st Qu.      x.Median        x.Mean     x.3rd Qu.        x.Max.
# 1 1000  0.0023000000  0.0555583333  0.1535833333  0.1482156667  0.2002250000  0.6932333333
# 2 1500 -0.0003666667  0.0307250000  0.0683000000  0.0919045000  0.1510833333  0.3494000000
# 3 2000  0.0002000000  0.0164250000  0.0370000000  0.0616894167  0.0909750000  0.2416000000
# 4 2500 -0.0002333333  0.0125750000  0.0251000000  0.0398609167  0.0479666667  0.1763333333

policytree's one-vs-all multi causal forest
#      n        x.Min.     x.1st Qu.      x.Median        x.Mean     x.3rd Qu.        x.Max.
# 5 1000  0.0066666667  0.0975833333  0.1288833333  0.1489028333  0.1830833333  0.7165333333
# 6 1500  0.0010666667  0.0370000000  0.1018333333  0.0952007500  0.1352833333  0.3595000000
# 7 2000  0.0007666667  0.0202583333  0.0487333333  0.0703962500  0.1187416667  0.3541000000
# 8 2500  0.0001333333  0.0140583333  0.0314333333  0.0499800833  0.0881250000  0.2044666667
```

PR #930, opened 2021-04-28
Store outcomes/treatments in separate data matrix

erikcs:
A simple way (at the cost of slight redundancy) of addressing 2. in #747. 

Will make a speed difference for multivariate stabilized splits where there are many calls to `get_treatments`.


erikcs:
Closing this for now, #921 can store the result of `get_treatments` in a temporary for re-use and obtain the same speedup, obviating the need for this Data modification.

PR #994, opened 2021-06-13
Prepare the 2.0 release

erikcs:


r-package/grf/DESCRIPTION
```diff
@@ -1,6 +1,6 @@
 Package: grf
 Title: Generalized Random Forests
-Version: 1.2.0.0
+Version: 2.0
```
jtibshirani, line 3: Should this be 2.0.0.0 (not sure if CRAN requires this)?

releases/CHANGELOG.md
```diff
@@ -4,6 +4,36 @@ All notable changes to `grf` will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [2.0] - 2021-?-?
```
jtibshirani, line 7: Let's update this to 2.0.0 here and in tarball name in case we need a patch release (like we did for 1.0.0 :))

jtibshirani (approved):
Looks good to me! I know we're waiting on https://github.com/grf-labs/grf/pull/988 and also CRAN acceptance before merging.

r-package/grf/DESCRIPTION
erikcs: Thanks, yes it should be `2.0.0`! the final digit was just something we tacked onto the github version so we could easily distinguish the development version from the CRAN version, since this development version lingered around for a year

Issue #292, opened 2018-09-05
Provide optional printout of progress


tibshirani:
I ran a  regression_forest for > 10 minutes and had no idea if it would complete in 15 min or an hour.

It would be great to have an argument "verbose" (default FALSE)  which causes the function to
print the function's progress,  to  help the user estimate the remaining time before completion.

ginward:
This feature looks important to me, too. Especially when training the forest on large datasets.

ginward:
It will be nice if it could be printed out that how many trees have been grown v.s. how many trees are left to grow.

erikcs:
For reference ranger does this here

https://github.com/imbs-hl/ranger/blob/master/src/Forest.cpp#L1072

PR #1025, opened 2021-07-29
Provide option for progress estimation when training a forest

patrickvossler18:
This pull request is a first attempt at addressing #292. For now, the progress percentage and time estimation are only implemented for training forests.

We create a `ProgressBar` class and it is initialized in the `ForestTrainer::train_trees` function. The progress bar is then passed to the `train_batch` function, and the progress estimate is updated once each of the trees has been grown.

`ProgressBar::write_time_estimate` uses the same logic as the ranger `showProgress` function of waiting a pre-specified amount of time (`STATUS_INTERVAL` in `globals.h`) before printing out the estimated time remaining.  Also, I added the `beautifyTime` function from ranger to `utility.cpp`. If there is a proper way to give attribution for using their function, please let me know.

At the moment, we show progress for any forests grown for orthogonalizing or other substeps of training a forest. But, should the time estimation only be shown for the final forest? In this first version, `ForestTrainer` has a different `verbose_operation_name` to clarify whether we are giving a time estimate for a substep like orthogonalization or training the final forest.

Remaining to-dos:
- [ ] Write tests?
- [ ] Decide whether to keep the estimation for substeps
- [ ] Improve formatting of printed output

EDIT:
~~I forgot that CRAN does not permit the use of `std::cout` in R packages so we either need a way to successfully put the `Rcout` print statement in a locking scope or print at some other point in the training loop.~~

~~I've tried just using a mutex around the [line](https://github.com/grf-labs/grf/pull/1025/files#diff-8d915af053729f388069502f3344f77c78c02438fbe92325cebacdfc8436cbdbR44) where we can replace std::cout with Rcout, but I still get a C stack usage error.  I've been looking into `RcppThread`'s `Rcout` function and it looks like it collects the print messages in a global buffer and only prints once it is called again from the main thread.~~

See comments below.

### DIFF
--- a/core/.gitignore
+++ b/core/.gitignore
@@ -13,3 +13,4 @@
 .settings/*
 .project
 .cproject
+cmake-build-debug/
--- a/core/src/commons/ProgressBar.cpp
+++ b/core/src/commons/ProgressBar.cpp
@@ -1,0 +1,9 @@
+
+#include "ProgressBar.h"
+
+ProgressBar::ProgressBar(std::string op_str, size_t max_prog, bool v) {
+    operation = std::move(op_str);
+    max_progress = max_prog;
+    verbose = v;
+}
+
--- a/core/src/commons/ProgressBar.h
+++ b/core/src/commons/ProgressBar.h
@@ -1,0 +1,125 @@
+#ifndef GRF_PROGRESSBAR_H
+#define GRF_PROGRESSBAR_H
+#include <atomic>
+#include <mutex>
+#include <iostream>
+#include <utility>
+#include "commons/utility.h"
+#include "commons/globals.h"
+
+
+using std::chrono::steady_clock;
+using std::chrono::duration_cast;
+using std::chrono::seconds;
+
+
+
+class ProgressBar {
+
+    public:
+        ProgressBar (std::string, size_t, bool);
+
+        void set_progress(float value_pct, size_t value_int) {
+            std::unique_lock<std::mutex> lock{mutex_};
+            progress_pct_ = value_pct;
+            progress_ += value_int;
+        }
+
+        void set_initial_times() {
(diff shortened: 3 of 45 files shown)


swager:
Thanks for putting this together @patrickvossler18! This has long been a popular user request, so we're excited to see progress here. We'll review this, and get back to you soon with more.

patrickvossler18:
I think my locking issue has to do with the mutex not being shared amongst the different threads. If I try to add a mutex for the ForestTrainer class, I get errors about "call to implicitly-deleted copy constructor of 'grf::ForestTrainer'". This is reaching the limit of my C++ knowledge, but I think this suggests that we need to redesign the `ProgressBar` class and maybe just include it in `ForestTrainer`. If anyone has suggestions for how to get around this issue, I'm happy to try to implement them.

patrickvossler18:
I realized I shouldn't be trying to print from the threads. I've updated the code to mirror the logic used in the ranger package.

patrickvossler18:
Hi, it looks like checks are failing because I am including the Rcpp header incorrectly. Does anyone know how I can fix this? Should the progress estimation files be moved to the `r-package` folder?

jtibshirani:
Hi @patrickvossler18, I'm sorry for misleading you with the 'good first issue' label. After digging in and seeing your PR, I realize that there are some real challenges to sort out. As @erikcs mentioned, a proper implementation would need to account for the fact that the logging can be called from different contexts (running from C++, R, or even Python).

I suggest we close this out. We can revive it when @erikcs or I have time to think through details and suggest a path forward. Although we aren't merging it, your PR was really helpful -- we now understand what's required and what the main challenges are.

patrickvossler18:
That makes sense and I'd be happy to help any way I can in the future!

PR #1067, opened 2021-11-17
Allow some inadmissible draws in tuning

### LINKED ISSUE #1066: Continue tuning in case some forest errors are NA
[this](https://github.com/grf-labs/grf/blob/master/r-package/grf/R/tune_forest.R#L48) condition is too strict, if just some estimates are NA, which can be the case for some parameter draws, then tuning can still continue on the non-NAs, as for example the case here, where some estimates are NA because the drawn sample.fraction happen to be so low just a single cluster is drawn:

```R
n <- 500
p <- 10
W <- rbinom(n, 1, 0.5)
cl <- sample(1:5, n, T)
cf <- causal_forest(X, Y, W, clusters = cl, tune.parameters = "sample.fraction")
# Warning messages:
#   1: In tune_forest(data = data, nrow.X = nrow(X), ncol.X = ncol(X),  :
#                       Could not tune forest because some small forest error estimates were NA.
```

let me know if you agree @halflearned

erikcs:
Continue tuning procedure even if some draws are inadmissible. Closes #1066.

### DIFF
--- a/r-package/grf/R/tune_forest.R
+++ b/r-package/grf/R/tune_forest.R
@@ -44,13 +44,19 @@ tune_forest <- function(data,
     error <- small.forest$debiased.error
     mean(error, na.rm = TRUE)
   })
+  # Some estimates may be all NaN because of arbitrary inadmissible parameter draws,
+  # such as a very low sample.fraction on data with few clusters.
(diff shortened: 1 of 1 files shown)


halflearned (approved):
Approved, thank you!

PR #1070, opened 2021-11-23
Use parent forest seed in auxiliary forest

### LINKED ISSUE #1068: Consider using parent forest seed in auxiliary forest
Some `average_treatment_effect` estimates rely on additional weights that are by default estimated using an auxiliary forest. This may be confusing to some users:

```
n <- 500
p <- 10
W <- rnorm(n, 1, 0.5)
c.forest <- causal_forest(X, Y, W, seed=2)
> average_treatment_effect(c.forest)
 estimate   std.err 
0.5332316 0.1056678 
> average_treatment_effect(c.forest)
 estimate   std.err 
0.5392227 0.1065802 
```

@halflearned and I  would say it is reasonable for a user to expect that the `seed` argument to c.forest is respected by `average_treatment_effect`. See #1032 and #1063 for users pointing this out.

We could have each forest store its `seed` attribute and by default use this in the auxiliary forest training. This avoids a user having to do a global `set.seed` or pass in their own debiasing weights (where they copy the code construction from the source), and is more consistent with how tuning now behaves #704.

Double checking if you are fine with this @swager as we briefly talked about this scenario back when doing the ATE refactor: https://github.com/grf-labs/grf/pull/615#discussion_r380324198

erikcs:
Closes #1068.

### DIFF
--- a/r-package/grf/R/causal_forest.R
+++ b/r-package/grf/R/causal_forest.R
@@ -260,6 +260,7 @@ causal_forest <- function(X, Y, W,
 
   forest <- do.call.rcpp(causal_train, c(data, args))
   class(forest) <- c("causal_forest", "grf")
+  forest[["seed"]] <- seed
   forest[["ci.group.size"]] <- ci.group.size
   forest[["X.orig"]] <- X
   forest[["Y.orig"]] <- Y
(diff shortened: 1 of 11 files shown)


halflearned (approved):
Looks like a clean PR. Things seem to be working. Thanks @erikcs!

swager (approved):

PR #1079, opened 2021-12-03
Remove an unnecessary transpose

erikcs:
This matrix is symmetric and the transpose is redundant here.


erikcs:
Leaving as is, in some cases numerical differences in A-A' on the order of 1e-16 is enough to trigger a different split.

PR #1086, opened 2022-01-16
Add rank_average_treatment_effect

erikcs:
Add `rank_average_treatment_effect`. 

```R
n <- 1500
p <- 5
W <- rbinom(n, 1, 0.5)
train <- sample(1:n, n / 2)
cf.priority <- causal_forest(X[train, ], Y[train], W[train])
cf.eval <- causal_forest(X[-train, ], Y[-train], W[-train])

priority.cate <- predict(cf.priority, X[-train, ])$predictions
priority.rand <- runif(n)[-train]

rate <- rank_average_treatment_effect(cf.eval, cbind(priority.cate, priority.rand), target = "AUTOC")
rate
# estimate    std.err                                target
# 0.5824841 0.09106396                 priority.cate | AUTOC
# 0.2661738 0.07895665                 priority.rand | AUTOC
# 0.3163103 0.12065272 priority.cate - priority.rand | AUTOC
```

(test with `devtools::install_github("erikcs/grf", subdir = "r-package/grf", ref = "RATE")`)


experiments/README.md
```diff
@@ -14,6 +14,8 @@ This directory contains replication code for
 
 * Wager and Athey (2018): This paper is not based on GRF, but on the deprecated `causalForest`. For replication code see https://github.com/swager/causalForest
 
+* Yadlowsky, Fleming, Shah, Brunskill, and Wager (2021): The method is available in the GRF function `rank_average_treatment_effect`. For replication code see https://github.com/som-shahlab/risk-vs-hte
```
scottfleming, line 17: I should create a new, clean, and targeted repo with just the code that went into the final experiments that are used in the paper. I'll take that as a to-do on my end. Maybe https://github.com/som-shahlab/rank-weighted-average-treatment-effects-paper?

r-package/grf/R/rank_average_treatment.R
```diff
@@ -0,0 +1,297 @@
...
+    num.ties <- num.ties[num.ties != 0] # ignore potential ranks not present in BS sample
+    grp.sum <- rowsum(data[indices, 1:2][idx, ], prio[idx], reorder = FALSE)
+    DR.avg <- grp.sum[, 1] / grp.sum[, 2]
+
+    DR.scores.sorted <- rep.int(DR.avg, rev(num.ties))
+    DR.scores.grid <- rowsum(cbind(DR.scores.sorted * sample.weights, sample.weights), grid.id, reorder = FALSE)
+    ATE <- sum(DR.scores.sorted * sample.weights) / sum(sample.weights)
+
+    TOC <- cumsum(DR.scores.sorted * sample.weights) / cumsum(sample.weights) - ATE
+    TOC.grid <- cumsum(DR.scores.grid[, 1]) / cumsum(DR.scores.grid[, 2]) - ATE
```
scottfleming, line 212: So this line is potentially problematic. I'll follow up via email as well to discuss the math, but basically with sample weights it gets tricky to think about TOC(_u_) for some arbitrary _u_ because when you randomly permute the samples that have tied priorities, if _u_ falls within that set of ties for the weighted empirical CDF, you may get a situation where under some permutations _k_ of the tied samples fit under _u_ but under others only _k_' fit.

experiments/README.md
erikcs: Sounds good, let me know what repo name you want, https://github.com/som-shahlab/rate-paper sounds like simpler url

r-package/grf/R/rank_average_treatment.R
erikcs: Got it, the grid calculations have been revised to allow for arbitrary query points q using fractional samples

scottfleming (commented):
This all looks great! One last thing before approving - I know there was talk of baking in the hypothesis testing machinery into the RATE. It's easy enough to implement, I can share the code I had to do it or bake into a pull request directly. Did we want to do that and, if so, as part of this pull request?

erikcs:
> This all looks great! One last thing before approving - I know there was talk of baking in the hypothesis testing machinery into the RATE. It's easy enough to implement, I can share the code I had to do it or bake into a pull request directly. Did we want to do that and, if so, as part of this pull request?

The user have everything they need to compute a 95% CI or p-value from `estimate` and `std.err`, just like in `average_treatment_effect`. But maybe it's isn't obvious for a novice that estimate +/- 1.96*std.err is a valid CI if they don't know RATE satisfies a CLT, so a docstring note in the example could maybe be useful?

```
# Construct a valid 95 \% CI
rate.diff$estimate + data.frame(lower=-1.96 * rate.diff$std.err, upper=1.96 * rate.diff$std.err)
```

erikcs:
Btw, if a user wants to compute more than one contrast and get a data.frame of estimates they can do something like:
```R
p1=priority
p2=priority.rand
p3=-priority
p4=p1+2*rnorm(length(p1))

r1=rank_average_treatment_effect(cf.eval, cbind(p1, p2))
r2=rank_average_treatment_effect(cf.eval, cbind(p1, p3))
r3=rank_average_treatment_effect(cf.eval, cbind(p1, p4))

df = do.call(rbind, lapply(list(r1,r2,r3), function(r) data.frame(r[-4])))
df = df[!duplicated(df$target), ]
df
#      estimate    std.err          target
# 1  0.54611358 0.07859955      p1 | AUTOC
# 2 -0.09116870 0.08559098      p2 | AUTOC
# 3  0.63728228 0.12222670 p1 - p2 | AUTOC
# 5 -0.31854153 0.07039931      p3 | AUTOC
# 6  0.86465511 0.13325744 p1 - p3 | AUTOC
# 8  0.49297409 0.08488581      p4 | AUTOC
# 9  0.05313949 0.06218908 p1 - p4 | AUTOC
```
The design idea was having RATE return a non-opaque dead-simple object with all public attributes which the user together with rank_average_treatment_effect can combine with common R idioms to express/perform most reasonable tasks

syadlowsky:
For the plotting, it would be good if there's a way to set the CI-specific parameters. Right now, `lty` is hard-coded as `2`. Perhaps a `ci.lty` argument and `ate.lty` argument for the `abline` parameter below it?

erikcs:
sounds reasonable @syadlowsky , the remaining plot args are added:
`plot(rate.diff, xlab = "treated frac.", legend.args = list(legend=c("cates", "RaNdoM")), abline.args = list(lty = 1), ci.args = list(lty = 3))`

Screen Shot 2022-02-10 at 15 12 50

r-package/grf/R/rank_average_treatment.R
```diff
@@ -0,0 +1,310 @@
...
+    n.bs <- floor(n / 2)
+    index.list <- replicate(R, unlist(samples.by.cluster[sample.int(n, n.bs, replace = FALSE)], use.names = FALSE), simplify = FALSE)
+  } else {
+    index.list <- replicate(R, unlist(samples.by.cluster[sample.int(n, replace = TRUE)], use.names = FALSE), simplify = FALSE)
+  }
+
+  t0 <- statistic(data, seq_len(NROW(data)), ...)
+  t0 <- matrix(unlist(t0), ncol = length(t0))
+
+  res <- lapply(seq_len(R), function(i) statistic(data, index.list[[i]], ...))
```
erikcs, line 303: This line can be parallelized with a drop-in replacement: `parallell::mclapply` instead of `lapply`

I am fine keeping it serial though, a quick bench shows it's reasonably fast, for ex n=1e6, continuous scoring and R=200 bs replications takes around 50-60 seconds, while mclapply with 24 cores barely bumps it down to ~25-30 seconds

I don't think that speedup is worth the hazzle of having to deal with `parallell` in the GRF package (even if it defaults to 1 core): it doesn't work on Windows, it relies on forking, which may cause problems in Rstudio, etc

r-package/grf/R/rank_average_treatment.R
```diff
@@ -0,0 +1,310 @@
...
+  } else if (target == "QINI") {
+    wtd.mean <- function(x, w) sum(cumsum(w) / sum(w) * w * x) / sum(w)
+  }
+
+  # Compute estimates, a function to be passed on to bootstrap routine.
+  # @data: a data.frame with the original data, column 1: DR.scores*sample.weights, column 2: sample.weights,
+  #   column 3: priority scores (integer vector)
+  # @indices: a vector of indices which define the bootstrap sample.
+  # @returns: an estimate of RATE, together with the TOC curve.
+  estimate <- function(data, indices, q) {
```
erikcs, line 179: Am also wondering about moving this function definition outside, but keeping it private, if you want "expert" users to be able to call directly into grf:::estimate_rate with their own DR estimates ...

erikcs:
> Thank you! Agreed that it's very helpful to have the `estimate_rate` function outside both for readability and for making it visible to advanced users. Might be good for us to also include something on this in the vignette so people know its available to them.

It's still private and unexported but users who want to poke into internals can always use `grf:::some_internal_function` at their own risk!

swager (approved):
This looks great! Thanks again for all your work on this; excited to have this merged in.

As part of trying this out, I ran the following pretty aggressive sanity test on weighting. I don't think we should necessarily add it to our regular tests (since using a large enough `n` to make the results stable makes this take a little bit of time, and since we already have similar tests), but thought I'd include it here as a comment just for future reference.

```R
n <- 40000
p <- 4
W <- rbinom(n, 1, 0.5)
U <- rbinom(n, 1, 0.75)
TAU <- U * pmax(X[, 1], 0)
Y <- TAU * W + X[, 2] + pmin(X[, 3], 0) + rnorm(n)

G <- 1 - 2/3 * U
cc <- which(rbinom(n, 1, G) == 1)

tau.forest.1 <- causal_forest(X, Y, W, W.hat = 0.5)
tau.forest.2 <- causal_forest(X[cc,], Y[cc], W[cc], W.hat = 0.5)
tau.forest.3 <- causal_forest(X[cc,], Y[cc], W[cc], sample.weights = 1/G[cc], W.hat = 0.5)

prio <- X[,1]
rate1 <- rank_average_treatment_effect(tau.forest.1, prio)
rate2 <- rank_average_treatment_effect(tau.forest.2, prio[cc])
rate3 <- rank_average_treatment_effect(tau.forest.3, prio[cc])

rate1
rate2
rate3

# These should be different
c(rate1$estimate, rate2$estimate)

# These should be similar
c(rate1$estimate, rate3$estimate)

# These should be similar (because in cc sample have 1/2 instead of 3/4 samples with non-zero TAU)
c(rate1$estimate * 2 / 3, rate2$estimate)
```

And output:
```R
> rate1
  estimate    std.err             target
 0.4368839 0.01139083 priorities | AUTOC
> rate2
  estimate    std.err             target
 0.2842102 0.01617574 priorities | AUTOC
> rate3
  estimate    std.err             target
 0.4211128 0.01898505 priorities | AUTOC
> 
> # These should be different
> c(rate1$estimate, rate2$estimate)
[1] 0.4368839 0.2842102
> 
> # These should be similar
> c(rate1$estimate, rate3$estimate)
[1] 0.4368839 0.4211128
> 
> # These should be similar (because in cc sample have 1/2 instead of 3/4 samples with non-zero TAU)
> c(rate1$estimate * 2 / 3, rate2$estimate)
[1] 0.2912559 0.2842102
```

erikcs:
Nice, how about I add the above script to https://github.com/grf-labs/grf/tree/master/r-package/grf/tests/benchmarks to run internally on major releases?

swager:
Sure, sounds good!

erikcs:
Tagging @AlexiaJM: the "AUC" measure for causal forest you asked about a while ago is available in the development version of GRF now: https://grf-labs.github.io/grf/reference/rank_average_treatment_effect.html

AlexiaJM:
much appreciated!

PR #1098, opened 2022-02-16
Extend multi arm causal forest input checks

### LINKED ISSUE #1096: Multi arm causal forest needs more consistent error messages when factor inputs are inconsistent
In multi-arm causal forest, when the user (possibly mistakenly) supplies `W` and `W.hat` such that `levels(W) != colnames(W.hat)`, then:

- If `W.hat` is a matrix, the code runs,
- If `W.hat` is a dataframe, the code errors out but the message is not informative.

```r
rm(list=ls())
n <- 2000
p <- 10
W <- as.factor(sample(c("a", "b", "c"), replace=T, size=n))
Y <- 100 * (W == "a") + rnorm(n)

W.hat <- matrix(1/3, n, 3)

# Case 1: W.hat is a matrix
colnames(W.hat) <- c("x", "y", "z")  # adding wrong column names
tau.forest <- multi_arm_causal_forest(X, Y, W, W.hat=W.hat)  # runs

colnames(W.hat) <- c("c", "b", "a")  # adding right column names in the wrong order
tau.forest <- multi_arm_causal_forest(X, Y, W, W.hat=W.hat)  # runs


# Case 2: W.hat is a data.frame
colnames(W.hat) <- c("x", "y", "z")  # adding wrong column names
tau.forest <- multi_arm_causal_forest(X, Y, W, W.hat=data.frame(W.hat)) # error
#  Error in W.matrix - W.hat : non-numeric argument to binary operator

colnames(W.hat) <- c("c", "b", "a")  # adding right column names in the wrong order
tau.forest <- multi_arm_causal_forest(X, Y, W, W.hat=data.frame(W.hat)) # error
#  Error in W.matrix - W.hat : non-numeric argument to binary operator
```

My proposal here is to always error (as when `W.hat` is a data.frame), but improve the error message.

erikcs:
* Make W.hat/Y.hat input check more robust, but "silently" allow data.frame as well as matrix for consistency, since that is the behavior for X and Y
* Print a warning if W.hat is provided with column names and they do not correspond to the treatment levels W

(closes #1096)

### DIFF
--- a/r-package/grf/R/multi_arm_causal_forest.R
+++ b/r-package/grf/R/multi_arm_causal_forest.R
@@ -250,6 +250,11 @@ multi_arm_causal_forest <- function(X, Y, W,
     if ((NROW(W.hat) != nrow(X)) || NCOL(W.hat) != nlevels(W)) {
       stop("W.hat has incorrect dimensions: should be a matrix of E[W_k | Xi] estimates.")
     }
+    if (!is.null(colnames(W.hat))) {
+      if (!identical(levels(W), colnames(W.hat))) {
+        warning("Column names are provided for W.hat, but do not correspond to the treatment levels W.")
+      }
+    }
   }
 
   W.matrix <- stats::model.matrix(~ W - 1)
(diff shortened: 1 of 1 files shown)


halflearned (approved):
Thanks @erikcs! To be clear: all cases I mentioned in #1096 now run? If so, approved and feel free to merge.

erikcs:
Thanks @halflearned, yes all the examples now run, with appropriate warnings

PR #1105, opened 2022-02-26
Add `sandwich` check for sample.weights=0

erikcs:
Some code paths in `average_treatment_effect`, and `best_linear_projection` calls into the `sandwich` package to compute standard errors. In some rare cases optional sample.weights that might be zero are passed onto this function, and they are currently not handled correctly. This workaround raises an appropriate error message.

The other code paths that relies on manually calculated SEs are updated to be correct for 0 weights in #1104. 

Kindly pinging @zeileis just to make sure he agrees with our version check `(utils::packageVersion("sandwich") <= numeric_version("3.0.1")` where we assume a future 3.0.2 release will be fine with zero weights?


zeileis:
Most likely, yes, but I think you should be on the safe side if you assume that the problem has been handled in 3.1-0...just in case 3.0-2 is a maintenance release only.

PR #1120, opened 2022-03-17
Add RATE vignette

erikcs:


swager (commented):
Thanks a lot for preparing this!

One high level edit, though -- In presenting the RATE, I would prefer starting with the general framework. The RATE is well defined for any prioritization rule, and measures how well it targets treatment.

Then given this definition, we can next discuss how:
- Given full knowledge of the DGP, the scoring rule that maximizes the RATE is the one that targets in decreasing order of CATE;
- With a lot of data, we should expect that targeting based on estimated CATE should also give a good RATE;
- But in some settings where we don't have enough data to power an accurate CATE estimator, other approaches (like risk-based targeting) may do better in finite samples.

In practice, RATEs estimated on a test set provide an agnostic way of comparing different methods for prioritizing treatment on a level playing field.

(The main point with this is that I don't want to give the idea that the "ideal" use of the RATE occurs when we know tau(x). Rather, it occurs when we don't, and need to compare candidate prioritization rules obtained via potentially vastly different means.)

erikcs:
Got it, the revision above starts with the general S(Xi) (@scottfleming also suggested that, I tried with S(Xi)=tau(X) first to make things as simple as possible, but it's hopefully still easy to understand!)

r-package/grf/vignettes/rate.Rmd
```diff
@@ -0,0 +1,216 @@
...
+
+The tool we propose to employ in this situation is the "RATE". It takes as input a prioritization rule $S(X_i)$ which may be based on estimated CATEs, risk scores, etc, and outputs a scalar metric that can be used to assess how well it targets treatment (it measures the *value* of a prioritization rule while being agnostic to exactly how it was derived). The RATE has two components: the "TOC" and the area under the TOC, which we explain in the following section.
```
syadlowsky, line 34: "is a RATE metric" instead of "is the ''RATE''" and similar in the end "A RATE [metric]"

r-package/grf/vignettes/rate.Rmd
```diff
@@ -0,0 +1,216 @@
...
+  annotate("segment", x = .1, xend = .05, y = ypoint, yend = ypoint,
+           arrow = arrow(type = "closed", length = unit(0.02, "npc"))) +
+  theme_classic() +
+  labs(y = "", title = "Targeting Operator Characteristic")
+```
+
+Note how high the TOC is for the first few units treated: this is intuitive since our toy CATEs are normally distributed with mean zero, if we identify and treat only the top 5% of units with treatment effects greater than $z_{0.95}$, the benefit over treating everyone (0) is large. As we move further along, mixing in more people with lower treatment effect, we get closer to the ATE, until at $q=1$ we equal it.
+
+## RATE: The area under the TOC
+This picture suggests a quite natural way to asses the value of treating every according to $S(X_i)$: sum up the area under the curve. This is exactly how the **RATE** is defined, it is the *A*rea *U*nder the *TOC* ("AUTOC")
```
syadlowsky, line 84: "assess the value of treating every according to ..." is probably better worded as "This picture suggests a quite natural way to summarize the TOC curve in terms of the ability of $S(X_i)$ to rank individuals by treatment benefit:"

r-package/grf/vignettes/rate.Rmd
```diff
@@ -0,0 +1,216 @@
...
+  geom_line() +
+  geom_hline(yintercept = 0, lty = 3) +
+  theme_classic() +
+  labs(y = "", title = "Targeting Operator Characteristic") +
+  geom_ribbon(aes(ymin = 0), ymax=TOC, alpha = 0.5, fill = "red") +
+  annotate(geom = 'text', x = 0.15, y = 0.6, color = 'black', hjust = -0.1,
+           label = "RATE")
+```
+
+There are different ways this area can be summed up, which gives rise to different RATE metrics. We refer to the integral above as the "AUTOC". This implies placing more weight on areas under the curve where the expected treatment benefit is largest. When non-zero treatment effects are concentrated among a small subset of the population this weighting has greater power for testing against a sharp null (RATE=0) as upweighting the individuals with a non-zero treatment effect increases the RATE.
```
syadlowsky, line 114: "This implies placing" -> "This places"

syadlowsky: I don't really understand "greater power for testing against a sharp null (RATE=0) as upweighting the individuals with a non-zero treatment effect increases the RATE." Perhaps we can just say "is powerful for testing against a sharp null (AUTOC=0)."?

r-package/grf/vignettes/rate.Rmd
```diff
@@ -0,0 +1,216 @@
...
+
+Another way to sum this area would be to weight each point on the curve by $q$: $\int_0^1 q \textrm{TOC}(q) dq$. We refer to this metric as the "QINI" (Radcliffe, 2007). This weighting implies placing as much weight on units with low treatment effects as on units with high treatment effects. When non-zero treatment effects are more diffuse across the entire population, this weighting tends to give greater power when testing against a null effect.
+
+Another thing to note is that the "high-vs-low" out-of-bag CATE quantile construction used in Athey and Wager (2019) can also be seen as a kind of RATE, where in place of Qini's identity weighting it would have point mass at q = a given quantile.
+
+## Estimating the RATE
```
syadlowsky, line 120: In general, I'm not sure what to think of "the RATE" vs "a RATE" (as RATEs are a class of metrics that require one to specify the weighing function, and a class of estimators that requires a specific weighting function _and_ CATE score).

r-package/grf/vignettes/rate.Rmd
```diff
@@ -0,0 +1,216 @@
...
+&\textrm{TOC}(q) = E[Y^{test}_i(1) - Y^{test}_i(0) \,|\, \hat \tau^{train}(X^{test}_i) \geq \hat F^{-1}(1 - q)]\\
+& \ \ \ \ \ \ \ \ \ \ \ \ \ \ \ \  - E[Y^{test}_i(1) - Y^{test}_i(0)],
+\end{split}
+\end{equation*}
+
+where $\hat F$ is the empirical distribution function of $\hat \tau^{train}(X^{test})$. GRF estimates
+
+$$\tau(X_i)= [Y^{test}_i(1) - Y^{test}_i(0) \,|\, X_i = x]$$
+
+using a suitable doubly robust score construction $\Gamma_i \approx \tau(X_i)$ which is valid in a variety of settings, enabling inference with the bootstrap, and the final AUTOC estimate is constructed from $\frac{1}{n} \sum_{i=1}^{n} \textrm{TOC}(\frac{i}{n})$. This functionality is implemented in the function `rank_average_treatment_effect` which as first argument takes an "evaluation" forest which computes the appropriate score construction $\hat \Gamma_i$ and second argument estimated prioritization scores for units in the evaluation set, using for example a CATE function estimated on a training set (with additional support for forests trained with sample weights or clusters).
```
syadlowsky, line 134: $\Gamma_i \approx \tau(X_i)$ is not generally true, however $\mathbb{E}[ \Gamma_i | X_i] \approx \tau(X_i)$ is.

syadlowsky: Also we should be careful about $\hat \Gamma_i$ vs $\Gamma_i$. Not sure which you intended to use here, or if you intended to use different ones in the beginning and end of paragraph to indicate the replacement of the true nuisance parameters with estimates. If the latter, it's good to be specific about it in the writing, as it can be confusing if you don't a priori know what it means.

r-package/grf/vignettes/rate.Rmd
```diff
@@ -0,0 +1,216 @@
...
+rate.accord
+```
+
+```{r}
+par(mfrow = c(1, 2))
+plot(rate.sprint, xlab = "Treated fraction", main = "TOC evaluated on SPRINT\n tau(X) estimated on ACCORD")
+plot(rate.accord, xlab = "Treated fraction", main = "TOC evaluated on ACCORD\n tau(X) estimated on SPRINT")
+```
+
+In this toy example there is no evidence of significant HTEs in the two trials (note: this can also be attributed to a) low power, as perhaps the sample size is not large enough to detect HTEs, or b) that the HTE estimator does not detect them). For a broader analysis comparing different prioritization strategies and CATE estimators see Yadlowsky et al., (2021).
```
syadlowsky, line 201: Perhaps also mention that it could be because the heterogeneity in the treatment effects along observable predictor variables are negligible.

syadlowsky (changes requested):
Overall, this is very well-written, and a helpful introduction. Thank you for taking the time to write this. I've added a few comments clarifying wording that we agonized over when writing the RATE paper to minimize confusion.

r-package/grf/vignettes/rate.Rmd
erikcs: I only wanted to get Athey and Wager (2019) into the vignette since it is a widely popular way people "check" for HTEs using causal forest and it seems to me RATE can replace this exercise? 

I am not exactly sure what the exact formal equivalence is, but I do suggest we get a note on this into the vignette, otherwise users may end up asking things like: should we still do quantile sorts ala Athey/Wager/etc? Maybe you a have good formulation in mind?

r-package/grf/vignettes/rate.Rmd
erikcs: I only wanted to convey to the user that given S(X), GRF computes the TOC using using doubly robust scores and not direct tau(X) estimates on the evaluation set, as I can imagine that would be something that could confuse users who end up poking into this, I've added a parenthesis to this paragraph, if you think it's better now?

r-package/grf/vignettes/rate.Rmd
erikcs: Done. Maybe also be extra careful and make sure this final sentence is 100% agreeable as I can imagine this is something many applied people will take as the literal conclusion after running RATE and copy/paste as appropriate

Also, the RATE paper didn't compute QINI for this example, I can imagine some people will ask why not?

r-package/grf/vignettes/rate.Rmd
erikcs: I think ideally there would be a one-liner explaining what a DR score is and why they're used - I did not manage to come up with a good one

r-package/grf/vignettes/rate.Rmd
erikcs: Also, as a bonus, if possible it would be great with a one-liner relating RATE to what the [test_calibration](https://grf-labs.github.io/grf/reference/test_calibration.html) "omnibus" test tells you? (I still don't get exactly what that word means here...). You and Stefan knows best, just asking if you have anything in mind! (trying to pre-empt future questions users are going to ask about rate)

r-package/grf/vignettes/rate.Rmd
```diff
@@ -0,0 +1,216 @@
...
+
+## Treatment prioritization rules
+We are in the familiar experimental setting (or unconfounded observational study) and are interesting in the problem of determining which individuals to assign a binary treatment $W=\{0, 1\}$ and the associated value of adhering to this treatment assignment. Given some subject characteristics $X_i$ we have access to a subject-specific *prioritization rule* $S(X_i)$ which assigns scores to subjects. This prioritization rule should give a high score to units which we believe to have a large benefit of treatment and a low score to units with a low benefit of treatment. By benefit of treatment, we mean the difference in outcomes $Y$ from receiving the treatment given some subject characteristics $X_i$, as given by the conditional average treatment effect (CATE)
```
scottfleming, line 27: "are interesting in the problem" --> "are interested in the problem"

r-package/grf/vignettes/rate.Rmd
```diff
@@ -0,0 +1,216 @@
...
+set.seed(123)
+options(warn = -1)
+```
+
+```{r setup}
+library(grf)
+library(ggplot2)
+```
+
+This vignette gives a brief introduction to how the *Rank-Weighted Average Treatment Effect* (*RATE*) available in the function `rank_average_treatment_effect` can be used to evaluate how good estimates of conditional average treatment effects or other treatment *rules* are at distinguishing subpopulations with different treatment effects, or whether there is any notable heterogeneity present. The second part of the vignette gives a worked example from a medical setting using (synthetic) data from the SPRINT and ACCORD hypertension drug trials. For complete details we refer to the [RATE paper](https://arxiv.org/abs/2111.07966).
```
scottfleming, line 24: perhaps change "treatment rules" --> "treatment prioritization rules" to be consistent with the subheader below.

r-package/grf/vignettes/rate.Rmd
scottfleming: The first sentence here is a bit difficult to parse... the "value of adhering to this treatment assignment" can sound like we're trying to understand the value of individual-level treatment adherence (i.e., how much benefit do patients get by complying and following through with treatment if they were assigned treatment, as well as how much benefit patients get from not taking the treatment if they were not assigned treatment). My intuition here rather that we're assessing the value of the overall allocation.

A potential modification: "associated value of adhering to this treatment assignment" --> "associated value of this treatment allocation strategy".

scottfleming, line 27: "prioritization rule" --> "treatment prioritization rule" to be consistent with the header

r-package/grf/vignettes/rate.Rmd
scottfleming: "estimates of conditional average treatment effects **or other treatment rules**" assumes that estimates of conditional average treatment effects are a kind of treatment rule but I don't know that this connection is clear/obvious until we say what we mean by a "treatment prioritization rule" in the section below. Wondering if even just switching the order here -- "[...] can be used to evaluate how good treatment prioritization rules (such as conditional average treatment effect estimates) are [...]". Alternatively, just dropping the word "other" could also do the trick, I think. Feel free to veto any of the above.

r-package/grf/vignettes/rate.Rmd
```diff
@@ -0,0 +1,216 @@
...
...ve the best purpose for treatment targeting? The answer is that this general problem formulation is quite convenient, as in some settings we may lack sufficient data to power an accurate CATE estimator and have to rely on other approaches to target treatment. Examples of other approaches are heuristics derived by domain experts or simpler models predicting risk scores (where risk is defined as $P[Y = 1 | X_i =  x]$ and $Y \in \{0, 1\}$ with $Y=1$ being an adverse outcome), which is quite common in clinical applications. Consequently, in finite samples we may sometimes do better by relying on simpler rules which are correlated with the CATEs, than on noisy and complicated CATE estimates (remember: CATE estimation is a hard statistical task, and by focusing on a general rule we may circumvent some of the problems of obtaining accurate non-parametric point estimates by instead asking for estimates that *rank* units according to treatment benefit). Also, even if you have an accurate CATE estimator, there may be many to choose from (neural nets/random forests/various metalearners/etc). The question is: given a set of treatment assignment rules $S(X_i)$, which one (if any) should we use?
```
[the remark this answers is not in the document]
scottfleming: "the question is this:" (nitpicking but colons should only follow complete sentences)

r-package/grf/vignettes/rate.Rmd
erikcs: I agree, reversing the sentence makes it nicer

scottfleming: Thinking more about this, I feel like we want to be crystal clear in this first paragraph about what questions the RATE can answer. So something like 
"This vignette gives a brief introduction to how the *Rank-Weighted Average Treatment* (*RATE*) available in the function `rank_average_treatment_effect` can be used to answer questions such as,

1. Is there any significant heterogeneity in treatment effects in this sample?
2. Are my conditional average treatment effect (CATE) estimates aligned with the true CATE?
3. Does my proposed treatment prioritization rule do any better than assigning treatment uniformly at random?
4. I have two treatment prioritization rules (e.g., CATE estimators, risk estimators); is one significantly better than the other?

The second part of the vignette [...]"

I don't have strong feelings about the exact wording or ordering here (maybe @syadlowsky or @swager do) but I just want to make sure that if someone is quickly searching through available tools for evaluation they can easily and quickly - within the first paragraph - determine whether this is the right tool to use.

r-package/grf/vignettes/rate.Rmd
```diff
@@ -0,0 +1,216 @@
...
+  geom_hline(yintercept = 0, lty = 3) +
+  annotate(geom = 'text', x = .1, y = ypoint, color = 'black', hjust = -0.1,
+           label = 'ATE of top 5% minus ATE') +
+  annotate("segment", x = .1, xend = .05, y = ypoint, yend = ypoint,
+           arrow = arrow(type = "closed", length = unit(0.02, "npc"))) +
+  theme_classic() +
+  labs(y = "", title = "Targeting Operator Characteristic")
+```
+
+Note how high the TOC is for the first few units treated: this is intuitive since our toy CATEs are normally distributed with mean zero, if we identify and treat only the top 5% of units with treatment effects greater than $z_{0.95}$, the benefit over treating everyone (0) is large. As we move further along, mixing in more people with lower treatment effect, we get closer to the ATE, until at $q=1$ we equal it.
```
scottfleming, line 81: "Note how high the TOC is for the first few units treated**.** This is intuitive since our toy CATEs are normally distributed with mean zero**;** if we identify and treat only the top 5% of units with treatment effects greater than $z_{0.95}$, the benefit over treating everyone (0) is large."

r-package/grf/vignettes/rate.Rmd
```diff
@@ -0,0 +1,216 @@
...
+
+## RATE: The area under the TOC
+This picture suggests a quite natural way to asses the value of treating every according to $S(X_i)$: sum up the area under the curve. This is exactly how the **RATE** is defined, it is the *A*rea *U*nder the *TOC* ("AUTOC")
+
+$$\textrm{RATE} = \int_0^1 \textrm{TOC}(q) dq .$$
+If there is barely any heterogeneity in $\tau(X_i)$ this area will be vanishingly small and in the special case where $\tau(X_i)$ is constant, it's zero. Thinking one step ahead, if our *estimated* rule $S(X_i)$ do well in identifying the individuals with very different benefit of treatment, we would expect this metric to be *positive*. Conversely, if they do badly, or there barely are any benefit to stratifying treatment, we would expect it to be *negative* or close to *zero*, respectively.
```
scottfleming, line 87: "if our *estimated* rule $S(X_i)$ do well" --> "if our *estimated* rule $S(X_i)$ does well"

scottfleming: "benefit of treatment" --> "benefits of treatment"?

scottfleming: "if they do badly" --> "if it does badly"

scottfleming, line 87: "or there barely are are any benefit" --> "or there are barely any benefits"

r-package/grf/vignettes/rate.Rmd
scottfleming: "[...] and second argument estimated [...]" --> "[...] and as second argument the estimated [...]"

scottfleming: @erikcs for DR score explanation what about something like the following: 

“where $\hat F$ is the empirical distribution function of $\hat \tau^{train}(X^{test})$. _In order to estimate these expectations, we employ the use of doubly robust scores “scores”, $\hat \Gamma_i$, with the property that they act as nearly unbiased (but noisy) proxies for the true CATE._ GRF estimates [...] using suitable  with the property that they act as nearly unbiased (but noisy) proxies for [...]"

scottfleming: may want to change "estimated prioritization scores" --> "estimated treatment priorities" to avoid confusing doubly robust scores with the output of the treatment prioritization rule

scottfleming: But yes I think the parenthetical is helpful here. I might also clarify that the "evaluation" forest is trained on the test set. Though maybe that's already obvious and trying to be explicit here would just be more confusing...

scottfleming: "using for example a CATE function estimated on a training set (with additional support for forests trained with sample weights or clusters)." --> "using for example a CATE estimator trained on a training set [...]" (just to clarify that the training is done on the training set but the estimation for the purposes of the RATE is actually done on the test/evaluation set)

scottfleming: And I agree with Steve Y that it's probably worth being at least consistent about our use of $\hat \Gamma$ vs. $\Gamma$. In particular the  $\mathbb{E}[ \Gamma_i | X_i] \approx \tau(X_i)$ should be  $\mathbb{E}[ \hat \Gamma_i | X_i] \approx \tau(X_i)$.

Might also be worth changing $\frac{1}{n} \sum_{i=1}^{n} \textrm{TOC}(\frac{i}{n})$ to $\frac{1}{n} \sum_{i=1}^{n} \hat \textrm{TOC}(\frac{i}{n})$, to maintain consistency for folks who are trying to connect equations in the paper with implementation in `grf`.

r-package/grf/vignettes/rate.Rmd
```diff
@@ -0,0 +1,216 @@
...
+
+## An application to SPRINT and ACCORD
+To illustrate RATE we consider an example application from a medical setting. Two large randomized trials *ACCORD* (ACCORD Study Group, 2010) and *SPRINT* (SPRINT Research Group, 2015) conducted on similar populations and designed to measure the effectiveness of a hypertension treatment find that the treatment is effective. However, the two trials differs in the magnitude of the estimated benefit. Various explanations for this finding have been proposed, we'll focus on one in particular here: the hypothesis that the difference is due to *heterogeneity in treatment effects* (see Yadlowsky et al., 2021 for references).
```
scottfleming, line 137: To clarify, the two trials actually came to different conclusions. *SPRINT* found the treatment was effective, *ACCORD* found that the treatment was not effective.

More specifically,

*[SPRINT](https://www.nejm.org/doi/full/10.1056/nejmoa1511939)* found that "Among patients at high risk for cardiovascular events but without diabetes, targeting a systolic blood pressure of less than 120 mm Hg, as compared with less than 140 mm Hg, resulted in lower rates of fatal and nonfatal major cardiovascular events and death from any cause". 

On the other hand, *[ACCORD](https://www.nejm.org/doi/full/10.1056/nejmoa1001286)* found that "In patients with type 2 diabetes at high risk for cardiovascular events, targeting a systolic blood pressure of less than 120 mm Hg, as compared with less than 140 mm Hg, did not reduce the rate of a composite outcome of fatal and nonfatal major cardiovascular events."

r-package/grf/vignettes/rate.Rmd
```diff
@@ -0,0 +1,216 @@
...
+
+This hypothesis has a testable implication implied by the previous section: if there is significant heterogeneity present and we are able to effectively estimate these with a powerful CATE estimator, then an estimated RATE on ACCORD and SPRINT should be positive and significant. In particular, our setup implies the following recipe:
+
+1) Estimate CATE functions $\hat \tau^{ACCORD}(\cdot)$ and $\hat \tau^{SPRINT}(\cdot)$ on ACCORD and SPRINT data.
+
+2) Use $\hat \tau^{ACCORD}(X^{SPRINT})$ to evaluate RATE on SPRINT, and vice versa: use $\hat \tau^{SPRINT}(X^{ACCORD})$ to evaluate RATE on ACCORD.
+
+3) If both populations exhibit systematic HTEs, then a powerful CATE estimator should yield a positive and significant RATE estimate.
```
scottfleming, line 145: "If both populations exhibit systematic HTEs" --> "If both populations exhibit similar HTEs". I'm not actually sure here what was meant by "systematic", but I think this is an obnoxiously important point to make b/c it's possible that the both trials exhibit HTEs but they're just different and so you would get a RATE \approx 0 if you trained on ACCORD and tested on SPRINT in that case but you would get a RATE > 0 if you trained on ACCORD and tested on ACCORD (which some research groups have done/are actively doing)

r-package/grf/vignettes/rate.Rmd
erikcs: > 

Yeah that kind of just reads like a tautology given E[Gamma] ~ tau, I was thinking of something along the efficiency explanation of DR scores used in ATE estimation in the first paragraph here https://grf-labs.github.io/grf/REFERENCE.html#average-treatment-effects

r-package/grf/vignettes/rate.Rmd
erikcs: Can you use github's "add a suggestion" when making comments so I can just commit your updated paragraph directly?

And is it possible to rephrase it using less (none if possible) medical jargon?  A newbie from a social science background will only understand a fraction of those words (composite included). Remember the way the example is represented here is supposed to demonstrate RATE for a newbie, I think it's safe to gloss over all the finer details that are not necessary to understands that you can use RATE in this setting (all those details are in the arxiv anyways)

r-package/grf/vignettes/rate.Rmd
erikcs: In the rate paper you write 
```
The approach for replacing τ with Γ has been used for average
treatment effect estimation (Chernozhukov et al., 2018a), conditional average treatment
effect estimation (Kennedy, 2020), policy learning (Athey and Wager, 2021), and structural
functions of treatment effects (Semenova and Chernozhukov, 2017). The key to proving
Theorem 4 (done in Section 8) is extending this technique to L-statistics
```

It would be nice with an easy to digest sentence that gives intuition on WHY this is essential. Then the paragraph in this section could end with something like "You might ask why we replace estimates τ with Γ when computing the TOC when τ are already "valid" point estimates of the CATE. This is the same strategy GRF takes in functions like `best_linear_projection` and `average_treatment_effect`, because this score construction is shown to give semiparametrically efficient and valid inference (see XXX). The intuition is that the CATE point estimates would understate the sampling variation in ... as they rely on estimated "nuisance" components ... <--- I don't know how to explain this well

r-package/grf/vignettes/rate.Rmd
scottfleming:
```suggestion
To illustrate RATE we consider an example application from a medical setting. Two large randomized trials *ACCORD* (ACCORD Study Group, 2010) and *SPRINT* (SPRINT Research Group, 2015) conducted on similar populations and designed to measure the effectiveness of a hypertension treatment found different results regarding the treatment's effectiveness: *SPRINT* found the treatment to be effective, while *ACCORD* found no significant treatment benefit. Various explanations for this finding have been proposed, we'll focus on one in particular here: the hypothesis that the difference is due to *heterogeneity in treatment effects* (see Yadlowsky et al., 2021 for references).
```

scottfleming: Yes! Sure thing - thanks for the "add a suggestion" tip, I'll definitely use that going forward! And totally agreed re:limiting jargon. Sorry, that was confusing - I'm not at all suggesting that we include unnecessary details about the trials, just wanted to clarify what the actual results were so we're all on the same page.

scottfleming (approved):
Oh sorry I see that you've already made the fix - this looks great!

scottfleming (approved):

r-package/grf/vignettes/rate.Rmd
```diff
@@ -0,0 +1,216 @@
...
+
+2) Use $\hat \tau^{ACCORD}(X^{SPRINT})$ to evaluate RATE on SPRINT, and vice versa: use $\hat \tau^{SPRINT}(X^{ACCORD})$ to evaluate RATE on ACCORD.
+
+3) If both populations exhibit similar HTEs, then a powerful CATE estimator should yield a positive and significant RATE estimate.
+
+For the purpose of this vignette example we'll not analyse the original SPRINT/ACCORD data, but rather a *smaller simulated* example, stored in the [GRF repository](https://github.com/grf-labs/grf/blob/master/r-package/grf/vignettes/synthetic_SPRINT_ACCORD.RData). For details on the simulator see the [RATE repository](https://github.com/som-shahlab/RATE-experiments/blob/main/experiments/section_5_SPRINT_and_ACCORD/simulators/simulators_description.pdf).
+
+The outcome data in this trial is *right-censored*, so we cannot use any ordinary "out-of-the-box" CATE estimator to estimate treatment effects, rather, we need one that takes censoring into account. For this reason we use GRF's `causal_survival_forest` (and refer to the documentation for details). Below is an illustration of the simulated trial data.
```
scottfleming, line 149:
```suggestion
The outcome data in this trial is *right-censored*, so we cannot use any ordinary "out-of-the-box" CATE estimator to estimate treatment effects, rather, we need one that takes censoring into account. For this reason we use GRF's `causal_survival_forest` (and refer to the [documentation](https://grf-labs.github.io/grf/reference/causal_survival_forest.html) for details). Below is an illustration of the simulated trial data.
```

r-package/grf/vignettes/rate.Rmd
```diff
@@ -0,0 +1,216 @@
...
+\begin{equation*}
+\begin{split}
+&\textrm{TOC}(q) = E[Y^{test}_i(1) - Y^{test}_i(0) \,|\, \hat \tau^{train}(X^{test}_i) \geq \hat F^{-1}(1 - q)]\\
+& \ \ \ \ \ \ \ \ \ \ \ \ \ \ \ \  - E[Y^{test}_i(1) - Y^{test}_i(0)],
+\end{split}
+\end{equation*}
+
+where $\hat F$ is the empirical distribution function of $\hat \tau^{train}(X^{test})$. GRF estimates
+
+$$\tau(X_i)= [Y^{test}_i(1) - Y^{test}_i(0) \,|\, X_i = x]$$
```
swager, line 132: This is not quite right -- GRF doesn't "estimate" tau(X_i) using the DR score; rather, the DR score is a nearly unbiased measurement of tau(X_i). The reason I wouldn't want to call the DR score an "estimate" is that it's not consistent (and doesn't try to be!).

I think it might be best to just say that the `rank_average_treatment_effect` function delivers a AIPW-style doubly robust estimator of RATE based on a test set; then we refer to the paper for a derivation of the doubly robust estimator and the associated central limit theorem.

r-package/grf/vignettes/rate.Rmd
```diff
@@ -0,0 +1,216 @@
...
+rate.accord
+```
+
+```{r}
+par(mfrow = c(1, 2))
+plot(rate.sprint, xlab = "Treated fraction", main = "TOC evaluated on SPRINT\n tau(X) estimated on ACCORD")
+plot(rate.accord, xlab = "Treated fraction", main = "TOC evaluated on ACCORD\n tau(X) estimated on SPRINT")
+```
+
+In this toy example both AUTOCs are insignificant at conventional levels, suggesting there is no evidence of significant HTEs in the two trials. Note: this can also be attributed to a) low power, as perhaps the sample size is not large enough to detect HTEs, b) that the HTE estimator does not detect them, or c) the heterogeneity in the treatment effects along observable predictor variables are negligible. For a broader analysis comparing different prioritization strategies and CATE estimators see Yadlowsky et al., (2021).
```
swager, line 201: "toy example" -> "semi-synthetic example"

update last sentence: For a broader analysis comparing different prioritization strategies on the SPRINT and ACCORD datasets, see Yadlowsky et al. (2021).

swager (approved):
Looks great, thank you!

r-package/grf/vignettes/rate.Rmd
erikcs: It's updated, I think ideally a vignette like this aimed at being understandable to basically _everyone_ we should avoid jargon abbreviations like AIPW and scary (to people who ain't PhDs) terms like "central limit theorem"  but I guess it's fine here

PR #1130, opened 2022-04-01
Add forest weight example to REFERENCE

erikcs:
Adding a simple example at the end of the Prediction section to help explain how GRF predictions differs from a "traditional" random forest


erikcs:
Going ahead with merge as OK'd by @swager

PR #1138, opened 2022-04-18
Add lm_forest

### LINKED ISSUE #733: Multi-treatment GRF
This is an overview issue - in the first round we aim for a causal forest with multiple categorical treatments `W` and a potentially vector valued outcome `Y`. Here is a table of the causal forest "taxonomy"

|    	| outcome Y   	| treatment W           	| instrument Z          	| name                         	|                               	|
|----	|-----	|-------------	|------------	|------------------------------	|-------------------------------	|
| a) 	| R   	| R           	|            	| `regression_forest`          	|                               	|
| b) 	| R  	| {1, ..., K}           	|            	| `probability_forest`          	|                               	|
| c) 	| R^K 	| R           	|            	| `multi_regression_forest`    	|                               	|
| d) 	| R   	| R / {0, 1}  	|            	| `causal_forest`              	| e(X): regression_forest       	|
| e) 	| R^K   	| {1, ..., K} 	|            	| `multi_arm_causal_forest` 	| e(X): probability_forest m(X): multi_regression_forest    	|
| f) 	| R   	| R^K         	|            	| `lm_forest`                          	| E(W|X): multi_regression_forest 	|
|    	|     	|             	|            	|                              	|                               	|
| g) 	| R   	| R / {0, 1}  	| R / {0, 1} 	| `instrumental_forest`        	|                               	|

_Building blocks:_

- [x] b) **Multi-class probability** estimation (`probability_forest`). #711 
    Future follow-ups:
    1. Incorporate sample weights into ProbabilitySplittingRule. #832 
    2. Tuning
    3. (Some potential performance modifications detailed in #711).
  <br/>
- [x] c) **Multi-outcome regression** (`multi_regression_forest`). #742 
    Can be used to estimate multivariate generalized propensity scores, and split on vector-valued pseudo outcomes.

    Future follow-ups:
    1. Tuning
    2. Variance estimates

_New forests:_

- [x] e) **Multi arm causal forest** (`multi_arm_causal_forest`). #748 
    Estimate treatment effects with multiple mutually exclusive categorical treatments 1,...K.
  
    Future follow-ups:
    1. Support for multivariate outcomes Y. #815 
    2. A `MultiRegressionSplittingRule` which supports more constraints, such as #W in each child. #921.
    3. Integrate with `best_linear_projection`
    4. Tuning
  <br/>

- [x] f) Some specific or generic forest-variant for multiple continuous treatments. #1138 

_General follow ups:_ 
- Performance enhancements to Data model, see issue #747
- [x] Update causal_benchmark.R. #831

(Thanks to @rajkumarkarthik for his original MGRF prototype in https://github.com/rajkumarkarthik/mgrf-develop)

erikcs:
Adding a final MGRF component, `lm_forest(X, Y, W)` which can be used to estimate h_k(x) in the conditional linear model `Y = c(x) + h_1(x)W_1 + ... + h_K(x)W_k`

An optional `gradient.weights` specifying which coefficient one wants to target heterogeneity is added separately in #1140.

(closes #733)

(test install the branch with `devtools::install_github("erikcs/grf", subdir = "r-package/grf", ref = "lm_forest")`)

[Basic rdd sim gist](https://gist.github.com/erikcs/bfb262c2bb78e97d63499195a0797208)


r-package/grf/R/lm_forest.R
```diff
@@ -0,0 +1,354 @@
...
+#' W <- as.numeric(Z >= cutoff)
+#' tau <- pmax(0.5 * X[, 1], 0)
+#' Y <- tau * W  + 1 / (1 + exp(2 * Z)) + 0.2 * rnorm(n)
+#'
+#' # Compute the Imbens-Kalyanaraman MSE-optimal bandwidth for a local linear regression.
+#' if (require("rdd", quietly = TRUE)) {
+#'   bandwidth <- IKbandwidth(Z, Y, cutoff)
+#'   # Compute kernel weights for a triangular kernel.
+#'   sample.weights <- kernelwts(Z, cutoff, bandwidth, "triangular")
+#' } else {
```
swager, line 96: I don't think we should use an arbitrary BW here, even though it's an example. How about making the loading of rdd non-optional for running this example, then add a comment showing how to alternatively hand-set the BW:

bandwidth <- # user can hand-specify this
dist <- abs((Z - cutoff) / bandwidth)
sample.weights <- (1 - dist) * (dist <= 1) / bandwidth

r-package/grf/R/lm_forest.R
```diff
@@ -0,0 +1,354 @@
...
+#'                          (for confidence intervals). This option is currently
+#'                          only supported for univariate outcomes Y.
+#' @param ... Additional arguments (currently ignored).
+#'
+#' @return A list with elements `predictions`: a 3d array of dimension [num.samples, K, M] with
+#' predictions for regressor W, for each outcome 1,..,M (singleton dimensions in this array can
+#' be dropped by passing the `drop` argument to `[`, or with the shorthand `$predictions[,,]`),
+#'  and optionally `variance.estimates`: a matrix with K columns with variance estimates.
+#'
+#' @examples
```
swager, line 264: Does it make sense to duplicate the example? Why not just say: See lm_forest for example usage

r-package/grf/tests/testthat/test_lm_forest.R
```diff
@@ -0,0 +1,61 @@
...
+  p <- 5
+  X <- matrix(rnorm(n * p), n, p)
+  W <- rbinom(n, 1, 0.5)
+  Y <- pmax(X[, 1], 0) * W + X[, 2] + pmin(X[, 3], 0) + rnorm(n)
+  wts <- sample(1:2, n, TRUE)
+
+  Y.hat <- predict(regression_forest(X, Y, num.trees = 500))$predictions
+  cf <- causal_forest(X, Y, W, Y.hat = Y.hat, W.hat = 0.5, sample.weights = wts, num.trees = 250, stabilize.splits = FALSE)
+  lmf <- lm_forest(X, Y, W, Y.hat = Y.hat, W.hat = 0.5, sample.weights = wts, num.trees = 250)
+  expect_lt(mean((predict(cf)$predictions - predict(lmf)$predictions[,,])^2), 0.12)
```
swager, line 13: This is a pretty loose tolerance for a DGP like this. It'd be reassuring to see something like this with a better tolerance.

swager (commented):
Thanks for putting this together! Here are some minor comments from a first read-through.

swager:
Also, I think we should add an error check to see if any of the passed-in W are constant. I imagine some users may (accidentally) want to pass in an intercept; we should have an error message that tells them we already have an intercept.

r-package/grf/tests/testthat/test_lm_forest.R
erikcs: Oh yes, I kept bumping the tol to make it never fail and accidentally made it not discriminate anything anymore

r-package/grf/R/lm_forest.R
erikcs: Yes that's just the template for all forests + predict.forest

erikcs:
> Also, I think we should add an error check to see if any of the passed-in W are constant. I imagine some users may (accidentally) want to pass in an intercept; we should have an error message that tells them we already have an intercept.

Good point! Made it a warning instead of error though (just imagining in the future of someone runs it as a large pipeline with random W input/subsets then lm_forests won't break it - it just doesn't split if a constant column appears)

r-package/grf/R/lm_forest.R
```diff
@@ -0,0 +1,359 @@
...
+#'
+#' # Alternatively, specify bandwith and triangular kernel weights without using the `rdd` package.
+#' # bandwidth <- # user can hand-specify this.
+#' # dist <- abs((Z - cutoff) / bandwidth)
+#' # sample.weights <- (1 - dist) * (dist <= 1) / bandwidth
+#'
+#' # Estimate a local linear regression with the running variable Z conditional on covariates X = x:
+#' # Y = c(x) + tau(x) W + b(x) Z.
+#' # Specify gradient.weights = c(1, 0) to target heterogeneity in the RDD coefficient tau(x).
+#' lmf <- lm_forest(X, Y, cbind(W, Z), sample.weights = sample.weights, gradient.weights = c(1, 0))
```
erikcs, line 295: ~equivalently (for a speedup on large examples) one could fit this forest on only the subset with non-zero weights:

```lmf <- lm_forest(X[sample.weights > 0, ], Y[sample.weights > 0], cbind(W, Z)[sample.weights > 0, ], sample.weights = sample.weights[sample.weights > 0], gradient.weights = c(1, 0))```

however I left that out to keep the example simple, and ideally a pro user would know

erikcs: Actually let's add this

swager (approved):

Issue #1262, opened 2023-02-21
Differential forest prediction test statistic not replicable despite seed set in relevant places


shbyrne:
**Description** 

Dear GRF community,

I am using GRF to investigate the presence of heterogeneous treatment effects from an RCT. Despite setting the seed in what I believe are the correct places, I cannot get the test result from my test calibration to replicate across sessions in R. The result is replicable within the session (training a second replication forest (see 'another.forest' below) but I will get a completely different result if I clear the environment, start a new session, and run this code again. The test statistic will swing in and out of significance. 

Similar issues have been flagged before (see #604), with the solution being repeated (and targeted) seed setting to tame the various layers of introduced randomness. This hasn't borne fruit in my case. Could I still be missing a seed placement somewhere? Any suggestions are welcome and greatly appreciated!

**Code extract**

```
data <- read_csv("F:/data.csv")

seed.val<-654321
set.seed(seed.val)

train_idx <- sample.int(nrow(data), size = floor(.8 * nrow(data)), replace = F)
data.train <- data[train_idx, ]
data.test <- data[-train_idx, ]

# create test/train inputs for the CF
X.test <- data.test %>% select(-c(outcome, treated)) %>% as.matrix()
W.test <- data.test %>% select(treated) %>% purrr::as_vector()
Y.test <- data.test %>% select(outcome) %>% purrr::as_vector()
X <- data.train %>% select(-c(outcome, treated)) %>% as.matrix()
W <- data.train %>% select(treated) %>% purrr::as_vector()
Y <- data.train %>% select(outcome) %>% purrr::as_vector()

# causal forest on the training data, tunes all parameters 
set.seed(seed.val)

train.forest = causal_forest(X,
                             Y,
                             W,
                             num.trees = 2500,
                             tune.parameters = c("all"),
                             seed = seed.val)

set.seed(seed.val)

tf_result <-test_calibration(train.forest,vcov.type="HC3")
tf_res_df <-as.data.frame(tf_result[,]) %>% 
  tibble::rownames_to_column(var = "term")


## Check that it replicates
set.seed(seed.val)

another.forest = causal_forest(X,
                             Y,
                             W,
                             num.trees = 2500,
                             tune.parameters = c("all"),
                             seed = seed.val)

set.seed(seed.val)

tf_result_2 <-test_calibration(another.forest,vcov.type="HC3")
tf_result_2
```

**Session info**

GRF version 2.2.0 
R version 4.1.2 (2021-11-01)
[...]

PR #1263, opened 2023-02-22
Assign random number seed independent of num_threads

sebp:
Addresses my comment in https://github.com/grf-labs/grf/issues/1262#issuecomment-1440227209

`ForestTrainer::train_trees` calls `split_sequence` to split the number of trees to grow into batches depending on the specified number of threads. However, when assigning a seed to grow an individual tree, only the batch index is considered, but not the tree index. Therefore, different values for the number of threads will yield different results, despite the user having specified the same (global) random number seed.

The patch initializes the random number generator based on the user provided seed and the index of the tree (`start + i`). Thus, each tree will be assigned the same seed, independent of the specified number of threads.


sebp:
I understand that you want to preserve backwards compatibility. If you someday release a new major version, it would be nice include this fix.

erikcs:
Thanks, will do.

PR #1277, opened 2023-03-23
Request: make `causal_forest` keep the RF's used to make Y.hat and W.hat

beniaminogreen:
Hi all, 

Thanks for such a great package! I imagine you all have considered this already, but would you consider making the `causal_forest` function also return the regression forests used to predict Y.hat and W.hat?  

I've written a pull request to implement this. The forests now have the fields 'forest.Y' and 'forest.W', which are the regression forests used to fit the nuisance parameters. If the forests are not trained, these fields are set to NULL. 

As I see it, the change would have the following effects: 

## Pros:
* Allows users to investigate / evaluate the fit of the forests fit to predict y.hat and w.hat
* Would allow the `conditional_means.causal_forest` and `double_robust_scores.causal_forest` to be extended to predict new conditional means / scores for unseen data. 

## Cons: 
* Would increase the size / memory usage of the causal forests.

### DIFF
--- a/r-package/grf/R/causal_forest.R
+++ b/r-package/grf/R/causal_forest.R
@@ -203,6 +203,7 @@ causal_forest <- function(X, Y, W,
                       num.threads = num.threads,
                       seed = seed)
 
+  forest.Y <- NULL
   if (is.null(Y.hat)) {
     forest.Y <- do.call(regression_forest, c(Y = list(Y), args.orthog))
     Y.hat <- predict(forest.Y)$predictions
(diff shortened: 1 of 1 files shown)


erikcs:
Thank you @beniaminogreen, I wonder if there instead are alternate approaches that would fit your use case better, like just stitching together your own utility function to predict AIPW score on X.test in your project?

For grf, storing Y.forest and W.forest could be convenient, but Y.hat and W.hat could also be user-supplied from e.g. boosting. Also, other estimand configurations needs more than W.hat and Y.hat to compute scores, for example IV forest needs to fit a new "compliance" forest.

So far the closest we've come to adding something like this was just including a code snippet in a docstring explaining how to do it manually: https://grf-labs.github.io/grf/reference/rank_average_treatment_effect.fit.html#examples

(grf has "get_scores" for AIPW scores for a contrast E[Y(1) - Y(0)|X], policytree has "double_robust_scores" for the conditional means E[Y(1)|X] and E[Y(0)|X])

PR #1280, opened 2023-03-24
Allow continuous treatment in causal_survival_forest

erikcs:
The reason CSF was restricted to a binary W was for simplicity, it simplifies nuisance estimation. In principle there is nothing preventing the CSF construction from being applied in the continuous exposure setting, the estimand and interpretation follows from causal forest.

This PR makes no changes to the case when W is binary. When W is continuous an additional survival forest is used to estimate m(x) = E[f(T) | X].


swager (approved):
🌲

PR #1416, opened 2024-05-29
Add cross-fold RATE vignette

erikcs:


r-package/grf/vignettes/rate_cv.Rmd
```diff
@@ -0,0 +1,438 @@
...
+text(2.5, 1.25, offset = 0, "2", col = "blue")
+text(3.5, 1.25, offset = 0, "3", col = "blue")
+text(4.5, 1.25, offset = 0, "4", col = "blue")
+text(5.5, 1.25, offset = 0, "5", col = "red")
+
+text(3, 5.5, "Train", col = "blue")
+text(4, 5.5, "Test", col = "red")
+```
+
+This construction is going to make our $t$-values satisfy a martingale property and allow for valid aggregation. The steps are:
```
swager, line 187: Add citations to LvdL and the sequential validation comment as references for the martingale property claim

r-package/grf/vignettes/rate_cv.Rmd
```diff
@@ -0,0 +1,438 @@
...
+
+  c(
+    reject.sequential = pval.sequential <= 0.05
+  )
+}))
+paste("Rejection rate:", mean(res) * 100, "%")
+#> [1] "Rejection rate: 5.4 %"
+```
+
+Now that we have a valid test for treatment heterogeneity using several folds, let's try it out on a simulated example and compare its performance with simple sample splitting. Below we generate data from synthetic observational data (Setup A in Nie & Wager (2021)) where treatment effect heterogeneity is hard to detect. In this case, we want our test to reject $H_0$ of no heterogeneity.
```
swager, line 250: Make it explicit that you're also implementing test/train sample splitting as a baseline; that test/train splitting is what's proven to be valid in the RATE paper; but that this is potentially lower power since you only have 50% of the data for validation.

r-package/grf/vignettes/rate_cv.Rmd
```diff
@@ -0,0 +1,438 @@
...
+
+## Heuristic: out-of-bag estimation with one-sided tests
+The fundamental challenge with estimating CATEs and evaluating the presence of heterogeneity using the same data is overfitting: we're simply constrained by the fact that we can't simply estimate and evaluate something using the same data set. Random forests come with a neat algorithmic workaround for this issue: out-of-bag (OOB) estimation. With this strategy, we can train a forest using all our data and obtain predictions for the same data. In terms of breaking dependence between training and evaluation, out-of-bag estimation is not as strong as sample splitting, but it can serve as a very useful middle ground.
+Performing traditional two-sided tests with OOB-estimated RATEs is problematic as estimates are going to be biased downwards. However, if we are willing to commit to a one-sided test, then we can use OOB estimates and still be relatively confident we are maintaining nominal error guarantees (we caution that this is a *heuristic* not backed by a formal central-limit result). In the example below we compare a two-sided test for RATE $\neq 0$ with a one-sided test for RATE $>0$ at a 5% level.
```
swager, line 289: There should be more discussion here about how basic OOB is just the leave-one-out version of naive cross-fold estimation, which we already found doesn't work.

The fix of using one sided tests could also be used with usual cross-fold approach.

So there are really two ideas here: One sided tests as a heuristic to fix the naive aggregation, and OOB as a computational speed up with forests.

swager (commented):
Looks very nice! Left some suggestions for small updates to the prose.

erikcs:
Thanks, the text is updated!

swager (approved):

Issue #257, opened 2018-06-30
Any python implementations ? 


snowde:

swager:
We don't have one currently available. We hope to have one eventually, but our first priority is getting the R implementation to 1.0.

If anyone is interested in contributing a Python implementation, please reach out to us.

snowde:
Thanks Stefan,

I have seen some development here, but I haven't tested it out. If you feel the need to comment about it please do.

https://github.com/kjung/scikit-learn
[https://avatars1.githubusercontent.com/u/1491410?s=400&v=4]<https://github.com/kjung/scikit-learn>

kjung/scikit-learn<https://github.com/kjung/scikit-learn>
A version of scikit-learn that includes implementations of Wager & Athey and Scott Powers causal forests.
github.com



Regards,
Derek
________________________________

swager:
Interesting -- it looks like that package uses the "old" causal forest. The grf package has a substantially different implementation of causal forests than we had in the first JASA paper on this (both statistically and computationally). See section 6.2 of https://arxiv.org/pdf/1610.01271.pdf for a discussion + simulation examples.

The upshot is that any Python implementation should just be an interface to the c++ core of grf. Hopefully this will be a reasonable amount of effort, as much of the complexity of this package is in the c++, not in the R.

halflearned:
Even if a Python implementation doesn't exist, you can still use grf in Python via the mostly excellent [rpy2](https://rpy2.readthedocs.io/en/version_2.8.x/) package.

```python
from rpy2.robjects.packages import importr
grf = importr("grf")
grf.regression_forest(X=X, Y=Y, W=W)
```

erikcs:
For reference [here](https://github.com/erikcs/pygrf) is an old Cython wrapping of some basic functionality in grf v 0.10.2, but as @halflearned mentions it is easier to just use rpy2.

crflynn:
I've been working on one here: https://github.com/crflynn/skgrf

It's a bit hastily done and might require some review, particularly on the boosted regressor, but it provides sklearn python classes for all the estimators available in 1.2.0 except for the custom forest.

PR #1429, opened 2024-07-07
WIP python bindings via pybind11

apoorvalal:
create bindings in `python/` that uses pybind11 to wrap grf code. Developing with `regression_forest` as first end-to-end example. Progress documented in `python/notebooks/scratch.ipynb`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,7 @@
 
 # vscode files
 .vscode/
+python/build/*
+python/grf/__pycache__/*
+python/grf.egg-info/*
+python/grf/**/*.so
--- a/core/src/forest/ForestPredictor.cpp
+++ b/core/src/forest/ForestPredictor.cpp
@@ -59,17 +59,28 @@ std::vector<Prediction> ForestPredictor::predict(const Forest& forest,
                                                  const Data& data,
                                                  bool estimate_variance,
                                                  bool oob_prediction) const {
-  if (estimate_variance && forest.get_ci_group_size() <= 1) {
-    throw std::runtime_error("To estimate variance during prediction, the forest must"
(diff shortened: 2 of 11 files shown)


erikcs:
Very nice Apoorva, maybe you could convert this to a "Draft PR" then just use this as a playground for learning and fighting with C++/bindings using grf's build system at your own pace? I can happily give some input if you tally up questions. For a general scaffold I think anything James Yang does is awesome and looking for inspiration there is probably a good idea: https://github.com/JamesYang007/adelie

apoorvalal:
Thanks Erik! Yeah, that was the plan (i was using the low-fi solution of prefixing with Work in progress but draft PR works better). 

I also spent some time re-building [skgrf](https://github.com/apoorvalal/skgrf) from scratch in cython and found it works fine now, btw; I might develop that in parallel since it is less daunting to have to juggle serialization / pointers etc on cython [all the wrapper code lives in [one place](https://github.com/apoorvalal/skgrf/blob/main/skgrf/grf.pyx) in contrast to the pybind11 solution).

Will keep you posted

erikcs:
Is this still something you want to devote large chunk of time to? If not let's close this for now? (for users who don't want to leave python, stitching together ryp/rpy2 could perhaps be an easier alternative)

erikcs:
That sounds like a great and practical idea! If you find it engaging, perhaps you could create an example gallery on your page to gather user feedback and refine the approach. (For contributions to the grf repo, we generally aim for the level of reliability seen in Eddelbuettel’s work. The ryp project, like many Python projects, might benefit from some additional bug-fixing to enhance its robustness.)

erikcs:
Closing in favor of #257 which discusses R/Python interop.

PR #1502, opened 2025-06-19
Clarify diagnostics vignette

### LINKED ISSUE #1501: Method in diagnostics tutorial over-rejects
This issue is about a method proposed in the tutorial, [Evaluating a causal forest fit](https://grf-labs.github.io/grf/articles/diagnostics.html). 

One heuristic method to detect heterogeneity described in the tutorial over-rejects under the null, I believe due to the winner's curse. (The same model is used to determine subgroups and implement estimation). Cross-fitting resolves this problem, and I have proposed a small modification to the tutorial that provides this suggestion in a pull-request (#1502 ). 

**Description of the bug**
```
tau.hat <- predict(cf)$predictions
high.effect <- tau.hat > median(tau.hat)
ate.high <- average_treatment_effect(cf, subset = high.effect)
ate.low <- average_treatment_effect(cf, subset = !high.effect)
ate.high[["estimate"]] - ate.low[["estimate"]] +
  c(-1, 1) * qnorm(0.975) * sqrt(ate.high[["std.err"]]^2 + ate.low[["std.err"]]^2)
#> [1] 0.6591796 1.0443646
```

(this is just the method as referenced in the tutorial, the bug is the average rejection rate)

**Steps to reproduce**
Even when the sharp null is true (i.e., treatment is not associated with outcomes), this method rejects at higher than nominal rates. 

I used the code from the tutorial, and created a gist showing over-rejection here:

https://gist.github.com/mollyow/c1690ac8fd4a8d333d61cdefeeef82a9

A longer write-up is available here: https://alexandercoppock.com/testing_with_grf.pdf

**GRF version**
2.4.0 (but it's just about the tutorial, not the underlying code).

mollyow:
This PR is about a method proposed in the tutorial, [Evaluating a causal forest fit](https://grf-labs.github.io/grf/articles/diagnostics.html). 

One heuristic method to detect heterogeneity described in the tutorial over-rejects under the null, I believe due to the winner's curse. (The same model is used to determine subgroups and implement estimation). Cross-fitting resolves this problem, and I have proposed a small modification to the tutorial that provides this suggestion in this pull-request.

### DIFF
--- a/r-package/grf/vignettes/diagnostics.Rmd
+++ b/r-package/grf/vignettes/diagnostics.Rmd
@@ -64,31 +64,16 @@ The forest summary function [test_calibration](https://grf-labs.github.io/grf/re
 test_calibration(cf)
 ```
 
-For another way to assess heterogeneity, see the function [rank_average_treatment_effect](https://grf-labs.github.io/grf/reference/rank_average_treatment_effect.html) and the accompanying [vignette](https://grf-labs.github.io/grf/articles/rate.html).
+This exercise and function is motivated by earlier developments in the econometrics literature. A more intuitive exercise is to look at subgroup ATEs where the subgroups are formed according to low or high CATE predictions (Athey & Wager, 2019). 
+While this approach may give some qualitative insight into heterogeneity, the grouping is naive, because the doubly robust scores used to determine subgroups are not independent of the scores used to estimate those group ATEs. 
 
-A heuristic for testing for heterogeneity involves grouping observations into a high and low CATE group, then estimating average treatment effects in each subgroup. The function [average_treatment_effect](https://grf-labs.github.io/grf/reference/average_treatment_effect.html) estimates ATEs using a double robust approach:
-
-```{r}
-tau.hat <- predict(cf)$predictions
-high.effect <- tau.hat > median(tau.hat)
-ate.high <- average_treatment_effect(cf, subset = high.effect)
-ate.low <- average_treatment_effect(cf, subset = !high.effect)
-```
-
-Which gives the following 95% confidence interval for the difference in ATE
-
-```{r}
-ate.high[["estimate"]] - ate.low[["estimate"]] +
-  c(-1, 1) * qnorm(0.975) * sqrt(ate.high[["std.err"]]^2 + ate.low[["std.err"]]^2)
-```
-
-While this approach may give some qualitative insight into heterogeneity, the grouping is naive, because the doubly robust scores used to determine subgroups are not independent of the scores used to estimate those group ATEs (see Athey and Wager, 2019). 
-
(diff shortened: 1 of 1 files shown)


mollyow:
The heuristic approach appears to greatly over-reject under a true null of no homogeneity, and over-rejection rates increase as the sample gets larger. My colleague and I ran some simulations that show sometimes greater than 45% rejection rates:

https://alexandercoppock.com/testing_with_grf.pdf

download

So even as a heurisitic, without sample splitting the information is of questionable use. You can see at the linked note that sample splitting completely resolves this issue.

I've updated the tutorial to first emphasize the rate, then provide the heuristic approach with an additional caveat, but removing my worked example. 

It is of course your tutorial, so you can close this PR and leave as-is, I largely created the issue/PR as a formal way to flag the statistical issue.

erikcs:
That makes sense, there's a note on what you and your colleague observe [here](https://grf-labs.github.io/grf/REFERENCE.html#subgroups-based-on-out-of-bag-oob-estimates), but I guess it's kind of buried (!)

This is an old vignette where the "assessing fit" paragraph has implicitly been superseded by [this](https://grf-labs.github.io/grf/articles/rate.html) and [that](https://grf-labs.github.io/grf/articles/rate_cv.html). I kinda just left this vignette in place to not break any existing URLs. On second thought, how do you feel about just deleting the OOB heuristic example and referring the reader to the (preferred) RATE vignette? Something like: (?)


```
(!Existing test calibration paragraph:)
The forest summary function [test_calibration](https://grf-labs.github.io/grf/reference/test_calibration.html) can be used to asses a forest's goodness of fit. A coefficient of 1 for `mean.forest.prediction` suggests that the mean forest prediction is correct and a coefficient of 1 for `differential.forest.prediction` suggests that the forest has captured heterogeneity in the underlying signal.

```{r}
test_calibration(cf)

(!delete old stuff and replace with:)

This exercise and function is motivated by earlier developments in the econometrics literature. A more intuitive exercise is to look at subgroup ATEs where the subgroups are formed according to low or high CATE predictions (Athey & Wager, 2019). If there is heterogeneity present, and our CATE model detects them, then we should observe statistically significantly different ATEs in the predicted subgroups.

The [RATE](https://grf-labs.github.io/grf/reference/rank_average_treatment_effect.html) function automates this exercise over all possible subgroups using the quantiles of the CATE predictions. If we use separate data to fit CATE models and estimate RATE metrics, we obtain a test statistic with expectation zero under no heterogeneity, which can be used to construct confidence intervals for the presence of treatment effect heterogeneity. For more details on this preferred approach, please see [this vignette](https://grf-labs.github.io/grf/articles/rate.html).

```

mollyow:
I am happy with that!

erikcs:
Great, thanks!

erikcs:
Did you want to make these edits to the vignette? 🙏

mollyow:
done!

PR #1509, opened 2025-09-23
Add AcceleratedSurvivalSplittingRule

erikcs:
Add faster splitting rule class `AcceleratedSurvivalSplittingRule` that computes an approximation of the logrank criterion in `SurvivalSplittingRule`, but which can be updated in constant time for every split point (a short note with the computational details is [here](https://arxiv.org/abs/2510.03665)). This works essentially exactly the same as `SurvivalSplittingRule`, except that the numerator of the criterion is updated efficiently, while the denominator of the criterion is replaced with a simple approximation that is also updated efficiently.

The statistical performance is practically the same, but prediction output may differ slightly compared to exact criterion. For backwards compatibility, a new argument `fast.logrank` defaulting to `FALSE` is added to `survival_forest` and `causal_survival_forest`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -169,7 +169,7 @@ Erik Sverdrup, Maria Petukhova, and Stefan Wager.
 
 Erik Sverdrup, James Yang, and Michael LeBlanc.
 <b>Efficient Log-Rank Updates for Random Survival Forests.</b> 2025.
-[<a href="TODO">arxiv</a>]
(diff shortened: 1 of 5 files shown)


erikcs:
Pinging @sebp in case this is of interest: This simple change updates the log-rank criterion used for splitting in random survival forests. The speed gains are large for data with many events, with negligible loss of accuracy. A short note is [here](https://github.com/user-attachments/files/22704609/Efficient_log_rank_updates.pdf).

